PR #1, opened 2021-11-09
feat: add some paths and utils

alexeagle:


lib/utils.bzl
```diff
@@ -0,0 +1,10 @@
+"Public API"
+
+load("//lib/private:utils.bzl", "utils")
+
+is_external_label = utils.is_external_label
```
alexeagle, line 5: do you think we should decompose the struct here? or just re-export the whole struct?

mattem (approved):

lib/utils.bzl
gregmagolan: I think people expect to import individual functions from a utils.bzl and not a struct of them

lib/private/utils.bzl
```diff
@@ -0,0 +1,102 @@
...
+        param: a string or label
+
+    Returns:
+        a bool
+    """
+    return len(_to_label(param).workspace_root) > 0
+
+# Path to the root of the monorepo
+def _path_to_root():
+    """ Retuns the path to the monorepo root under bazel
```
gregmagolan, line 81: monorepo => workspace

lib/private/utils.bzl
```diff
@@ -0,0 +1,102 @@
...
+
+    Returns:
+        Path to the monorepo root
```
gregmagolan, line 84: monorepo => workspace

lib/private/utils.bzl
```diff
@@ -0,0 +1,102 @@
...
+        "block-network",
+        "requires-fakeroot",
+        "exclusive",
+        "manual",
+        "external",
+    ]
+
+    return [tag for tag in tags if tag in WELL_KNOWN_TAGS]
+
+def _is_darwin(rctx):
```
gregmagolan, line 30: suppose we should include `_is_windows` and `_is_linux`?

gregmagolan (approved):
🌮

lib/private/paths.bzl
```diff
@@ -0,0 +1,47 @@
+"""Path utils built on top of Skylib's path utils"""
+
+load("@bazel_skylib//lib:paths.bzl", _spaths = "paths")
+
+def _relative_file(to_file, frm_file):
```
gregmagolan, line 5: Would these params be better named `to` and `from` instead of `to_file` and `frm_file`?

alexeagle: from is a reserved keyword

PR #3, opened 2021-11-10
feat: add runfiles helpers needed for rules_js

alexeagle:


lib/utils.bzl
```diff
@@ -7,3 +7,62 @@ glob_directories = utils.glob_directories
 path_to_workspace_root = utils.path_to_workspace_root
 propagate_well_known_tags = utils.propagate_well_known_tags
 to_label = utils.to_label
+
+BATCH_RLOCATION_FUNCTION = r"""
```
cvarier, line 11: Some suggestions:
- If I'm not mistaken, this was ported over from https://github.com/bazelbuild/rules_nodejs/blob/aa09b5796faaf2470b84246a8f00e6e27181c63b/internal/common/windows_utils.bzl. From what I understand, this is only used for building on Windows, do you think it would be better to put this into a file called `windows_utils.bzl`?
- I don't see that this is being used anywhere currently; if it is only intended to be exposed for client consumption would it be worth adding unit tests for?

lib/utils.bzl
```diff
@@ -7,3 +7,62 @@ glob_directories = utils.glob_directories
...
+  echo>&2 ERROR: !runfile_path! not found in runfiles manifest
+  exit 1
+)
+set %~2=!abs_path!
+exit /b 0
+:rlocation_end
+:: End of rlocation
+"""
+
+BASH_RLOCATION_FUNCTION = r"""
```
cvarier, line 57: Same suggestions as above, but additionally: 
- Since there's not much commentary going on here, it's a little hard to understand what it's doing or why it's needed.
- As this was copy pasted, would you be able to link to the source?

lib/private/paths.bzl
```diff
@@ -42,6 +42,28 @@ def _relative_file(to_file, frm_file):
         )
     )
 
+# 
+#
```
cvarier, line 46: nit: empty comment lines

lib/utils.bzl
alexeagle: Yes, that's where it came from. Making a windows specific file SGTM, we'll have other helpers to go there.
(I'm assuming we'll have to refactor the organization here a bit as we add more code)

It will only be used in downstream rules (rules_js to start). Testing is hard because we don't have windows executors on GitHub actions yet.

lib/utils.bzl
alexeagle: good suggestions thanks :)

PR #11, opened 2021-12-09
Add jq toolchain and rule

kormide:
@gregmagolan This is still being drafted and needs some work and tests, but I wouldn't mind an interim review since this is my first toolchain.

Usage
```
# WORKSPACE
load("@aspect_bazel_lib//lib:repositories.bzl", "register_jq_toolchains")
register_jq_toolchains(version="1.6")
```

```
# BUILD.bazel
jq(
  name = "merge",
  srcs = [":in1.json", ":in2.json"],
  filter = ".[0] * .[1]",
  args = ["-s"],
)
```


lib/private/jq.bzl
```diff
@@ -0,0 +1,41 @@
+"TODO"
+
+def _jq_impl(ctx):
+    filter = ctx.attr.filter
+
+    out = ctx.actions.declare_file(ctx.attr.name + "_jq.json")
```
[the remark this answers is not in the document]
gregmagolan: Generally it is really nice to have yea

lib/private/jq.bzl
```diff
@@ -0,0 +1,41 @@
...
+        use_default_shell_env = True,
+    )
+
+    return DefaultInfo(files = depset([out]))
+
+jq = rule(
+    implementation = _jq_impl,
+    attrs = {
+        "input": attr.string(doc = "String json input (alternative to srcs)"),
+        "srcs": attr.label_list(allow_files = [".json"], doc = "Json files to be filtered"),
```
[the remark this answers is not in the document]
gregmagolan: Raw string is pretty useless. I think take that out of the API. In bazel land you can always use a `write_file` and then pipe that file to this rule if you wanted to hard-code a string for a test case.

lib/private/jq.bzl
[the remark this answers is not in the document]
gregmagolan: And probably allow the user to specify the `out` filename and default to `"%s.json" % name` in the macro if not specified

lib/repositories.bzl
```diff
@@ -0,0 +1,17 @@
+"Register toolchains"
+
+load("//lib/private:toolchain.bzl", "JQ_PLATFORMS", "jq_repositories", "jq_toolchains_repo")
+
+def rules_jq_toolchains(version, name = ""):
```
gregmagolan, line 5: Name shouldn't default to "" and should come first:
```
def rules_jq_toolchains(name, version)
```

lib/private/jq.bzl
```diff
@@ -0,0 +1,41 @@
...
+        "filter": attr.string(mandatory = True, doc = "jq filter specification"),
+        "args": attr.string_list(doc = "Arguments to pass to jq"),
+    },
+    toolchains = ["@aspect_bazel_lib//lib/private:jq_toolchain_type"],
```
alexeagle, line 40: the toolchain type should be public

lib/repositories.bzl
gregmagolan, line 5: Probably just `register_jq_toolchains` or `jq_register_toolchains` with the `rules_js` since it is not is its own ruleset

lib/private/toolchain.bzl
```diff
@@ -0,0 +1,171 @@
...
+
+    build_content = """# Generated by lib/toolchain.bzl
+#
+# These can be registered in the workspace file or passed to --extra_toolchains flag.
+# By default all these toolchains are registered by the jq_register_toolchains macro
+# so you don't normally need to interact with these targets.
+
+load(":defs.bzl", "resolved_toolchain")
+
+resolved_toolchain(name = "resolved_toolchain", visibility = ["//visibility:public"])
```
alexeagle, line 113: can you make an example that calls jq from a `genrule` instead to demonstrate this?

alexeagle:
The term "rules_jq" shouldn't appear, it's not a separate ruleset

lib/private/jq.bzl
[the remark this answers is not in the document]
kormide: Added optional pre-declared output.

lib/private/jq.bzl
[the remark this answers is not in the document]
kormide: Removed.

lib/repositories.bzl
kormide: Fixed name.

kormide: So my intention here was to allow the user to modify the name for namespacing purposes in case the default repository names we generate conflict with anything else they have. That seems unlikely...

What do you usually do in these scenarios?

kormide:
Updated to address feedback and add tests.

lib/tests/jq/BUILD.bazel
```diff
@@ -0,0 +1,104 @@
...
+
+# Call jq within a genrule
+genrule(
+    name = "case_genrule",
+    srcs = ["a.json"],
+    outs = ["genrule_output.json"],
+    cmd = "$(JQ_BIN) '.' $(location a.json) > $@",
+    tags = ["manual"],
+    toolchains = ["@_aspect_bazel_lib_jq_linux64//:jq_toolchain"],
+)
```
kormide, line 98: @alexeagle I was hoping that I could reference the toolchain type for this test, but according to the docs you have to provide a specific instance of a toolchain rule.

This should fail on non 64 bit linux platforms. What would you normally do in this situation?

lib/private/toolchain.bzl
```diff
@@ -0,0 +1,182 @@
...
+
+JQ_VERSIONS = {
+    "1.6": {
+        "linux32": "sha384-hBGwNC3R0WyEbDQnrabzvcURSSV9BGxVrUVXLCH1C+Ilo7YDlzfTJSr4gadVssVI",
+        "linux64": "sha384-+K6tuwxrC/P4WBYRJ7YXcpeLS7GesbbnUhq4r9w7k0lCUC1KlhyXXf0sFQgOg0dI",
+        "oxs-amd64": "sha384-ZLZljM9OyKCJbJbv7s1SRYSeMbVxfRc6kFNUlk9U/IL10Xm2xr4cxx3SZKv93QFO",
+        "win32": "sha384-PO+MMFELa0agwy35NuKhrxn8C6GjNq8gnzL3NvYSWNx/pwClCl7yzCONGhLFknMc",
+        "win64": "sha384-O4qdyhb+0zU1XAuUKc1Mil5hlbSbCUcPQOGRtkJUqryv7X0IeKcMCIuZw97q9WGr",
+    },
+    # TODO: other versions
```
kormide, line 44: I will add some more versions before this gets merged.

lib/private/toolchain.bzl
kormide: Added

internal_deps.bzl
```diff
@@ -6,9 +6,15 @@ statement from these, that's a bug in our distribution.
 
 load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
 load("@bazel_tools//tools/build_defs/repo:utils.bzl", "maybe")
+load("//lib:repositories.bzl", "register_jq_toolchains")
+
+def bazel_lib_internal_deps(name = ""):
+    """Fetch deps needed for local development
+
+    Args:
+        name: unused, needed to stop buildifier from complaining
```
alexeagle, line 15: I'd probably just suppress it instead. 

> `# buildifier: disable=unnamed-macro`

from https://github.com/bazelbuild/buildtools/blob/master/WARNINGS.md#unnamed-macro

lib/private/jq.bzl
```diff
@@ -0,0 +1,48 @@
...
+        tools = [jq_bin],
+        inputs = ctx.files.srcs,
+        outputs = [out],
+        command = cmd,
+        use_default_shell_env = True,
+    )
+
+    return DefaultInfo(files = depset([out]))
+
+jq = rule(
```
alexeagle, line 32: a nicer pattern to allow others to customize is to export a struct instead of a rule

```
jq = struct(
    implementation = _jq_impl,
...
```

Now someone else can make their own rule with different attrs or something, using our implementation function as a part of theirs.

The /lib/jq.bzl file would then be the actual declaration of the rule we expose

lib/jq.bzl
```diff
@@ -0,0 +1,5 @@
+"TODO"
```
alexeagle, line 1: add the docs/ change to this PR as well

We should add those docs now, no reason to leave a TODO for later

lib/private/jq.bzl
```diff
@@ -0,0 +1,48 @@
+"""Implementation for jq rule"""
+
+def _jq_impl(ctx):
```
alexeagle, line 3: did you get any relevant goodies from @f0rmiga already?

kormide:
I suppose I should also run the build and tests on a windows box.

lib/private/jq.bzl
```diff
@@ -0,0 +1,48 @@
...
+    if ctx.outputs.out:
+        out = ctx.outputs.out
+    else:
+        out = ctx.actions.declare_file("%s.json" % ctx.attr.name)
```
alexeagle, line 7: opt: I'm on the fence about whether it's worth having a default for the out. Even if it is, we might want that in a wrapper macro rather than the rule, just to keep it simpler.

lib/private/jq.bzl
```diff
@@ -0,0 +1,48 @@
...
+    filter = ctx.attr.filter
+    sources = " ".join([file.path for file in ctx.files.srcs])
+
+    args = ctx.attr.args
+
+    # jq hangs when there are no input sources unless --null-input flag is passed
+    if len(ctx.attr.srcs) == 0 and "-n" not in args and "--null-input" not in args:
+        args = args + ["--null-input"]
+
+    cmd = "%s %s '%s' %s > %s" % (jq_bin.path, " ".join(args), ctx.attr.filter, sources, out.path)
```
alexeagle, line 20: nit: instead of joining args yourself, consider using the ctx.actions.args helper.

It has the advantage of doing lazy computation to create the args string only when the action is actually being spawned, rather than during the analysis phase when this rule code is executed.

lib/private/jq.bzl
```diff
@@ -0,0 +1,48 @@
...
+        out = ctx.actions.declare_file("%s.json" % ctx.attr.name)
+
+    jq_bin = ctx.toolchains["@aspect_bazel_lib//lib:jq_toolchain_type"].jqinfo.bin
+
+    filter = ctx.attr.filter
+    sources = " ".join([file.path for file in ctx.files.srcs])
+
+    args = ctx.attr.args
+
+    # jq hangs when there are no input sources unless --null-input flag is passed
```
alexeagle, line 16: nice job commenting the surprising behavior

lib/private/jq.bzl
```diff
@@ -0,0 +1,48 @@
...
+def _jq_impl(ctx):
+    if ctx.outputs.out:
+        out = ctx.outputs.out
+    else:
+        out = ctx.actions.declare_file("%s.json" % ctx.attr.name)
+
+    jq_bin = ctx.toolchains["@aspect_bazel_lib//lib:jq_toolchain_type"].jqinfo.bin
+
+    filter = ctx.attr.filter
+    sources = " ".join([file.path for file in ctx.files.srcs])
```
alexeagle, line 12: avoid the join here if you can help it.

Also it clearly breaks when a source file contains a space in the name, though it's probably a problem somewhere else anyway

lib/private/jq.bzl
```diff
@@ -0,0 +1,48 @@
...
+
+    ctx.actions.run_shell(
+        tools = [jq_bin],
+        inputs = ctx.files.srcs,
+        outputs = [out],
+        command = cmd,
+        use_default_shell_env = True,
+    )
+
+    return DefaultInfo(files = depset([out]))
```
alexeagle, line 30: may want to expose it as a runfile as well, so any binaries depending on this rule can find a data dependency on the produced file.

```DefaultInfo(files = depset([out]), runfiles = [out])```
I think

File: lib/private/jq.bzl [hunk 17]
```diff
@@ -0,0 +1,48 @@
...
+    implementation = _jq_impl,
+    attrs = {
+        "srcs": attr.label_list(
+            allow_files = [".json"],
+            doc = "Json files to be filtered",
+            mandatory = True,
+            allow_empty = True,
+        ),
+        "filter": attr.string(
+            doc = "jq filter specification",
```
alexeagle, line 42: shouldn't it be mandatory also?

lib/private/toolchain.bzl
alexeagle: we have a common pattern to use a GitHub action to periodically check for new releases upstream and send the PR updating this file. Otherwise you are signing yourself or someone else up for a bunch of toil later.

We can do it as a followup

lib/private/toolchain.bzl
```diff
@@ -0,0 +1,182 @@
...
+}
+
+JqInfo = provider(
+    doc = "Provide info for executing jq",
+    fields = {
+        "bin": "Executable jq binary",
```
alexeagle, line 50: there's also a `yq` variant we use at RH, which handles yaml files. Maybe a TODO for another PR.

lib/private/toolchain.bzl
```diff
@@ -0,0 +1,182 @@
...
+    return [
+        toolchain_info,
+        toolchain_info.default,
+        toolchain_info.jqinfo,
+        toolchain_info.template_variables,
+    ]
+
+# Copied from java_toolchain_alias
+# https://cs.opensource.google/bazel/bazel/+/master:tools/jdk/java_toolchain_alias.bzl
+resolved_toolchain = rule(
```
alexeagle, line 108: fun, huh? Bazel so many rough edges

lib/tests/jq/BUILD.bazel
```diff
@@ -0,0 +1,104 @@
+load("@bazel_skylib//rules:diff_test.bzl", "diff_test")
+load("//lib:jq.bzl", "jq")
+
+# No filter produces identical json
+jq(
+    name = "case_no_filter",
+    srcs = ["a.json"],
+    tags = ["manual"],
```
alexeagle, line 8: why manual??

lib/tests/jq/BUILD.bazel
alexeagle: this is where you use the `@aspect_bazel_lib_jq_toolchains//:resolved_toolchain`

alexeagle (commented):
close! nice job :)

lib/tests/jq/BUILD.bazel
kormide: Ooooh that's what that's for..

lib/private/jq.bzl
kormide: Passing in an empty filter is equivalent to passing in `.` in jq.

kormide: I suppose I could still make it mandatory though, and users have to write `filter = ""`.

lib/private/jq.bzl
kormide: He pointed me to a tool that SUSE used to fetch binaries, but I don't think there was a publicly available jq rule.

lib/tests/jq/BUILD.bazel
kormide: The bazel docs on testing recommend adding manual tags so that these test targets don't get built on `bazel build //...`.

lib/private/jq.bzl
f0rmiga: What @alexeagle said + this is really bad for readability. Use named format strings, e.g.:
```python
"'{jq}' {args} '{jq_filter}'".format(
    jq = jq_bin.path,
    " ".join(args),
    jq_filter = ctx.attr.filter,
)
```

lib/private/jq.bzl
```diff
@@ -0,0 +1,48 @@
...
+        args = args + ["--null-input"]
+
+    cmd = "%s %s '%s' %s > %s" % (jq_bin.path, " ".join(args), ctx.attr.filter, sources, out.path)
+
+    ctx.actions.run_shell(
+        tools = [jq_bin],
+        inputs = ctx.files.srcs,
+        outputs = [out],
+        command = cmd,
+        use_default_shell_env = True,
```
f0rmiga, line 27: I think by default, this should be `False` on rules we write unless there's a clear reason for doing otherwise.

lib/private/jq.bzl
f0rmiga: Make users type `.`, it reads better since it's what users do on the terminal.

lib/tests/jq/BUILD.bazel
alexeagle: Really? that doesn't make sense, there's no reason to skip build and test everything.

Where is that doc? maybe they meant that just for targets that are dummies as inputs to some test

lib/private/jq.bzl
kormide: It seems like if I use `ctx.actions.args` then I still have to reference them positionally in the command string, e.g., `jq $1 $2 ...`, so it doesn't help with readability. But using `ctx.actions.args` probably deals with the case of a file name having a space.

lib/tests/jq/BUILD.bazel
kormide: I misread the docs. It recommends adding this tag for failure testing---rules that are intended to fail.

lib/private/jq.bzl
f0rmiga: While you are under Bash, you will need to quote paths regardless of using `ctx.actions.args` or not.

f0rmiga:
If you keep force-pushing, it makes review hard since the GitHub will not track the previous comments. Try to avoid amends and interactive rebases while the review is in progress. The merge will be squashed in any case at the end, so no need to keep the commits tidy.

kormide:
> If you keep force-pushing, it makes review hard since the GitHub will not track the previous comments. Try to avoid amends and interactive rebases while the review is in progress. The merge will be squashed in any case at the end, so no need to keep the commits tidy.

Sure. Do you have your preferred practices documented somewhere?

kormide:
After this jq addition, is this statement in the README still true?

> This is code we would contribute to bazel-skylib, but the declared scope of that project is narrow and it's very difficult to get anyone's attention to review PRs there.

lib/private/jq.bzl
gregmagolan: That's a good point. If its in the wrapper macro then the default `<name>.json` is a pre-declared output. Should put it there to start.

kormide: Oops, I added a wrapper macro but forgot to delete this part.

lib/jq.bzl
```diff
@@ -0,0 +1,34 @@
...
+            mandatory = True,
+        ),
+        "args": attr.string_list(doc = "Arguments to pass to jq"),
+        "out": attr.output(doc = "Name for the output json file", mandatory = True),
+    },
+    toolchains = ["@aspect_bazel_lib//lib:jq_toolchain_type"],
+)
+
+def jq(name, filter, srcs, args = [], out = None, **kwargs):
+    if out == None:
```
gregmagolan, line 25: nit: `if not out:` will cover the empty string case as well

lib/jq.bzl
```diff
@@ -0,0 +1,34 @@
+"""A rule that applies a jq filter to one or more json files"""
+
+load("//lib/private:jq.bzl", _jq = "jq")
+
+_jq_rule = rule(
+    implementation = _jq.implementation,
+    attrs = {
+        "srcs": attr.label_list(
+            allow_files = [".json"],
+            doc = "Json files to be filtered",
```
gregmagolan, line 10: I believe the docstrings needs to move to the `jq` macro; tho there may be some trick to get them to get picked up properly for the generated docs. @alexeagle knows best on that one.

lib/jq.bzl
gregmagolan, line 25: doc string with args :)

.prettierignore
```diff
@@ -1 +1,2 @@
 docs/*.md
+lib/tests/jq/*.json
```
alexeagle, line 2: I assume this is still required?

lib/private/jq_toolchain.bzl
```diff
@@ -0,0 +1,188 @@
...
+            "@platforms//cpu:x86_32",
+        ],
+    ),
+    "linux64": struct(
+        compatible_with = [
+            "@platforms//os:linux",
+            "@platforms//cpu:x86_64",
+        ],
+    ),
+    "oxs-amd64": struct(
```
alexeagle, line 16: spelling!

I figured out how we'll get tests for this from the Bazel team on their buildkite :)
But not until release-time

lib/private/jq_toolchain.bzl
```diff
@@ -0,0 +1,188 @@
...
+            "@platforms//cpu:x86_64",
+        ],
+    ),
+}
+
+JQ_VERSIONS = {
+    "1.6": {
+        "linux32": "sha384-hBGwNC3R0WyEbDQnrabzvcURSSV9BGxVrUVXLCH1C+Ilo7YDlzfTJSr4gadVssVI",
+        "linux64": "sha384-+K6tuwxrC/P4WBYRJ7YXcpeLS7GesbbnUhq4r9w7k0lCUC1KlhyXXf0sFQgOg0dI",
+        "oxs-amd64": "sha384-ZLZljM9OyKCJbJbv7s1SRYSeMbVxfRc6kFNUlk9U/IL10Xm2xr4cxx3SZKv93QFO",
```
alexeagle, line 40: search for `oxs`

.prettierignore
kormide: Yeah I don't want prettier to touch my json test files, so that I can test that `jq` beautifies them.

lib/private/jq_toolchain.bzl
kormide: :scream_cat:

kormide:
Updated and ready for review. I fixed the platform issues, tests on windows, generated docs, and simplified the repository names.

lib/private/jq.bzl
```diff
@@ -0,0 +1,47 @@
...
+    ),
+    "args": attr.string_list(),
+    "out": attr.output(mandatory = True),
+}
+
+def _jq_impl(ctx):
+    jq_bin = ctx.toolchains["@aspect_bazel_lib//lib:jq_toolchain_type"].jqinfo.bin
+
+    out = ctx.outputs.out
+    filter = ctx.attr.filter
```
gregmagolan, line 20: filter looks unused as you're using `ctx.attr.filter` below.

Also looks like you can remove the `out` reassignment here and just use `ctx.outputs.out.path` below

lib/private/jq.bzl
```diff
@@ -0,0 +1,47 @@
...
+    args = ctx.attr.args
+
+    # jq hangs when there are no input sources unless --null-input flag is passed
+    if len(ctx.attr.srcs) == 0 and "-n" not in args and "--null-input" not in args:
+        args = args + ["--null-input"]
```
gregmagolan, line 25: nit: `args.append("--null-input")` is simpler as args is only referred to once

lib/private/jq_toolchain.bzl
```diff
@@ -0,0 +1,193 @@
...
+    ),
+    "win64": struct(
+        compatible_with = [
+            "@platforms//os:windows",
+            "@platforms//cpu:x86_64",
+        ],
+    ),
+}
+
+JQ_VERSIONS = {
```
gregmagolan, line 35: nit: leave a comment as to where to look for future releases & how to get the SHAs for future updaters

lib/repositories.bzl
```diff
@@ -0,0 +1,37 @@
...
+        http_archive,
+        name = "bazel_skylib",
+        sha256 = "c6966ec828da198c5d9adbaa94c05e3a1c7f21bd012a0b29ba8ddbccb2c93b0d",
+        urls = [
+            "https://github.com/bazelbuild/bazel-skylib/releases/download/1.1.1/bazel-skylib-1.1.1.tar.gz",
+            "https://mirror.bazel.build/github.com/bazelbuild/bazel-skylib/releases/download/1.1.1/bazel-skylib-1.1.1.tar.gz",
+        ],
+    )
+
+# buildifier: disable=unnamed-macro
```
kormide, line 19: Can get rid of this now that I have a `name`.

lib/jq.bzl
```diff
@@ -0,0 +1,30 @@
...
+    toolchains = ["@aspect_bazel_lib//lib:jq_toolchain_type"],
+)
+
+def jq(name, srcs, filter, args = [], out = None):
+    """Invoke jq with a filter on a set of json input files.
+
+    Args:
+        name: Name of the rule
+        srcs: List of input json files
+        filter: jq filter specification
```
gregmagolan, line 17: nit: drop a link to docs on jq filters

lib/jq.bzl
```diff
@@ -0,0 +1,30 @@
...
+
+_jq_rule = rule(
+    attrs = _jq_lib.attrs,
+    implementation = _jq_lib.implementation,
+    toolchains = ["@aspect_bazel_lib//lib:jq_toolchain_type"],
+)
+
+def jq(name, srcs, filter, args = [], out = None):
+    """Invoke jq with a filter on a set of json input files.
+
```
gregmagolan, line 13: nit: drop a link to the jq project homepage

lib/jq.bzl
```diff
@@ -0,0 +1,30 @@
...
+    Args:
+        name: Name of the rule
+        srcs: List of input json files
+        filter: jq filter specification
+        args: jq-specific args to pass
```
gregmagolan, line 18: nit: "additional args to pass to jq"

lib/private/jq.bzl
```diff
@@ -0,0 +1,47 @@
+"""Implementation for jq rule"""
+
+_jq_attrs = {
+    "srcs": attr.label_list(
+        allow_files = [".json"],
+        mandatory = True,
+        allow_empty = True,
+    ),
+    "filter": attr.string(
+        mandatory = True,
```
gregmagolan, line 10: can jq do anything useful without a filter? if so, then it shouldn't be mandatory; if it is mandatory then lets mention that in the docstring

lib/private/jq.bzl
kormide: `args` is imutable because it comes from `ctx.attr.args`.

lib/private/jq.bzl
kormide: It can't do anything useful without a filter.

lib/private/jq.bzl
gregmagolan: oh.. so it is. nevermind :)

gregmagolan (approved):
🌮

kormide:
Squashed

PR #16, opened 2021-12-15
docs: add jq usage examples

kormide:


lib/jq.bzl
```diff
@@ -13,6 +13,42 @@ def jq(name, srcs, filter, args = [], out = None):
...
     For jq documentation, see https://stedolan.github.io/jq/.
 
+    Usage examples:
+
+        # Remove fields from package.json
+        jq(
+            name = "no_dev_deps",
+            srcs = ["package.json"],
+            filter = "del(.devDependencies)",
+        )
```
kormide, line 23: I was surprised I didn't have to add any backticks for this to be formatted as code in the .md file. Adding backticks just made the backtick literals appear.

alexeagle: indent is also markdown syntax for a codeblock.
However you don't have the syntax hint.

Better to dedent and use triple-backtick fence with "starlark" hint

kormide:
Updated.

alexeagle (approved):

PR #19, opened 2021-12-18
fix: allow tags/visibility on jq rule

### LINKED ISSUE #18: jq rule doesn't accept tags
it probably should, along with visibility and other common attrs

alexeagle:
fixes #18


kormide (approved):
Oops. Thanks.

PR #21, opened 2021-12-20
feat: load jq filter from file

kormide:
@gregmagolan || @alexeagle 

I need to use a dynamically generated filter file for the angular cli build.


alexeagle (approved):
very nice 😎

PR #22, opened 2021-12-21
fix: add windows implementation of copy_to_directory

gregmagolan:
Still have to make diff_test work on Windows but the impl of copy_to_directory for Windows is done and appears to be working as expected from manual inspection of outputs


alexeagle (approved):

kormide (changes requested):
I think there's a doc string at the top of the file referencing incompatibility with windows.

kormide (approved):
lgtm

kormide:
nit: might be helpful to separate your changes to diff_test in a commit in case someone else looks back to the history to make changes or fix any bugs

PR #23, opened 2021-12-21
license under apache 2

alexeagle:
Note, we think the bazel-lib is so useful and reduces our toil in maintaining rules like rules_nodejs, and want other rules authors to feel comfortable contributing here, so we are choosing the same license as that repository rather than a commercial license here.


gregmagolan (approved):
🌮

PR #24, opened 2022-01-14
[utils] propagate CPU tags

mattem:


lib/private/utils.bzl
```diff
@@ -29,7 +29,12 @@ def _propagate_well_known_tags(tags = []):
         "external",
     ]
 
-    return [tag for tag in tags if tag in WELL_KNOWN_TAGS]
+    CPU_PREFIX = "cpu:"
```
alexeagle, line 32: maybe a link to the docs would be nice, if there is something explaining this tags behavior

alexeagle (approved):

PR #25, opened 2022-01-17
build: fix ci build caching

kormide:


.github/workflows/release.yml
```diff
@@ -14,6 +14,13 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v2
+      - name: Mount bazel repo cache
```
kormide, line 17: I figure we should never mount the action cache for release builds in case there's stamping.

alexeagle: from slack - we should mount the action cache and rely on our keys being "stable" so they are part of the action key

alexeagle (approved):

.github/workflows/ci.yaml
```diff
@@ -21,18 +21,16 @@ jobs:
...
+            "~/.cache/bazel"
+            "~/.cache/bazel-repo"
+          key: bazel-cache-${{ hashFiles('**/BUILD.bazel', '**/*.bzl', 'WORKSPACE') }}
+          restore-keys: bazel-cache-
       - name: bazel test //...
         env:
           # Bazelisk will download bazel to here
           XDG_CACHE_HOME: ~/.cache/bazel-repo
-        run: bazel --bazelrc=.github/workflows/ci.bazelrc --bazelrc=.bazelrc test //...
+        run: bazel --bazelrc=.github/workflows/ci.bazelrc --bazelrc=.bazelrc test //...
```
f0rmiga, line 36: You need to fix the bazelrc at the same time to remove the environment variable from there. The bazelrc DOES NOT support environment variables. You will need to pass the flags that use the `HOME` env var here.

f0rmiga (changes requested):

.github/workflows/ci.yaml
kormide: Fixed. Going through the other PRs to fix this everywhere.

alexeagle: well, you can put `~` character in bazelrc.

I just tested with `common --disk_cache=~/.thing/bazel` and

```
$ ls $HOME/.thing
bazel
```

Sorry to be a PITA but I think it's a lot nicer if the CI configs can rely on the .bazelrc to keep the flags consistent, otherwise it's very error prone to keep the various commands using the same location. It's easy to add some `bazel query ` command or something when the CI config gets long, and not notice that some flags were supposed to be repeated everywhere.

f0rmiga: You are not being a PITA. I agree that what we did so far is more error prone when adding more Bazel commands.

I have mixed feelings about the expansion of ~ in bazelrc. From a person who is used to do Bash, my head spins that it's able to do such a Bash-y thing and not handle environment variables... on the other hand, if I turn the pragmatic mode on, I like it for the reasons you mentioned that makes it less error prone.

alexeagle: I spent some time searching Bazel docs for a mention of the special handling of ~ but failed 😞

kormide: I agree that it's nicer. I'll do another round on the open prs and on the rules-template.

f0rmiga: @alexeagle do you still think it's better even it being undocumented api? Exercise some Hyrum's law here?

kormide:
Updated.

.github/workflows/ci.yaml
f0rmiga, line 36:
```suggestion
        run: bazel --bazelrc=.github/workflows/ci.bazelrc --bazelrc=.bazelrc test //...

```

f0rmiga (approved):

alexeagle (approved):

PR #28, opened 2022-01-27
feat: write_source_files

kormide:
Rule to update a file in the source tree.

This is a copy of @gregmagolan's implementation with some changes to use more rules rather than macros. I want to have a look over Chuck's updatesrc implementation to see what's different and how we can combine them before we go ahead with this.

### DIFF
--- a/lib/write_source_files.bzl
+++ b/lib/write_source_files.bzl
@@ -1,0 +1,131 @@
+"Public API for write_source_files"
+
+load("//lib/private:write_source_files.bzl", _lib = "write_source_files_lib")
+load("//lib:utils.bzl", _to_label = "to_label")
+load("@bazel_skylib//rules:diff_test.bzl", _diff_test = "diff_test")
+load("//lib/private:fail_with_message_test.bzl", "fail_with_message_test")
+
+_write_source_files = rule(
+    attrs = _lib.attrs,
+    implementation = _lib.implementation,
+    executable = True,
+)
+
+def write_source_files(name, files, **kwargs):
+    """Write to one or more files in the source tree. Stamp out tests that ensure the files exists and are up to date.
+
+    Usage:
+
+    ```starlark
+    load("@aspect_bazel_lib//lib:write_source_files.bzl", "write_source_files")
+
+    write_source_files(
+        name = "write_foobar",
+        files = {
+            "foobar.json": "//some/generated:file",
+        },
+    )
+    ```
+
+    To update the source file, run:
+    ```bash
+    bazel run //:write_foobar
+    ```
+
+    A test will fail if the source file doesn't exist
+    ```bash
+    bazel test //...
+
+    //:foobar.json does not exist. To create & update this file, run:
+
+        bazel run //:write_foobar
+    ```
+
+    ...or if it's out of date.
+    ```bash
+    bazel test //...
+
+    //:foobar.json is out-of-date. To update this file, run:
+
+        bazel run //:write_foobar
+    ```
+
+    Args:
+        name: Name of the executable target that creates or updates the source file
+        files: A dict where the keys are source files to write to and the values are labels pointing to the desired content.
+            Source files must be within the same bazel package as the target.
+        **kwargs: Other common named parameters such as `tags` or `visibility`
+    """
+
+    out_files = files.keys()
+    in_files = [files[f] for f in out_files]
+
+    # Stamp an executable rule that writes to the out file
+    _write_source_files(
+        name = name,
+        in_files = in_files,
+        out_files = out_files,
+        is_windows = select({
+            "@bazel_tools//src/conditions:host_windows": True,
+            "//conditions:default": False,
+        }),
+        visibility = kwargs.get("visibility"),
+        tags = kwargs.get("tags"),
+    )
+
+    # Fail if user passes args that would conflict with stamped out targets below
+    if kwargs.pop("file1", None) != None:
+        fail("file1 not a valid parameter in write_source_file")
+    if kwargs.pop("file2", None) != None:
+        fail("file2 not a valid parameter in write_source_file")
+    if kwargs.pop("failure_message", None) != None:
+        fail("failure_message not a valid parameter in write_source_file")
+
+    for i in range(len(out_files)):
+        out_file = _to_label(out_files[i])
+        out_file_missing = _is_file_missing(out_file)
+
+        name_test = "%s_%d_test" % (name, i)
+
+        if out_file_missing:
+            # Stamp out a test that fails with a helpful message when the source file doesn't exist.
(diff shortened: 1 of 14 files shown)


lib/update_source_file.bzl
```diff
@@ -0,0 +1,83 @@
...
+load("@bazel_skylib//rules:diff_test.bzl", _diff_test = "diff_test")
+load("@bazel_skylib//rules:write_file.bzl", _write_file = "write_file")
+
+update_source_file = rule(
+    attrs = _update_source_file_lib.attrs,
+    implementation = _update_source_file_lib.implementation,
+    executable = True,
+)
+
+def update_source_file_with_test(name, in_file, update_file, **kwargs):
```
kormide, line 14: Do you just want to expose this macro that stamps the rule with tests, or both the executable rule and the macro with test version?

kormide: Or should we just call this macro `update_source_file` and implicitly include the tests?

lib/private/update_source_file.bzl
```diff
@@ -0,0 +1,51 @@
+"update_source_file implementation"
+
+load("//lib:utils.bzl", "is_external_label")
+
+_update_source_file_attrs = {
+    "in_file": attr.label(allow_single_file = True, mandatory = True),
+    "update_file": attr.label(allow_single_file = True, mandatory = True),
```
gregmagolan, line 7: lets bikeshed the naming of the rule and attrs here.

Maybe `out_file` is better than `update_file`. Perhaps `write_source_file` is clearer than `update_source_file`?

cc @alexeagle

kormide: I like these suggestions.

kormide:
Some things that Chuck's has that ours doesn't updatesrc:
* exposes the update rule and the rule+diff_test wrapper as two different things
* accepts multiple srcs and outs in the same rule
* accepts `deps` with provider `UpdateSrcsInfo` so that other rules can provide info about sources to be updated
* an executable rule that queries for all sources to be updated and runs them all

Are any of the above things you want to me to add before we land the first iteration?

cgrindel:
Thanks for sharing the PR with me. I have a few comments:

- Is there functionality missing in [updatesrc](https://github.com/cgrindel/bazel-starlib/tree/main/updatesrc) that would prevent you from just using that instead of reimplementing this functionality? If so, I would be glad to take contributions.
- As a bit of background, [updatesrc](https://github.com/cgrindel/bazel-starlib/tree/main/updatesrc) provides two ways to provide source-output mappings because I have source formatting rules that perform their operations and then passes along the provider with the source-output mappings. The conventional `srcs` and `outputs` are for more declarative use cases.
- The `update_all` functionality in [updatesrc](https://github.com/cgrindel/bazel-starlib/tree/main/updatesrc) is nice. The user does not need to hunt down the update command to run to get things green. Just run the `update_all` and then execute your tests again.
- I do not see any tests in this PR. Are they in a follow-on PR?

lib/write_source_file.bzl
```diff
@@ -0,0 +1,122 @@
...
+    if kwargs.pop("file2", None) != None:
+        fail("file2 not a valid parameter in write_source_file")
+    if kwargs.pop("failure_message", None) != None:
+        fail("failure_message not a valid parameter in write_source_file")
+
+    name_test = "%s_test" % name
+
+    if out_file_missing:
+        # Stamp out a test that fails with a helpful message when the source file doesn't exist
+        fail_with_message_test(
```
gregmagolan, line 87: Try `fail(msg)` here instead of a rule

kormide: That's right. Run fails if I just put a `fail()` here.

kormide:
Thinking about how to write a test for this... would the cleanest way be to write a `sh_test` integration test that sets up a workspace and the `write_source_file` rule, then calls `bazel run` on it and checks the source tree?

cgrindel:
For `updatesrc`, I did something similar to what you are suggesting. However, I am using [rules_bazel_integration_test](https://github.com/cgrindel/rules_bazel_integration_test) to exercise the rules/macros in a child workspace. You can see the types of tests that I added [here](https://github.com/cgrindel/bazel-starlib/tree/main/tests/updatesrc_tests/workspace).

alexeagle:
FWIW I think we can totally take a dependency from aspect_bazel_lib to cgrindel/rules_bazel_integration_test because either of
- it's an internal dependency not exposed to users so they don't need to change WORKSPACE
- that repo is going to be on bazelbuild org soon once they give us the keys

kormide:
Updated with a test (thanks @cgrindel!)

lib/write_source_files.bzl
```diff
@@ -0,0 +1,128 @@
...
+    bazel test //...
+
+    //:foobar.json is out-of-date. To update this file, run:
+
+        bazel run //:write_foobar
+    ```
+
+    Args:
+        name: Name of the executable target that creates or updates the source file
+        files: A dict where the keys are source files to write to and the values are labels pointing to the desired content
```
gregmagolan, line 55: Can you add to the docstring on files that "Source files must be within the same bazel package as the target."?

lib/write_source_files.bzl
```diff
@@ -0,0 +1,128 @@
...
+        fail("failure_message not a valid parameter in write_source_file")
+
+    for i in range(len(out_files)):
+        out_file = _to_label(out_files[i])
+        out_file_missing = _is_file_missing(out_file)
+
+        name_test = "%s_%d_test" % (name, i)
+
+        if out_file_missing:
+            # Stamp out a test that fails with a helpful message when the source file doesn't exist
```
gregmagolan, line 90: Maybe mention here that a simple `fail(msg)` would not work since that would fail during analysis and prevent the user from calling `bazel run //update/the:file`

gregmagolan:
> 

We can punt on bazel-in-bazel testing for now. We can test the updater binary without that and the rule itself & macro it simple enough that we can make do with a smoke test for now of an instance of it.

gregmagolan:
> FWIW I think we can totally take a dependency from aspect_bazel_lib to cgrindel/rules_bazel_integration_test because either of
> 
> * it's an internal dependency not exposed to users so they don't need to change WORKSPACE
> * that repo is going to be on bazelbuild org soon once they give us the keys

We can punt on bazel-in-bazel testing for now. The updater binary can be tested without that and the macro is simple enough that we can make do with a simple smoke test instance of it.

lib/write_source_file.bzl
gregmagolan: 👍

kormide:
Updated

gregmagolan (approved):
🚀

PR #31, opened 2022-02-04
docs: add a directory to the readme

kormide:
@gregmagolan I figured it would be nice to give a quick preview of what we offer to visitors to our github page. Wdyt?


gregmagolan:
It's a good idea. My concern is that it will fall out of date quickly unless there it is automated. Lets think about this a bit more.

PR #35, opened 2022-02-11
fix: fix buildifier 5.0.0 errors

kormide:
@alexeagle 

See: https://buildkite.com/bazel/bazel-lib/builds/4#9229dcdf-05e4-415c-bd93-9cc028fa8582

### DIFF
--- a/lib/private/paths.bzl
+++ b/lib/private/paths.bzl
@@ -65,7 +65,7 @@ def _to_manifest_path(ctx, file):
     else:
         return ctx.workspace_name + "/" + file.short_path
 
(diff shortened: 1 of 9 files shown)


lib/private/paths.bzl
```diff
@@ -65,15 +65,14 @@ def _to_manifest_path(ctx, file):
     else:
         return ctx.workspace_name + "/" + file.short_path
 
-def _to_workspace_path(ctx, file):
+def _to_workspace_path(file):
```
kormide, line 68: Not used.

alexeagle (approved):

PR #37, opened 2022-02-17
feat: write_source_files windows implementation

kormide:
Not ready to merge. Still resolving issues with the test rule implementation.


kormide:
@gregmagolan The rule implementation works but not the test rule. Might need to pair next week.

kormide:
Updated with several totally intuitive windows workarounds.

lib/private/write_source_files.bzl
```diff
@@ -54,6 +37,68 @@ chmod 644 "$out"
...
 
+    return updater
+
+def _write_source_files_bat(ctx):
+    updater = ctx.actions.declare_file(
+        ctx.label.name + "_update.bat",
+    )
+
+    content = """
+@rem Generated by copy_to_directory.bzl, do not edit.
```
kormide, line 48: TODO: fix comment

lib/tests/write_source_files/write_source_files_test.bzl
```diff
@@ -55,9 +55,80 @@ assert_same() {
...
+
+def _impl_bat(ctx):
+    test = ctx.actions.declare_file(
+        ctx.label.name + "_test.bat",
+    )
+
+    # BUG: Windows is having trouble executing the symlink to the write_source_files executable,
+    # but it is able to execute the actual output script so point to that for now.
+    # write_source_files = ctx.executable.write_source_files_target.short_path.replace("/", "\\")
+    write_source_files = "..\\..\\%s" % ctx.executable.write_source_files_target.basename
```
kormide, line 68: This gets out of the runfiles dir into the output dir.

lib/private/write_source_files.bzl
```diff
@@ -54,6 +37,68 @@ chmod 644 "$out"
...
+set runfiles_dir=%cd%
+if defined BUILD_WORKSPACE_DIRECTORY (
+    cd %BUILD_WORKSPACE_DIRECTORY%
+)
+""" + "\n".join(["""
+set in=%runfiles_dir%\\{in_file}
+set out={out_file}
+
+if not defined BUILD_WORKSPACE_DIRECTORY (
+    @rem Weird windows issue: when this script run in a test rule, the out file
```
gregmagolan, line 59: Yeah. This is because on Windows you're outside of the sandbox by default so you're in the actual runfiles folder in the execroot where the outfile is a symlink to the actual outfile in the execroot which is a symlink to the source file.

lib/tests/write_source_files/write_source_files_test.bzl
gregmagolan: How does this end up being two levels up + the basename? Can you add paths examples in the comment here for clarity?

lib/tests/write_source_files/write_source_files_test.bzl
```diff
@@ -66,7 +137,9 @@ _write_source_files_test = rule(
         "write_source_files_target": attr.label(
             allow_single_file = True,
             executable = True,
-            cfg = "exec",
+            # Should be cfg = "exec" but a bazel bug causes a wrong executable symlink on windows
+            cfg = "target",
```
gregmagolan, line 141: With the `../../` fix above, will this now with with `cfg = "exec"` ?

lib/private/write_source_files.bzl
```diff
@@ -8,24 +8,7 @@ _write_source_files_attrs = {
...
-        if is_external_label(out_file_label):
-            fail("out file %s must be a source file in the user workspace" % out_file_label)
-
-        if not ctx.files.out_files[i].is_source:
-            fail("out file %s must be a source file, not a generated file" % out_file_label)
-
-        if out_file_label.package != ctx.label.package:
-            fail("out file %s (in package '%s') must be a source file within the target's package: '%s'" % (out_file_label, out_file_label.package, ctx.label.package))
-
+def _write_source_files_sh(ctx):
```
gregmagolan, line 11: Should add `is_executable = True,` to the `ctx.actions.write` for the bash case as well

lib/tests/write_source_files/write_source_files_test.bzl
kormide: This goes from the execution directory in runfiles `some-runfiles-dir/aspect-bazel-lib` to the location of the script in the output dir. The runfiles dir is always in the same directory as the output script, right?

Yeah I can add an example for clarity.

lib/private/write_source_files.bzl
kormide: Ah, so it's not unexpected behaviour? I'll update the comment.

gregmagolan: You could probably repro on macos with `--spawn_stragegy=local` which turns off the sandbox

lib/tests/write_source_files/write_source_files_test.bzl
gregmagolan: Ahh yes. That is correct. The runfiles dir is created in the the package dir in the output tree of where the binary target is.

You could determine `../..` programatically using the same pattern as https://github.com/aspect-build/bazel-lib/blob/main/lib/private/utils.bzl#L85 but using ctx.package()

gregmagolan:
```
"\\".join([".."] * len(ctx.package().split("/")))
```

kormide: I don't think that would be correct since we need to back out of two directories (`runfiles_dir/repo name`) but the package name has more components.

kormide: It's not the location of the script inside of the runfiles dir that we're cancelling, but the cwd when that script is executing.

lib/tests/write_source_files/write_source_files_test.bzl
kormide: It's still needed unfortunately.

kormide:
Updated. Could you do another pass over my comments to double check that they're accurate?

lib/tests/write_source_files/write_source_files_test.bzl
gregmagolan: Oh right. You're cwd is the root of the repo and not the package. Nevermind :)

gregmagolan (approved):
🌮

PR #40, opened 2022-02-26
feat: bulk updates for write_source_files

kormide:
@gregmagolan @alexeagle 

`write_source_files` now accepts `deps` on other write_source_files targets so it can be used recursively to create hierarchies of update groups, or just a bunch of targets for individual files and one master update at the root.

### DIFF
--- a/docs/write_source_files.md
+++ b/docs/write_source_files.md
@@ -7,7 +7,7 @@ Public API for write_source_files
 ## write_source_files
 
 <pre>
-write_source_files(<a href="#write_source_files-name">name</a>, <a href="#write_source_files-files">files</a>, <a href="#write_source_files-kwargs">kwargs</a>)
+write_source_files(<a href="#write_source_files-name">name</a>, <a href="#write_source_files-files">files</a>, <a href="#write_source_files-additional_update_targets">additional_update_targets</a>, <a href="#write_source_files-suggested_update_target">suggested_update_target</a>, <a href="#write_source_files-kwargs">kwargs</a>)
 </pre>
 
 Write to one or more files in the source tree. Stamp out tests that ensure the files exists and are up to date.
@@ -30,23 +30,54 @@ To update the source file, run:
 bazel run //:write_foobar
 ```
 
-A test will fail if the source file doesn't exist
-```bash
-bazel test //...
+A test will fail if the source file doesn't exist or if it's out of date with instructions on how to create/update it.
 
-//:foobar.json does not exist. To create & update this file, run:
+You can declare a tree of generated source file targets:
 
-    bazel run //:write_foobar
+```starlark
+load("@aspect_bazel_lib//lib:write_source_files.bzl", "write_source_files")
+
+write_source_files(
+    name = "write_all",
+    additional_update_targets = [
+        # Other write_source_files targets to run when this target is run
+        "//a/b/c:write_foo",
+        "//a/b:write_bar",
+    ]
+)
 ```
 
-...or if it's out of date.
+And update them with a single run:
+
 ```bash
-bazel test //...
-
-//:foobar.json is out-of-date. To update this file, run:
-
-    bazel run //:write_foobar
+bazel run //:write_all
 ```
+
+When a file is out of date, you can leave a suggestion to run a target further up in the tree by specifying `suggested_update_target`. E.g.,
+
+```starlark
+write_source_files(
+    name = "write_foo",
+    files = {
+        "foo.json": ":generated-foo",
+    },
+    suggested_update_target = "//:write_all"
+)
+```
+
+A test failure from foo.json being out of date will yield the following message:
+
+```
+//a/b:c:foo.json is out of date. To update this and other generated files, run:
+
+    bazel run //:write_all
+
+To update *only* this file, run:
+
+    bazel run //a/b/c:write_foo
+```
+
+If you have many sources that you want to update as a group, we recommend wrapping write_source_files in a macro that defaults `suggested_update_target` to the umbrella update target.
 
 
 **PARAMETERS**
@@ -56,6 +87,8 @@ bazel test //...
 | :------------- | :------------- | :------------- |
 | <a id="write_source_files-name"></a>name |  Name of the executable target that creates or updates the source file   |  none |
 | <a id="write_source_files-files"></a>files |  A dict where the keys are source files to write to and the values are labels pointing to the desired content. Source files must be within the same bazel package as the target.   |  none |
+| <a id="write_source_files-additional_update_targets"></a>additional_update_targets |  (Optional) List of other write_source_files targets to update in the same run   |  <code>[]</code> |
+| <a id="write_source_files-suggested_update_target"></a>suggested_update_target |  (Optional) Label of the write_source_files target to suggest running when files are out of date   |  <code>None</code> |
 | <a id="write_source_files-kwargs"></a>kwargs |  Other common named parameters such as <code>tags</code> or <code>visibility</code>   |  none |
 
 
--- a/lib/private/write_source_files.bzl
+++ b/lib/private/write_source_files.bzl
@@ -2,9 +2,17 @@
 
 load("//lib:utils.bzl", "is_external_label")
 
+_WriteSourceFilesInfo = provider(
+    "Provider to enforce deps are other write_source_files targets",
+    fields = {
+        "executable": "Generated update script",
+    },
+)
+
 _write_source_files_attrs = {
-    "in_files": attr.label_list(allow_files = True, allow_empty = False, mandatory = True),
-    "out_files": attr.label_list(allow_files = True, allow_empty = False, mandatory = True),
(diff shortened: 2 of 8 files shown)


docs/write_source_files.md
```diff
@@ -30,23 +30,54 @@ To update the source file, run:
 bazel run //:write_foobar
 ```
 
-A test will fail if the source file doesn't exist
-```bash
-bazel test //...
+A test will fail if the source file doesn't exist or if it's out of date with instructions on how to create/update it.
+
+You can declare generated source files hierachically:
```
alexeagle, line 35: hierarchically spelled wrong, but also it's better for documentation to use simpler language where possible (e.g. to be inclusive for many users whose english is not their first language)

"you can declare a tree of generated source file targets"?

docs/write_source_files.md
```diff
@@ -30,23 +30,54 @@ To update the source file, run:
...
-//:foobar.json is out-of-date. To update this file, run:
+You can point the out-of-date error test messages to the umbrella update target with by specifying `update_target`. E.g.,
+
+```starlark
+write_source_files(
+    name = "write_foo",
+    files = {
+        "foo.json": ":generated-foo",
+    },
+    update_target = "//:write_all"
```
alexeagle, line 64: it's too bad you have a conceptual cycle in the dependency graph here.

also, what's the user experience?
- only one write_source_files is out-of-date, so it doesn't matter which updater command I run. Then this update_target isn't necessary and is just one more line of BUILD file to maintain
- more than one write_source_files is out-of-date, then each one that fails prints another copy of the write_all updater command

in the second case it would be nice to get only one test failure, is that possible?

lib/write_source_files.bzl
```diff
@@ -81,41 +115,74 @@ def write_source_files(name, files, **kwargs):
     if kwargs.pop("failure_message", None) != None:
         fail("failure_message not a valid parameter in write_source_file")
 
+    # if update_target == None:
```
alexeagle, line 118: remove?

lib/write_source_files.bzl
```diff
@@ -32,28 +32,61 @@ def write_source_files(name, files, **kwargs):
...
     ```
 
+    If you have many sources that you want to update as a group, we recommend wrapping write_source_files in a macro that defaults `update_target` to the umbrella update target.
+
     Args:
         name: Name of the executable target that creates or updates the source file
         files: A dict where the keys are source files to write to and the values are labels pointing to the desired content.
             Source files must be within the same bazel package as the target.
+        deps: (Optional) List of other write_source_files targets to update in the same run
+        update_target: (Optional) String label of the write_source_files target to suggest running when files are out of date
```
alexeagle, line 89: since it's a string instead of a label attribute, there's no checking of whether the user provided a valid value huh?

alexeagle (commented):
cool

docs/write_source_files.md
gregmagolan: the tests are all independent diff tests and we're not chaining those into a graph here so not possible with the current shape; hence the conceptual cycle. tho I think it is better to have it than not as it will teach users that there is only one command they have to call regardless of which file needs updating

lib/write_source_files.bzl
gregmagolan: hmmm. we should call `to_label(update_target)` and it will at least validate the it is either a `string` type of a `Label` type. That util function could do some more error handling as well to check if the string is an invalid label.. but it won't check if the label exists at the macro level.

lib/private/write_source_files.bzl
```diff
@@ -2,9 +2,17 @@
...
+        "executable": "Generated update script",
+    },
+)
+
 _write_source_files_attrs = {
-    "in_files": attr.label_list(allow_files = True, allow_empty = False, mandatory = True),
-    "out_files": attr.label_list(allow_files = True, allow_empty = False, mandatory = True),
+    "in_files": attr.label_list(allow_files = True, allow_empty = False, mandatory = False),
+    "out_files": attr.label_list(allow_files = True, allow_empty = False, mandatory = False),
+    "deps": attr.label_list(allow_files = False, providers = [_WriteSourceFilesInfo], mandatory = False),
```
gregmagolan, line 15: I was humming and hawing about the `deps` naming over the weekend. `deps` seems like the wrong name as this target technically doesn't depend on the targets in deps; we're just chaining the update targets together here.

Perhaps `additional_update_targets` or `also_update_targets`?

lib/write_source_files.bzl
```diff
@@ -11,7 +11,7 @@ _write_source_files = rule(
     executable = True,
 )
 
-def write_source_files(name, files, **kwargs):
+def write_source_files(name, files, deps = [], update_target = None, **kwargs):
```
gregmagolan, line 14: perhaps `suggested_update_target` to be clearer that the attribute is load bearing for the suggestion only?

gregmagolan:
this looks great; just some bike shedding on the attribute names needed I think so the API is clearer for users who don't read the docstrings

docs/write_source_files.md
kormide: @alexeagle the user experience if someone runs a `bazel test //...` is:

`X file is out of date. Run Y to update it` (for each file that's out of date)

If the developer doesn't know about the parent write_source_files, they'd have to update those one by one with a different bazel run command. With the label that creates a conceptual cycle, the user would see:

`X files is out of date. Run Y to update it. Or, run Z to update all the files.`

kormide: The second case you described would be possibly if we created separate rules, I think. One for declaring the source files (maybe that rule just outputs a provider) and one for actually stamping the tests and running the generation on some group of those source declarations. Though we wouldn't get test stamps for the individual files..

kormide: We might be able to do all of that with one rule if we dropped the requirement that the declaration of the source file has to be in the same directory as the rule. So you could have...


```
write_source_files(
  name = "write_many",
  files = {
     "foo/bar/a.json": ":a",
     "b.json": ":b"
  }
)
```

And that could stamp a single test for as many sources as you want.

gregmagolan: the glob trick to check for file existence wouldn't work if the source files are not in the same package; also a bit of an anti-pattern in bazel to write to files not in the same package as the target

lib/private/write_source_files.bzl
```diff
@@ -99,10 +116,19 @@ def _write_source_files_impl(ctx):
     else:
         updater = _write_source_files_sh(ctx)
 
-    return DefaultInfo(
-        executable = updater,
-        runfiles = ctx.runfiles(files = ctx.files.in_files),
-    )
+    runfiles = ctx.runfiles(files = ctx.files.in_files)
+    for target in ctx.attr.additional_update_targets:
+        runfiles = runfiles.merge(target[DefaultInfo].default_runfiles)
```
kormide, line 121: @gregmagolan Could you double check that I'm doing this right? First time forming up runfiles manually.

lib/write_source_files.bzl
```diff
@@ -32,28 +32,61 @@ def write_source_files(name, files, **kwargs):
...
 
-    //:foobar.json does not exist. To create & update this file, run:
+    ```starlark
+    load("@aspect_bazel_lib//lib:write_source_files.bzl", "write_source_files")
 
-        bazel run //:write_foobar
+    write_source_files(
+        name = "write_all",
+        additional_update_targets = [
+            # Other write_source_files targets
```
gregmagolan, line 45: nit: "Other write_source_files targets to run when this target is run"

lib/write_source_files.bzl
```diff
@@ -32,28 +32,61 @@ def write_source_files(name, files, **kwargs):
...
+    ```
+
+    If you have many sources that you want to update as a group, we recommend wrapping write_source_files in a macro that defaults `suggested_update_target` to the umbrella update target.
+
     Args:
         name: Name of the executable target that creates or updates the source file
         files: A dict where the keys are source files to write to and the values are labels pointing to the desired content.
             Source files must be within the same bazel package as the target.
+        additional_update_targets: (Optional) List of other write_source_files targets to update in the same run
+        suggested_update_target: (Optional) String label of the write_source_files target to suggest running when files are out of date
```
gregmagolan, line 89: nit: I'd say just "Label" instead of "String label". Its implies that you can pass a string that represents a label.

lib/private/write_source_files.bzl
gregmagolan: Looks right. You can add an optimization using "merge_all" which is available in newer Bazel releases

see `js_library` impl:
https://github.com/bazelbuild/rules_nodejs/blob/stable/internal/js_library/js_library.bzl#L244
```
    # Perf optimization available in newer Bazel releases
    if "merge_all" in dir(runfiles):
        runfiles = runfiles.merge_all(deps_runfiles)
    else:
        for d in deps_runfiles:
            runfiles = runfiles.merge(d)
```

gregmagolan (approved):
🌮

PR #44, opened 2022-03-02
experimental: write source files from depset

kormide:
DO NOT MERGE

Decided to try this out. The writing to source files works: it will take each file in the depset and pop it into the source directory. But we can't stamp out missing/out-of-date tests when we don't know what the output files are..

But if we're committing to source we should know what the expected files are after the first time we run the generation. Maybe what we really need is the ability to do something like:

```
files = {
  "some-file.js": struct(
      target = "//foo/bar:some-group",
      take_file: "some-file.js",
   )
}
```
Then we could stamp out our tests.

Wdyt?

### DIFF
--- a/lib/tests/write_source_files/BUILD.bazel
+++ b/lib/tests/write_source_files/BUILD.bazel
@@ -15,6 +15,20 @@ genrule(
 )
 
 genrule(
+    name = "g-desired",
+    outs = ["g/g.js"],  # Generated in a subdirectory
+    cmd = "echo 'console.log(\"g*\")' > $@",
+)
+
+filegroup(
+    name = "fg-group",
+    srcs = [
+        ":g-desired",
+        "//lib/tests/write_source_files/subdir:f-desired",
+    ],
+)
+
+genrule(
     name = "e",
     outs = ["e.js"],
     cmd = "echo 'console.log(\"e*\")' > $@",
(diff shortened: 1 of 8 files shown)


lib/tests/write_source_files/BUILD.bazel
```diff
@@ -47,4 +61,7 @@ write_source_files(
         "b2.js": ":b-desired",
         "e_dir": ":e_dir-desired",
     },
+    include_files = [
+        ":fg-group",
+    ],
 )
```
kormide, line 67: This is the interesting part of this PR.

alexeagle:
did you figure out what stackb/rules_proto does to handle this? I guess it always knows what proto files are being written and has labels declared for them?

kormide:
> did you figure out what stackb/rules_proto does to handle this? I guess it always knows what proto files are being written and has labels declared for them?

Yeah, the sources are passed in as labels. The test rule `proto_compile_gencopy_test` then tests that there's a corresponding output file in the source tree. So it's using knowledge of how proto_compile works to determine which file to test for.

kormide:
So if the user knows that `foo.js` should be generated, but it's somewhere in `:foo-outputs`, I think we would have to write our own diff_test to be able to look for some file inside of `:foo-outputs`.

PR #45, opened 2022-03-03
reimplement stardoc_with_diff_test to use write_source_files

kormide:
Messages look like 
```
//docs:expand_make_vars.md is out of date. To update this and other generated files, run:

    bazel run //docs:update

To update *only* this file, run:

    bazel run //docs:docs_expand_make_vars-update
```

Problems:
* breaking change, users have to declare the `out_labels` in the root update rule, which isn't as user friendly
* we default the root update rule name to `name="update"`, so for the suggested update rule to be accurate in the error messages we either need to let the `stardoc_with_diff_test` rules override that target or disallow setting a `name` on `update_docs`. This is of course only if they don't use the default `name="update"`.

If you like this approach I can clean things up and update the documentation. Personally, I don't think this creates a better DX. Wdyt?


alexeagle:
my main issue with `stardoc_with_diff_test` is that it added more path nesting, which appears as a stutter `docs/docs` on the docsite URLs
https://docs.aspect.build/aspect-build/bazel-lib/v0.5.0/docs/docs/write_source_files-docgen.html

I know that's separate from your questions. The current thing uses the magic `native.existing_rules` to discover the grouping. Maybe you can do that in write_source_files, or maybe the `stardoc_files` can be a macro that just wraps the write_source_files with the existing_rules discovery?

alexeagle:
> couldn't find a way to deduce the full labels
This still feels like the thing to try one more time on. Maybe we could pair on it? I don't understand the problem enough to say whether it ought to be solvable or not.

kormide:
Yeah pairing on this will be helpful.

kormide:
Sorry to move ahead on this before pairing, but I found a wait to get it working with `native.existing_rules()` as before.

lib/private/docs.bzl
```diff
@@ -1,60 +1,59 @@
...
+        name = stardoc_label,
+        out = name + "-docgen.md",
         input = bzl_library_target + ".bzl",
         deps = [bzl_library_target],
         aspect_template = aspect_template,
         func_template = func_template,
         header_template = header_template,
         provider_template = provider_template,
         rule_template = rule_template,
+        tags = ["package:" + native.package_name()],  # Tag the package name which will help us reconstruct the write_source_files label in update_docs
```
kormide, line 44: By attaching the package name in a tag, I can now reconstruct the full label for all of `write_source_file` rules and use that in the update all macro. The issue I was having before was that `native.existing_rules()` only gives the name of the rule, but not the package it's in.

lib/private/write_source_files.bzl
```diff
@@ -10,8 +10,8 @@ _WriteSourceFilesInfo = provider(
 )
 
 _write_source_files_attrs = {
-    "in_files": attr.label_list(allow_files = True, allow_empty = False, mandatory = False),
-    "out_files": attr.label_list(allow_files = True, allow_empty = False, mandatory = False),
+    "in_files": attr.label_list(allow_files = True, allow_empty = True, mandatory = False),
```
alexeagle, line 13: this seems surprising, why run the target with no inputs? maybe instead the macro just shouldn't stamp out a write_source_files in this case, and do something else to make it a no-op

lib/private/docs.bzl
```diff
@@ -70,29 +69,19 @@ def update_docs(
...
-            content.append("cp -fv bazel-bin/{0}/{1} {2}".format(docs_folder, doc_gen, doc_dest))
-
-    update_script = name + ".sh"
-    write_file(
-        name = "gen_" + name,
-        out = update_script,
-        content = content,
-    )
+            for tag in r["tags"]:
+                if tag.startswith("package:"):
```
alexeagle, line 79: yeah I'm okay with this since it's hidden inside the macro.

alexeagle (approved):

lib/private/write_source_files.bzl
kormide: I think it's a legit case when it's just for grouping of child `write_source_files` targets and doesn't specify any source files itself. Those are declared in `additional_update_targets`.

kormide:
I want to improve the docs a bit before we merge.

PR #52, opened 2022-03-26
feat: add diff_test attr to write_source_file(s) for .gitignored files

gregmagolan:
Adds a `diff_test` attribute that can be set to false for the `.gitignored` file case.

Also fixes `write_source_file` so that out_file is a `attr.string()` instead of a `attr.label()`so that it does not show up in the source files of the target when doing a source file query such as 

```
bazel query 'kind("source file", deps(//lib/tests/write_source_files:write_dist))'
```

This prevents ibazel from watching the source file as an output file and going into an infinite loop of update and notify.

Usage:

```
genrule(
    name = "dist",
    outs = ["dist.js"],
    cmd = "echo 'dist' > $@",
)

# ibazel run //lib/tests/write_source_files:write_dist
write_source_files(
    name = "write_dist",
    diff_test = False,
    files = {"dist.js": ":dist"},
)
```


lib/write_source_files.bzl
```diff
@@ -98,47 +99,99 @@ def write_source_files(name, files = {}, additional_update_targets = [], suggest
...
+            suggested_update_target = suggested_update_target,
+            diff_test = False,
+            **kwargs
+        )
 
-        if single_update_target:
-            test_target_name = "%s_test" % name
-        else:
-            test_target_name = "%s_%d_test" % (name, i)
+def write_source_file(
```
kormide, line 128: Is there a reason we need to expose this publicly alongside the plural version?

lib/ibazel_runner.bzl
```diff
@@ -0,0 +1,31 @@
...
+)
+
+def ibazel_runner(
+        name,
+        run_targets,
+        **kwargs):
+    """TODO
+
+    Args:
+        name: Name of the executable target that creates or updates the source file
```
kormide, line 21: copy pasta?

kormide:
Could I get some more context on what the `ibazel_runner` is for?

lib/write_source_files.bzl
gregmagolan: Don't have a strong preference either way. It may be useful but it is a simpler API and docs if we only the plural is public. I'll switch it up.

lib/ibazel_runner.bzl
gregmagolan: Yeah. Ran out of time last night. Cleaned up now.

lib/write_source_files.bzl
gregmagolan: Done.

gregmagolan:
Cleaned up and ready for final review

lib/write_source_files.bzl
```diff
@@ -90,6 +87,7 @@ def write_source_files(name, files = {}, additional_update_targets = [], suggest
             Sources must be within the same bazel package as the target.
         additional_update_targets: (Optional) List of other write_source_file or other executable updater targets to call in the same run
         suggested_update_target: (Optional) Label of the write_source_file target to suggest running when files are out of date
+        diff_test: (Optional) Generate a test target to check that the source file(s) exist and are up to date with the generated files(s).
```
jbedard, line 90: Should we say that it's enabled by default? Most optional args are opt-in where this is opt-out..?

gregmagolan:
> Could I get some more context on what the `ibazel_runner` is for?

Oh. I thought it was needed but turns out it wasn't so it's removed now

lib/write_source_files.bzl
gregmagolan: The doc generator show the default but I think you're right, the (optional) in the doc string is confusing. Will remove

kormide (approved):
👍

lib/write_source_files.bzl
kormide: Yeah I think it's cleaner with just the one api.

lib/private/write_source_file.bzl
```diff
@@ -180,7 +295,19 @@ def _write_source_file_impl(ctx):
...
+
+def _is_file_missing(label):
+    """Check if a file is missing by passing its relative path through a glob()
+
+    Args
+        label: the file's label
+    """
+    file_abs = "%s/%s" % (label.package, label.name)
+    file_rel = file_abs[len(native.package_name()) + 1:]
+    file_glob = native.glob([file_rel], exclude_directories = 0)
```
jbedard, line 312: Could `file_rel` not have glob characters in it? idk if there's any easy way to escape it though?

jbedard: Looks like this was just moved though 🤷

I don't know the code so it's a bit hard to tell what's new vs changed vs moved...

lib/private/write_source_file.bzl
```diff
@@ -123,13 +247,9 @@ def _write_source_file_impl(ctx):
...
             fail("in_file must be specified if out_file is set")
-        if is_external_label(ctx.attr.out_file.label):
-            fail("out file %s must be in the user workspace" % ctx.attr.out_file.label)
-        if ctx.attr.out_file.label.package != ctx.label.package:
-            fail("out file %s (in package '%s') must be a source file within the target's package: '%s'" % (ctx.attr.out_file.label, ctx.attr.out_file.label.package, ctx.label.package))
 
-    if ctx.attr.in_file and not ctx.attr.out_file:
-        if not ctx.attr.in_file:
+    if ctx.attr.in_file:
+        if not ctx.attr.out_file:
```
jbedard, line 252: Was probably like this before (?) but why nested `if`s instead of a single one?

jbedard: Looks like we do a mix of the two throughout the PR 🤷

jbedard (approved):

lib/private/write_source_file.bzl
gregmagolan: Oh right. It could be collapsed.

Looks like I had made a typo in the original
```
    if ctx.attr.in_file and not ctx.attr.out_file:
        if not ctx.attr.in_file:
```

That nested if would never have been hit before.

PR #53, opened 2022-03-30
fix(write_source_files): fix writing to workspace root

jbedard:


BUILD.bazel
```diff
@@ -20,3 +20,16 @@ bzl_library(
...
+
+# write_source_files() to a git ignored subdirectory of the root
+load("//lib:write_source_files.bzl", "write_source_files")
+genrule(
+    name = "write_source_file_root",
+    outs = ["write_source_file-root_directory"],
+    cmd = "mkdir -p $@ && echo 'test' > $@/test.txt",
+    visibility = ["//visibility:private"],
+)
+write_source_files(
```
jbedard, line 32: @gregmagolan did you actually get your `write_dist` running as a test? This only failed previously when doing `bazel run` and not `bazel test`...

BUILD.bazel
```diff
@@ -20,3 +20,17 @@ bzl_library(
...
+    name = "write_source_file_root-test",
+    diff_test = False,
+    files = {"test-out/dist/write_source_file_root-test": ":write_source_file_root"},
+)
```
kormide, line 36: Want to add `//visibility:private` to this as well?

kormide:
Nothing actually runs this test, right? I'm wondering if we should just leave it out of WORKSPACE.

kormide:
> Nothing actually runs this test, right? I'm wondering if we should just leave it out of WORKSPACE.

Or at least call it out as a manual test that ought to be some sort of e2e test.

jbedard:
This is the same as what @gregmagolan did in `lib/tests/write_source_files:write_dist`. I think we should find a way to actually run them in CI...

BUILD.bazel
jbedard: Shouldn't tests be public?

kormide: Hm, the angle I was gettting at was whether this should be available to anyone who depends on our repository.

kormide:
> This is the same as what @gregmagolan did in `lib/tests/write_source_files:write_dist`. I think we should find a way to actually run them in CI...

That should be pretty easy to do. Just add another run step in `.github/workflows/ci.yaml`.
```yaml
- name: manual integration test
   run: bazel run //:write_source_file_root-test
```

lib/private/write_source_file.bzl
```diff
@@ -266,7 +266,7 @@ def _write_source_file_impl(ctx):
         else:
             fail("in file %s must be a single file or a target that provides DefaultOutputPathInfo or DirectoryPathInfo" % ctx.attr.in_file.label)
 
-        out_path = "/".join([ctx.label.package, ctx.attr.out_file])
+        out_path = "/".join([ctx.label.package, ctx.attr.out_file]) if len(ctx.label.package) > 0 else ctx.attr.out_file
```
gregmagolan, line 269: I believe you can shorten this to
```
out_path = "/".join([ctx.label.package, ctx.attr.out_file]) if ctx.label.package else ctx.attr.out_file
```
empty string is a falsy in starlark

BUILD.bazel
gregmagolan: It is private by default unless there is a default_visiblity in the BUILD to set another default

BUILD.bazel
gregmagolan: Needs an integration test but I'd prefer to avoid the added complexity of bazel-in-bazel

kormide (approved):

PR #54, opened 2022-03-30
test - add bash script tests to CI

jbedard:


kormide:
What do you think about putting this in an `e2e` folder?

kormide:
I find the `.spec.sh` extension a little awkward---makes me think it's a unit test.

kormide:
> > What do you think about putting this in an e2e folder?
> 
> I guess that's a better description? 🤷
> 
> > I find the .spec.sh extension a little awkward---makes me think it's a unit test.
> 
> I'm open to whatever. Could be a subfolder that contains all the tests, or test.sh can be put in the root etc.

Personally prefer separating the e2e tests and putting them in one folder 🤷

.github/workflows/ci.yaml
```diff
@@ -34,3 +34,6 @@ jobs:
           # Bazelisk will download bazel to here
           XDG_CACHE_HOME: ~/.cache/bazel-repo
         run: bazel --bazelrc=.github/workflows/ci.bazelrc --bazelrc=.bazelrc test //...
+      - name: integration tests
+        # Find all '*.spec.*', within the current directory, of type file, execute them, and fail on error (+)
```
kormide, line 38: Comment no longer accurate

kormide (approved):
LGTM

jbedard:
You can see an example with 2 tests running in the other PR: https://github.com/aspect-build/bazel-lib/runs/5763137160

PR #55, opened 2022-03-30
feat: platform_transition_filegroup

alexeagle:
Lifted from https://github.com/aspect-build/gcc-toolchain/pull/8/files
See also https://github.com/fmeum/rules_meta/blob/main/meta/internal/meta.bzl#L4


lib/BUILD.bazel
```diff
@@ -96,7 +90,37 @@ bzl_library(
 
 bzl_library(
     name = "diff_test",
-    srcs = ["diff_test.bzl"],
+    srcs = ["diff_test.bzl"],  # keep
```
kormide, line 93: What's this comment for?

alexeagle: without it, running Gazelle in this repo removes the file.

Gazelle seems to have some (new?) heuristics for avoiding "test code" being included, so files ending with "_test.bzl" are a problem. I had to totally revert the changes Gazelle makes to the lib/private/BUILD.bazel file where it removes a bunch...

kormide (approved):
LGTM structurally, but not familiar with transitions so maybe @gregmagolan should also review.

gregmagolan:
I'll try this at rh for my use case today

gregmagolan:
Hmm. rules_docker transition outputs multiple platforms: https://github.com/bazelbuild/rules_docker/blob/master/container/image.bzl#L787

```
_image_transition = transition(
    implementation = _image_transition_impl,
    inputs = [],
    outputs = [
        "//command_line_option:platforms",
        "@io_bazel_rules_docker//platforms:image_transition_cpu",
        "@io_bazel_rules_docker//platforms:image_transition_os",
    ],
)
```

gregmagolan:
Should this support multiple platforms with `target_platforms`?

alexeagle:
I guess it could support a list of `target_platforms` - wouldn't that mean you transition to a constraint that requires *all* of those platforms? I'm not seeing how that would be useful, since platforms are disjoint/mutually exclusive?

gregmagolan:
Sorted it out for the rules_docker transition use case. This won't be useful there as the transition of just the target_platform is insufficient so I'll need to land a similar rule in rules_docker that uses the transitions there which transitions the target_platform and a few other settings:

```
"Rules for working with transitions."

load("@io_bazel_rules_docker//container:image_transition.bzl", _image_transition = "image_transition")

def _docker_transition_filegroup_impl(ctx):
    files = []
    runfiles = ctx.runfiles()
    for src in ctx.attr.srcs:
        files.append(src[DefaultInfo].files)

    runfiles = runfiles.merge_all([src[DefaultInfo].default_runfiles for src in ctx.attr.srcs])
    return [DefaultInfo(
        files = depset(transitive = files),
        runfiles = runfiles,
    )]

docker_transition_filegroup = rule(
    _docker_transition_filegroup_impl,
    attrs = {
        # Required to Opt-in to the transitions feature.
        "_allowlist_function_transition": attr.label(
            default = "@bazel_tools//tools/allowlists/function_transition_allowlist",
        ),
        "architecture": attr.string(
            doc = "The desired rules_docker CPU architecture to transition to.",
            default = "amd64",
        ),
        "operating_system": attr.string(
            doc = "The desired rules_docker OS to transition to.",
            default = "linux",
        ),
        "srcs": attr.label_list(
            allow_empty = False,
            cfg = _image_transition,
            doc = "The input to be transitioned to the rules_docker platform.",
        ),
    },
    doc = "Transitions the srcs to the rules_docker platform.",
)
```

gregmagolan (approved):

PR #57, opened 2022-04-01
feat: add run_binary with output directory support & improved makevar expansion

gregmagolan:


lib/private/run_binary.bzl
```diff
@@ -0,0 +1,103 @@
...
+                  " directory named the same as the target.",
+        ),
+        "outs": attr.output_list(
+            doc = "Output files generated by the action.<br/><br/>These labels are available for" +
+                  " <code>$(location)</code> expansion in <code>args</code> and <code>env</code>.",
+        ),
+        "args": attr.string_list(
+            doc = "Command line arguments of the binary.<br/><br/>Subject to" +
+                  "<code><a href=\"https://docs.bazel.build/versions/main/be/make-variables.html#location\">$(location)</a></code>" +
+                  " expansion.",
```
kormide, line 100: The html doesn't get rendered in the git preview. Is it supposed to, or is this just for the docsite?

kormide (approved):

docs/copy_file.md
```diff
@@ -0,0 +1,48 @@
...
+
+A rule that copies a file to another place.
+
+native.genrule() is sometimes used to copy files (often wishing to rename them).
+The 'copy_file' rule does this with a simpler interface than genrule.
+
+The rule uses a Bash command on Linux/macOS/non-Windows, and a cmd.exe command
+on Windows (no Bash is required).
+
+This fork of bazel-skylib's copy_file adds directory support.
```
alexeagle, line 11: there was a response on https://github.com/bazelbuild/bazel-skylib/pull/323

I think we should go with `copy_directory` in place of `copy_file(output_dir=True)` so there's some chance of going upstream

lib/private/run_binary.bzl
gregmagolan: Hmm. The html comes from upstream, https://github.com/bazelbuild/bazel-skylib/blob/main/rules/run_binary.bzl#L71. Looks good on the docsite https://docs.aspect.build/bazelbuild/bazel-skylib/1.2.1/docs/run_binary_doc_gen.html#run_binary-tool so I'm going to leave it.

lib/private/expand_make_vars.bzl
```diff
@@ -1,30 +1,9 @@
 "Helpers to expand make variables"
```
alexeagle, line 1: wow we had so much rules_nodejs legacy crap in here that wasn't really needed??

docs/copy_file.md
gregmagolan: Sounds reasonable to me

lib/private/expand_make_vars.bzl
gregmagolan: Yeah. That was some old code from RNJ. We should flip it in RNJ on the next major as well.

PR #61, opened 2022-04-01
update buildifier and fix formatting

kormide:


BUILD.bazel
```diff
@@ -22,13 +23,13 @@ bzl_library(
 )
 
 # write_source_files() to a git ignored subdirectory of the root
-load("//lib:write_source_files.bzl", "write_source_files")
```
jbedard, line 25: Was this done automatically? I did this intentionally to try isolating this test at the bottom, since it's kinda weird having a test for `write_source_files` within the root BUILD :/

jbedard (approved):

BUILD.bazel
kormide: If it's ever not used then buildifier will complain and we can delete the load. You might also be able to disable the buildifier warning with some comment directive. Btw did you install the `pre-commit` for this repo? I think this would have been caught on commit.

PR #63, opened 2022-04-03
feat: add copy_directory

kormide:
@gregmagolan @alexeagle 

https://github.com/aspect-build/bazel-lib/issues/62

https://buildkite.com/bazel/bazel-lib/builds?branch=kormide%3Acopy_directory

Should I remove directory support from `copy_file` if we want to upstream this to skylib?

### DIFF
--- a/docs/BUILD.bazel
+++ b/docs/BUILD.bazel
@@ -8,6 +8,11 @@ stardoc_with_diff_test(
 )
 
 stardoc_with_diff_test(
+    name = "copy_directory",
+    bzl_library_target = "//lib:copy_directory",
+)
+
+stardoc_with_diff_test(
     name = "copy_to_directory",
     bzl_library_target = "//lib:copy_to_directory",
 )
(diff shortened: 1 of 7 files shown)


alexeagle (approved):
great!
Yes I think we should fail in copy_file if the src.is_directory so that forces developers to use copy_directory instead, we don't want the ambiguity and we're still pre-1.0 here so it's okay to make the breaking change.

kormide:
> great! Yes I think we should fail in copy_file if the src.is_directory so that forces developers to use copy_directory instead, we don't want the ambiguity and we're still pre-1.0 here so it's okay to make the breaking change.

Sure, so then our fork of `copy_file` will just be there to support `DirectoryPathInfo`?

alexeagle:
yeah exactly, we're still forced to have a fork of copy_file but maybe we can rely on upstream copy_directory for things like making npm package folder copies from the source tree to bazel-out

PR #65, opened 2022-04-04
fix(write_source_files): fix nested directories

jbedard:


lib/tests/write_source_files/BUILD.bazel
```diff
@@ -132,3 +132,15 @@ write_source_files(
...
+    name = "subdir",
+    outs = ["subdir_test"],
+    cmd = "mkdir -p $@/a/b/c; echo 'test' > $@/a/b/c/test.txt",
+)
+
+write_source_files(
+    name = "write_subdir",
+    diff_test = False,
+    files = {"subdir_test": ":subdir_test"},
+)
```
kormide, line 146: Want to add a comment to connect this to the integration test?

kormide (approved):

lib/private/write_source_file.bzl
```diff
@@ -164,11 +164,12 @@ echo "Copying $in to $out in $PWD"
...
 if [[ -f "$in" ]]; then
     cp -f "$in" "$out"
-    chmod 664 "$out"
+    chmod +w "$out"
 else
+    rm -Rf "$out"/*
     mkdir -p "$out"
-    cp -fR "$in"/* "$out"
-    chmod 664 "$out"/*
+    cp -fRL "$in"/* "$out"
```
kormide, line 171: I think capital `-R` is or will be deprecated on mac? I recall @gregmagolan had a PR out about that.

kormide: What is the `L` for?

jbedard: `L` makes it follow symlinks instead of copy them.

If bazel creates a directory with symlinks to other locations we don't want to just copy the links but copy the content they point to. Otherwise we end up with symlinks into the sandbox copied into the source dir...

jbedard: The rules_nodejs `pkg_web` seems to create these symlinks which is where we saw the problem. I think `pkg_web` creates a directory structure for the package but then the files within those directories are symlinks to the original source of files being "packaged". Basically avoid copying files all over the place and use symlinks as much as possible.

When writing that package into the source dir we want to copy the real files and not the symlinks...

jbedard:
So far I've failed to write a test for:
* the symlink issues (`-L` on `cp`)
* files not copying due to lack of write permission, the e2e test only reproduces files not copying due to lack of execute permission on directories

lib/private/write_source_file.bzl
```diff
@@ -164,11 +164,12 @@ echo "Copying $in to $out in $PWD"
 
 if [[ -f "$in" ]]; then
     cp -f "$in" "$out"
-    chmod 664 "$out"
+    chmod +w "$out"
```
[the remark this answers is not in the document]
kormide: User and group is probably safe?

kormide (approved):
:+1:

lib/private/write_source_file.bzl
gregmagolan: -R is the correct one with consistent meaning on both platforms. It is the lowercase `-r` that has a different meaning on macos vs linux.

gregmagolan (approved):

PR #69, opened 2022-04-12
feat: add copy_to_bin rule

gregmagolan:
Will be useful in rules_js


docs/copy_to_bin.md
```diff
@@ -0,0 +1,38 @@
+<!-- Generated with Stardoc: http://skydoc.bazel.build -->
+
+AA rule that copies source files to the output tree.
```
alexeagle, line 3: double A

docs/copy_to_bin.md
```diff
@@ -0,0 +1,38 @@
...
+
+This rule uses a Bash command (diff) on Linux/macOS/non-Windows, and a cmd.exe
```
alexeagle, line 5: maybe a note about where we copied it from?

alexeagle: there was a thread on bazel-skylib about it with Yun

alexeagle (approved):

kormide (approved):

lib/copy_to_bin.bzl
```diff
@@ -0,0 +1,12 @@
+"""AA rule that copies source files to the output tree.
```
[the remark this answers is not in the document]
gregmagolan: Whoops.

docs/copy_to_bin.md
gregmagolan: Will note that we pulled it from rules_nodejs

PR #74, opened 2022-04-14
fix: make tool of run_binary cfg=exec

gregmagolan:


alexeagle (commented):
Does this match upstream?

gregmagolan:
> Does this match upstream?

Upstream is "host"

kormide (approved):

PR #75, opened 2022-04-14
feat: add create_windows_native_launcher_script to lib/windows_utils.bzl

gregmagolan:
Very useful for rules_js

### DIFF
--- a/lib/windows_utils.bzl
+++ b/lib/windows_utils.bzl
@@ -1,3 +1,17 @@
+# Copyright 2017 The Bazel Authors. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
 "Helpers for rules running on windows"
 
 # cmd.exe function for looking up runfiles.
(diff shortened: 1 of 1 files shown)


lib/windows_utils.bzl
```diff
@@ -1,3 +1,17 @@
+# Copyright 2017 The Bazel Authors. All rights reserved.
```
alexeagle, line 1: Is this intentional?

alexeagle (approved):

lib/windows_utils.bzl
gregmagolan: Yes. It is copied over verbatim from rules_nodejs. Original author was Yun Ping. A few files in this repo that were copied verbatim have this copyright.

lib/windows_utils.bzl
```diff
@@ -50,3 +64,51 @@ exit /b 0
...
+    name = shell_script.basename
+    if name.endswith(".sh"):
+        name = name[:-3]
+    win_launcher = ctx.actions.declare_file(name + ".bat", sibling = shell_script)
+    ctx.actions.write(
+        output = win_launcher,
+        content = r"""@echo off
+SETLOCAL ENABLEEXTENSIONS
+SETLOCAL ENABLEDELAYEDEXPANSION
+set RUNFILES_MANIFEST_ONLY=1
```
iamricard, line 95: 👋🏼 hi and sorry for the necromancy!

This effectively makes the windows launcher not remote executable. Is there a chance you could unset this value or make it configurable (all the way up to js_binary)? We'd also need to be able to change the filepath passed as `bash_bin`. As it is, this makes `js_binary` (and derivatives) unusable on Windows RE setups. 

I don't know if this is documented/discussed in official bazel docs, but our docs on [Windows RE & runfiles](https://docs.engflow.com/re/client/windows.html?h=remote+execution#runfiles) do cover this topic.

gregmagolan: Hi Richard. I don't have a lot of context on this launcher code. Original author was Yun Ping from the Bazel team back in the rules_nodejs days. I'm guessing this line is here because the local Windows case doesn't have runfiles so it assumes the runfiles manifest is present, which doesn't work on RE. Windows RE is not something we have test coverage for and not out top priority at the moment, however, I'd be happy review PRs for improving this code.

PR #81, opened 2022-04-15
chore: fix buildifier errors

kormide:


alexeagle (approved):
Thanks, want to wire up the CI enforcement for our stuff being buildifier-clean?
I haven't decided if it's really worth it for unimportant style nits.

alexeagle:
If so, I think this `buildifier-prebuilt` approach is probably best, since pre-commit.ci doesn't seem to work
https://github.com/thundergolfer/bazel-mypy-integration/pull/50

PR #82, opened 2022-04-22
feat: add os_arch_name() function to repo_utils and cleanup yq & jq toolchains for repository names follow this convention

gregmagolan:
This is so that we can use `os_arch_name(rtcx)` downstream to predict the name of the jq & yq repositories for the host


kormide:
Will wait until https://buildkite.com/bazel/bazel-lib/builds?branch=gregmagolan%3Acleanup_toolchains is green to approve.

alexeagle:
I think this is the wrong approach though - you shouldn't predict a name. Instead the toolchain should expose two "resolved" versions - one for repository rules and one for toolchain resolvers

gregmagolan:
I've made the repository names consistent so they are easier to reason about.

Predictable repository name for the host is now just `jq_%s` % (repo_utils.os_arch_name(rctx)` and `jq_%s` % (repo_utils.os_arch_name(rctx)`

Do you mean adding `@yq_host` and `@jq_host` repositories?

gregmagolan:
We do predict the name in npm_install:
```
    node_repo = "_".join([repository_ctx.attr.node_repository, os_name(repository_ctx)])
```

kormide (approved):

PR #83, opened 2022-04-22
feat: add jq_host and yq_host repositories

gregmagolan:
Even better for upstream repositories rules. They can now use `@yq_host//:yq` and `@jq_host//:jq` as the yq & jq binaries if they need them.

```
$ ls -l $(bazel info output_base)/external | grep jq_
-rw-r--r--   1 greg  wheel    96 21 Apr 18:34 @jq_darwin_amd64.marker
-rw-r--r--   1 greg  wheel    96 21 Apr 18:34 @jq_host.marker
-rw-r--r--   1 greg  wheel    96 21 Apr 18:32 @jq_toolchains.marker
drwxr-xr-x   5 greg  wheel   160 21 Apr 18:34 jq_darwin_amd64
drwxr-xr-x   5 greg  wheel   160 21 Apr 18:34 jq_host
drwxr-xr-x   5 greg  wheel   160 21 Apr 18:32 jq_toolchains
```

```
$ ls -l $(bazel info output_base)/external | grep yq_
-rw-r--r--   1 greg  wheel    96 21 Apr 18:34 @yq_darwin_amd64.marker
-rw-r--r--   1 greg  wheel    96 21 Apr 18:33 @yq_host.marker
-rw-r--r--   1 greg  wheel    96 21 Apr 18:32 @yq_toolchains.marker
drwxr-xr-x   5 greg  wheel   160 21 Apr 18:34 yq_darwin_amd64
drwxr-xr-x   5 greg  wheel   160 21 Apr 18:33 yq_host
drwxr-xr-x   5 greg  wheel   160 21 Apr 18:32 yq_toolchains
```

```
$ bazel run @jq_host//:jq
INFO: Analyzed target @jq_host//:jq (0 packages loaded, 0 targets configured).
INFO: Found 1 target...
Target @jq_darwin_amd64//:jq up-to-date (nothing to build)
INFO: Elapsed time: 0.188s, Critical Path: 0.00s
INFO: 1 process: 1 internal.
INFO: Build completed successfully, 1 total action
INFO: Build completed successfully, 1 total action
jq - commandline JSON processor [version 1.6]

Usage:	/private/var/tmp/_bazel_greg/d429ac343bccdd9460beaa3e7ac46428/external/jq_darwin_amd64/jq [options] <jq filter> [file...]
	/private/var/tmp/_bazel_greg/d429ac343bccdd9460beaa3e7ac46428/external/jq_darwin_amd64/jq [options] --args <jq filter> [strings...]
	/private/var/tmp/_bazel_greg/d429ac343bccdd9460beaa3e7ac46428/external/jq_darwin_amd64/jq [options] --jsonargs <jq filter> [JSON_TEXTS...]

jq is a tool for processing JSON inputs, applying the given filter to
its JSON text inputs and producing the filter's results as JSON on
standard output.

The simplest filter is ., which copies jq's input to its output
unmodified (except for formatting, but note that IEEE754 is used
for number representation internally, with all that that implies).

For more advanced filters see the jq(1) manpage ("man jq")
and/or https://stedolan.github.io/jq

Example:

	$ echo '{"foo": 0}' | jq .
	{
		"foo": 0
	}

For a listing of options, use /private/var/tmp/_bazel_greg/d429ac343bccdd9460beaa3e7ac46428/external/jq_darwin_amd64/jq --help.
```

```
$ bazel run @yq_host//:yq
INFO: Analyzed target @yq_host//:yq (2 packages loaded, 2 targets configured).
INFO: Found 1 target...
Target @yq_darwin_amd64//:yq up-to-date (nothing to build)
INFO: Elapsed time: 0.122s, Critical Path: 0.00s
INFO: 1 process: 1 internal.
INFO: Build completed successfully, 1 total action
INFO: Build completed successfully, 1 total action
Usage:
  yq [flags]
  yq [command]

Examples:

# yq defaults to 'eval' command if no command is specified. See "yq eval --help" for more examples.

# read the "stuff" node from "myfile.yml"
yq '.stuff' < myfile.yml

# update myfile.yml in place
yq -i '.stuff = "foo"' myfile.yml

# print contents of sample.json as idiomatic YAML
yq -P sample.json


Available Commands:
  completion       Generate the autocompletion script for the specified shell
  eval             (default) Apply the expression to each document in each yaml file in sequence
  eval-all         Loads _all_ yaml documents of _all_ yaml files and runs expression once
  help             Help about any command
  shell-completion Generate completion script

Flags:
  -C, --colors                        force print with colors
  -e, --exit-status                   set exit status if there are no matches or null or false is returned
      --expression string             forcibly set the expression argument. Useful when yq argument detection thinks your expression is a file.
      --from-file string              Load expression from specified file.
  -f, --front-matter string           (extract|process) first input as yaml front-matter. Extract will pull out the yaml content, process will run the expression against the yaml content, leaving the remaining data intact
      --header-preprocess             Slurp any header comments and separators before processing expression. (default true)
  -h, --help                          help for yq
  -I, --indent int                    sets indent level for output (default 2)
  -i, --inplace                       update the file inplace of first file given.
  -p, --input-format string           [yaml|y|props|p|xml|x] parse format for input. Note that json is a subset of yaml. (default "yaml")
  -M, --no-colors                     force print with no colors
  -N, --no-doc                        Don't print document separators (---)
  -n, --null-input                    Don't read input, simply evaluate the expression given. Useful for creating docs from scratch.
  -o, --output-format string          [yaml|y|json|j|props|p|xml|x] output format type. (default "yaml")
  -P, --prettyPrint                   pretty print, shorthand for '... style = ""'
  -s, --split-exp string              print each result (or doc) into a file named (exp). [exp] argument must return a string. You can use $index in the expression as the result counter.
      --unwrapScalar                  unwrap scalar, print the value with no quotes, colors or comments (default true)
  -v, --verbose                       verbose mode
  -V, --version                       Print version information and quit
      --xml-attribute-prefix string   prefix for xml attributes (default "+")
      --xml-content-name string       name for xml content (if no attribute name is present). (default "+content")
      --xml-strict-mode               enables strict parsing of XML. See https://pkg.go.dev/encoding/xml for more details.

Use "yq [command] --help" for more information about a command.
```


kormide (approved):
:+1:

alexeagle:
After some discussion in rules_python, @mattem and @f0rmiga and I determined that the name the user supplies, e.g. to `register_yq_toolchains(name = "yq")` ought to be the name of the generated host repository, so it would be `yq` rather than `yq_host`. Otherwise we're not really following the bazel idiom (like a macro, where one generated target should be named the same as the macro) and also introducing a bit of "magic" where the user can't tell what to load from based on the name they supplied.

Can we still change this one?

PR #88, opened 2022-04-27
feat: support bzlmod usage of jq/yq

alexeagle:


alexeagle:
This has to disable our docs generation due to https://github.com/bazelbuild/bazel/issues/14140 

It also has to remove our gazelle setup for maintaining the `bzl_library` targets due to https://github.com/bazelbuild/bazel-skylib/issues/250

alexeagle:
Reduced scope here, instead of trying to switch from WORKSPACE to MODULE.bazel for local development, just added an e2e folder that verifies users can use it as a bzlmod module.

Needs some CI setup in a followup so the e2e is actually getting tested. I verified manually with `cd e2e/bzlmod; bazel test ...`

MODULE.bazel
```diff
@@ -0,0 +1,18 @@
...
+        "@yq_toolchains//:all",
+    ],
+)
+
+bazel_dep(name = "bazel_skylib", version = "1.1.1")
+bazel_dep(name = "platforms", version = "0.0.4")
+
+ext = use_extension("@aspect_bazel_lib//lib:extensions.bzl", "lib")
+use_repo(ext, "jq_toolchains")
+use_repo(ext, "yq_toolchains")
```
kormide, line 18: Do consumers need this to be in our MODULE file? Wouldn't they just include it in theirs to use the toolchains?

lib/extensions.bzl
```diff
@@ -0,0 +1,17 @@
...
+    "register_yq_toolchains",
+)
+
+def _toolchain_extension(_):
+    register_yq_toolchains(version = DEFAULT_YQ_VERSION, register = False)
+    register_jq_toolchains(version = DEFAULT_JQ_VERSION, register = False)
+
+lib = module_extension(
+    implementation = _toolchain_extension,
+)
```
kormide, line 17: Should we add the ability to specify versions through tags?

kormide:
> Reduced scope here, instead of trying to switch from WORKSPACE to MODULE.bazel for local development, just added an e2e folder that verifies users can use it as a bzlmod module.
> 
> Needs some CI setup in a followup so the e2e is actually getting tested. I verified manually with `cd e2e/bzlmod; bazel test ...`

I think you can just add a simple shell script under e2e that does a cd then calls bazel, then the [ci will run it](https://github.com/aspect-build/bazel-lib/blob/main/.github/workflows/ci.yaml#L37).

kormide:
Not for this PR, but I guess we'll need to publish sample MODULE.bazel snippets on release to show how to use the extensions.

kormide:
I think we're missing a bzlmod dependency on stardoc.

lib/extensions.bzl
```diff
@@ -0,0 +1,17 @@
...
+    "DEFAULT_YQ_VERSION",
+    "register_jq_toolchains",
+    "register_yq_toolchains",
+)
+
+def _toolchain_extension(_):
+    register_yq_toolchains(version = DEFAULT_YQ_VERSION, register = False)
+    register_jq_toolchains(version = DEFAULT_JQ_VERSION, register = False)
+
+lib = module_extension(
```
kormide, line 15: Should we call this something else since it gets referenced in downstream consumer's MODULE.bazel? Maybe `ext` or `bazel_lib`?

MODULE.bazel
alexeagle: yes, it needs to be in our file in order for the new e2e to succeed, because we reference it above.

kormide: Can we declare it a dev dependency?

alexeagle: if it's a dev dependency, then the user gets
```
ERROR: /home/alexeagle/Projects/bazel-lib/e2e/bzlmod/BUILD.bazel:5:3: While resolving toolchains for target //:case_no_sources: invalid registered toolchain '@jq_toolchains//:all': @jq_toolchains is not visible from repository `@aspect_bazel_lib`
```

which I think makes sense, when the user evaluates this file it's like all the dev_dependency lines don't exist

alexeagle: In this example, we use the toolchain under our `jq` rule implementation.
If users want to reference the toolchain themselves, then yes they'd need this too.

lib/extensions.bzl
alexeagle: I'm out of time for that right now, I added a TODO here

MODULE.bazel
kormide: I see. So users are forced to download the toolchain even if they don't use `jq`?

lib/extensions.bzl
alexeagle: renamed to `ext`

MODULE.bazel
alexeagle: great question, looking in the `external/` folder under the e2e/bzlmod I see

```
aspect_bazel_lib.ext.jq_linux_amd64
aspect_bazel_lib.ext.jq_toolchains
aspect_bazel_lib.ext.yq_toolchains
aspect_bazel_lib.lib.jq_linux_amd64
aspect_bazel_lib.lib.jq_toolchains
aspect_bazel_lib.lib.yq_toolchains
```

Which is what I hoped - the yq toolchain wasn't used by this example, so Bazel never fetched the binary.

kormide (approved):

e2e/bzlmod/BUILD.bazel
```diff
@@ -1,6 +1,19 @@
 load("@aspect_bazel_lib//lib:jq.bzl", "jq")
+
+# We don't use yq, so we can confirm that Bazel never fetches it.
```
kormide, line 3: How can we confirm that?

kormide (approved):

e2e/bzlmod/BUILD.bazel
alexeagle: good point, added more to this comment about how I checked.

In theory we should have an eager-fetch test like that blog post, but we'd be testing bazel's behavior, not our own.

PR #89, opened 2022-04-28
feat: make it easier to resolve jq/yq toolchains from a repository rule

kormide:
The host repo now just has a symlink to the executable and exports that symlink.


lib/private/jq_toolchain.bzl
```diff
@@ -190,30 +190,22 @@ jq_platform_repo = repository_rule(
...
-""".format(
-        name = repository_ctx.name,
-        host_platform = repo_utils.os_arch_name(repository_ctx),
-        maybe_windows = ".exe" if repo_utils.is_windows_os(repository_ctx) else "",
-    ))
+    repository_ctx.symlink("../../external/{name}_{platform}/jq{ext}".format(
+        name = repository_ctx.attr.name,
+        platform = repo_utils.os_arch_name(repository_ctx),
+        ext = bin_ext,
+    ), "jq{ext}".format(ext = bin_ext))
```
kormide, line 208: I couldn't use `repository_ctx.path` on a label pointing to the platform toolchains. I kept getting a relative path within this repository.

kormide:
Er wait a moment. If we wanted to pass a label instead of a repository name into a repository rule, then the existing aliases should work, right?

kormide:
I played around in rules_js and had `npm_import` take in a label instead of a repository name. It complains about the alias not being a "regular file".
```
    "yq_bin": attr.label(
        default = "@yq//:yq",
    ),
```
So I think we'll need symlinks like I did here. However, a further problem is that if we export the symlink then we need the correct extension on windows. We wouldn't be able to default the above label for windows. In that case could we just make a `filegroup` target named `yq` that only contains the executable (plus extension)?

lib/private/jq_toolchain.bzl
```diff
@@ -190,30 +190,22 @@ jq_platform_repo = repository_rule(
...
     # Base BUILD file for this repository
     repository_ctx.file("BUILD.bazel", """# Generated by @aspect_bazel_lib//lib/private:jq_toolchain.bzl
 package(default_visibility = ["//visibility:public"])
 alias(name = "jq", actual = "@{name}_{platform}//:jq")
-exports_files(["index.bzl"])
+exports_files(["jq{ext}"])
 """.format(
-        name = repository_ctx.name,
-        platform = repo_utils.os_arch_name(repository_ctx),
+        ext = bin_ext,
```
alexeagle, line 201: nit: make the variable name match the template parameter name

alexeagle (approved):

alexeagle:
Windows, meh. I think we could just force the caller to add the `.exe` on windows when they reference the bin label

kormide:
Updated to include the generated `aspect_bazel_lib_host` repo.

kormide:
Removed the host repo rule. I can't add the `@yq//yq` alias alongside the symlink because then the symlink and the directory `yq` would have the same name.

e2e/run_yq_symlinked_bin.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
+
+set -e
+
+bazel run @yq//:yq -- --null-input .a=5
```
alexeagle, line 5: the downside is that this e2e would fail on windows, right?
I think we can resolve that later when we have a windows CI

alexeagle (approved):

e2e/run_yq_symlinked_bin.sh
kormide: Ah yeah, good point. Agreed we can worry about it later.

gregmagolan: You could check the platform in the bash script and call either
```
bazel run @yq//:yq -- --null-input .a=5
```
or 
```
bazel run @yq//:yq.exe -- --null-input .a=5
```

lib/private/yq_toolchain.bzl
```diff
@@ -188,7 +188,7 @@ yq_toolchains_repo = repository_rule(
 )
 
 def _yq_platform_repo_impl(repository_ctx):
-    is_windows = repository_ctx.attr.platform.startswith("windows_")
+    is_windows = repo_utils.is_windows(repository_ctx)
```
gregmagolan, line 191: hah. I made the same mistake earlier. you have to check against `repository_ctx.attr.platform` since `repo_utils.is_windows(repository_ctx)` will return the always return the host platform

lib/private/jq_toolchain.bzl
```diff
@@ -155,7 +155,7 @@ jq_toolchains_repo = repository_rule(
 )
 
 def _jq_platform_repo_impl(repository_ctx):
-    is_windows = repository_ctx.attr.platform.startswith("windows_")
+    is_windows = repo_utils.is_windows(repository_ctx)
```
gregmagolan, line 158: also here

e2e/run_yq_symlinked_bin.sh
gregmagolan: Could use a test for `bazel run @yq//yq` as well once that is added in.

lib/private/yq_toolchain.bzl
kormide: I'm not sure I follow your meaning. What's wrong with this?

e2e/run_yq_symlinked_bin.sh
kormide: I don't think we can make that alias work, since we'd have the symlink `yq`, and a directory named `yq` parallel to it with a build file. Bazel complains when creating one or the other.

lib/private/repo_utils.bzl
```diff
@@ -77,8 +77,8 @@ def _platform(rctx):
     # Once we drop support for anything older than Bazel 5.1.1 than we can simplify
     # this function.
     if os == "windows":
-        proc_arch = (_get_env_var(rctx, "PROCESSOR_ARCHITECTURE", "", False) or
-                     _get_env_var(rctx, "PROCESSOR_ARCHITEW6432", "", False))
+        proc_arch = (_get_env_var(rctx, "PROCESSOR_ARCHITECTURE", "") or
+                     _get_env_var(rctx, "PROCESSOR_ARCHITEW6432", ""))
```
kormide, line 81: @gregmagolan Added this fix for windows. Could you verify this is right?

kormide:
Updated bash tests to work with windows. Verified on my virtual machine.

e2e/run_yq_symlinked_bin.sh
gregmagolan: oh right, the directory name conflicts with the binary name

gregmagolan: perhaps a convention such as `bazel run @yq//bin` for all our toolchains but lets discuss offline as that is better as a follow up and not that important atm

lib/private/yq_toolchain.bzl
gregmagolan: when repositories stamps out multiple of these (one per platform) `is_windows` will be the same value for all of them depending on whether the host is windows or not

gregmagolan: what you actually want to check is
```
repo_utils.is_windows(repository_ctx.attr.platform)
```
but `is_windows` in repo_utils takes an rtcx and not a platform string

kormide: Oh oops. Good catch.

kormide: Removed those changes.

lib/private/repo_utils.bzl
gregmagolan: Oh thanks. I messed that one up when I copied it over from here

https://github.com/bazelbuild/bazel/commit/32d1606dac2fea730abe174c41870b7ee70ae041#diff-22f86c807b9c37861c2637acaf8f6e2509858457a8ab5d5dc791b774e7896bf5L192

The upstream version of get_env_var has an additional param

```
def get_env_var(repository_ctx, name, default = None, enable_warning = True):
    """Find an environment variable in system path. Doesn't %-escape the value!"""
    if name in repository_ctx.os.environ:
        return repository_ctx.os.environ[name]
    if default != None:
        if enable_warning:
            auto_configure_warning("'%s' environment variable is not set, using '%s' as default" % (name, default))
        return default
    auto_configure_fail("'%s' environment variable is not set" % name)
```

https://github.com/bazelbuild/bazel/commit/32d1606dac2fea730abe174c41870b7ee70ae041#diff-22f86c807b9c37861c2637acaf8f6e2509858457a8ab5d5dc791b774e7896bf5R120

gregmagolan (approved):
🌮

PR #90, opened 2022-04-28
feat: add @aspect_bazel_lib_host repository and normalize function names in repo_utils

gregmagolan:
BREAKING


lib/private/host_repo.bzl
```diff
@@ -0,0 +1,37 @@
+"An repository that exposes information about the host platform"
```
kormide, line 1: nit: `A`

lib/repo_utils.bzl
```diff
@@ -3,10 +3,10 @@
 load("//lib/private:repo_utils.bzl", "repo_utils")
 load("//lib/private:patch.bzl", _patch = "patch")
 
-is_darwin_os = repo_utils.is_darwin_os
-is_linux_os = repo_utils.is_linux_os
-is_windows_os = repo_utils.is_windows_os
+is_darwin = repo_utils.is_darwin
```
alexeagle, line 6: I think this should just re-export the whole struct rather than each symbol from it

lib/private/host_repo.bzl
```diff
@@ -0,0 +1,37 @@
...
+
+load(":repo_utils.bzl", "repo_utils")
+
+def _impl(rctx):
+    # Base BUILD file for this repository
+    rctx.file("BUILD.bazel", """# Generated by @aspect_bazel_lib//lib/private:host_repo.bzl
+load("@bazel_skylib//:bzl_library.bzl", "bzl_library")
```
kormide, line 8: Does `defs.bzl` need to be exported?

lib/repositories.bzl
```diff
@@ -17,6 +18,8 @@ def aspect_bazel_lib_dependencies():
         ],
     )
 
+    maybe(host_repo, name = "aspect_bazel_lib_host")
```
alexeagle, line 21: can remove this?

lib/private/host_repo.bzl
```diff
@@ -0,0 +1,37 @@
...
+    rctx.file("BUILD.bazel", """# Generated by @aspect_bazel_lib//lib/private:host_repo.bzl
+load("@bazel_skylib//:bzl_library.bzl", "bzl_library")    
+bzl_library(
+    name = "defs",
+    srcs = ["defs.bzl"],
+    visibility = ["//visibility:public"],
+)
+""")
+
+    # index.bzl file for this repository
```
kormide, line 16: Comment incorrect

lib/repositories.bzl
kormide: Isn't this needed?

lib/private/host_repo.bzl
gregmagolan: .bzl files don't need to be exported to load them; only if they are to be used as deps

lib/repositories.bzl
gregmagolan: sure

lib/repo_utils.bzl
gregmagolan: good call. lets do that now.

lib/repositories.bzl
kormide: Without this how will anyone have access to the host repo?

lib/private/host_repo.bzl
gregmagolan: nice catch

lib/repositories.bzl
gregmagolan: User's would explicit instantiate it in their workspace

```
host_repo("some_repo_name")
```

kormide: Oh. Should we have a public api file for this then?

kormide (approved):

lib/repositories.bzl
gregmagolan: We should. I'll add one and then land.

PR #95, opened 2022-04-29
chore: replace is_windows select pattern with target_platform_has_constraint pattern

gregmagolan:
Based on https://github.com/bazelbuild/proposals/blob/main/designs/2019-11-11-target-platform-constraints.md#proposal-a-check-method-that-takes-a-constraint-value-label
Flag added in bazel 2.1.0 https://docs.bazel.build/versions/main/skylark/lib/ctx.html#target_platform_has_constraint

### DIFF
--- a/lib/copy_to_directory.bzl
+++ b/lib/copy_to_directory.bzl
@@ -100,9 +100,5 @@ def copy_to_directory(
         include_external_repositories = include_external_repositories,
         exclude_prefixes = exclude_prefixes,
         replace_prefixes = replace_prefixes,
-        is_windows = select({
-            "@bazel_tools//src/conditions:host_windows": True,
-            "//conditions:default": False,
-        }),
         **kwargs
     )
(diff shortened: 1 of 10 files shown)


kormide (approved):

alexeagle:
Saw this go by on Bazel slack:

Chad McElligott
How do I see what os and cpu platform bazel detects for my system?
I am trying to update a custom toolchain to support arm macos and want to verify that i'm specifying the platform identifiers correctly

Cristian Hancila
bazel query @local_config_platform//... --output=build (edited) 
there's also constraints.bzl in that same package with HOST_CONSTRAINTS variable

```
# DO NOT EDIT: automatically generated constraints list for local_config_platform
# Auto-detected host platform constraints.
HOST_CONSTRAINTS = [
  '@platforms//cpu:aarch64',
  '@platforms//os:osx',
]
```

PR #98, opened 2022-05-02
fix: fix a bug where yq failed to build in the root package

kormide:
Fixes https://github.com/aspect-build/bazel-lib/issues/85

### DIFF
--- a/lib/private/yq.bzl
+++ b/lib/private/yq.bzl
@@ -40,7 +40,7 @@ def _yq_impl(ctx):
 
     # For split operations, yq outputs files in the same directory so we
     # must cd to the correct output dir before executing it
-    bin_dir = ctx.bin_dir.path + "/" + ctx.label.package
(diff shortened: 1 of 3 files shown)


lib/private/yq.bzl
```diff
@@ -40,7 +40,7 @@ def _yq_impl(ctx):
 
     # For split operations, yq outputs files in the same directory so we
     # must cd to the correct output dir before executing it
-    bin_dir = ctx.bin_dir.path + "/" + ctx.label.package
+    bin_dir = "/".join([ctx.bin_dir.path, ctx.label.package]) if ctx.label.package else ctx.bin_dir.path
```
alexeagle, line 43: nit: we might want a path.join like https://github.com/bazelbuild/rules_nodejs/blob/ee2b155cbce2e7feb7855a67181baaa6924051da/nodejs/private/ts_config.bzl#L62-L63

alexeagle (approved):

PR #101, opened 2022-05-03
fix: fix a bug where bcr action wouldn't follow redirects to artifact

kormide:
My downloader didn't follow http `302`, which for some reason I wasn't running into locally.

### DIFF
--- a/.github/workflows/create-bcr-entry.mjs
+++ b/.github/workflows/create-bcr-entry.mjs
@@ -5,8 +5,8 @@ import {
   existsSync,
   copyFileSync,
   writeFileSync,
-  appendFileSync,
-  readdirSync,
+  createWriteStream,
+  unlink,
 } from "fs";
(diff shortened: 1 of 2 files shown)


.github/workflows/create-bcr-entry.mjs
```diff
@@ -180,18 +179,29 @@ function getVersionFromTag(version) {
...
         });
-        response.on("end", () => {
-          resolve();
-        });
-      })
-      .on("error", (err) => {
-        reject(new Error(err.message));
-      });
+        response.pipe(file);
+      } else if (response.statusCode === 302 || response.statusCode === 301) {
```
alexeagle, line 192: this is probably fine, but we also have the option of using some library from npm if it gets ugly. We can use the `ncc` trick to make a statically-linked file out of it.

alexeagle (approved):

PR #104, opened 2022-05-03
refactor: show a more helpful error message in copy_to_bin

kormide:


lib/private/copy_to_bin.bzl
```diff
@@ -35,6 +36,23 @@ def copy_file_to_bin_action(ctx, file, is_windows = False):
...
         return file
+    if ctx.label.package != file.owner.package:
+        fail(
+            """
+Expected to find file {file_basename} in {package}, but instead it is in {file_package}.
+
+To use copy_to_bin, either move {file_basename} to {package}, or move the copy_to_bin
+target to {file_basename}'s package using:
+    
+    buildozer 'new copy_to_bin {target_name}' {file_package}:__pkg__ && buildozer 'add srcs {file_basename}' {file_package}:{target_name}
```
alexeagle, line 47: nice, can you have it add the `load` statement as well?

alexeagle (approved):
awesome 😎

lib/private/copy_to_bin.bzl
kormide: Sure

kormide: Added

kormide: Added. Is the multiline command awkward?

kormide:
What it looks like:
```
ERROR: /home/derek/Code/bazel-lib/lib/tests/copy_to_bin/BUILD.bazel:14:12: in _copy_to_bin rule //lib/tests/copy_to_bin:fail: 
        File "/home/derek/Code/bazel-lib/lib/private/copy_to_bin.bzl", line 84, column 38, in _impl
                files = copy_files_to_bin_actions(ctx, ctx.files.srcs, is_windows = is_windows)
        File "/home/derek/Code/bazel-lib/lib/private/copy_to_bin.bzl", line 79, column 36, in copy_files_to_bin_actions
                return [copy_file_to_bin_action(ctx, file, is_windows = is_windows) for file in files]
        File "/home/derek/Code/bazel-lib/lib/private/copy_to_bin.bzl", line 40, column 13, in copy_file_to_bin_action
                fail(
Error in fail: 
Expected to find file a.json in //lib/tests/copy_to_bin, but instead it is in //lib/tests/jq.

To use copy_to_bin, either move a.json to //lib/tests/copy_to_bin, or move the copy_to_bin
target to a.json's package using:
    
    buildozer 'new copy_to_bin a' //lib/tests/jq:__pkg__ \
        && buildozer 'add srcs a.json' //lib/tests/jq:a \
        && buildozer 'new_load @aspect_bazel_lib//lib:copy_to_bin.bzl copy_to_bin' //lib/tests/jq:__pkg__
```

alexeagle (approved):
I think you can just print the three commands to be run, without the `&&`

Alternatively if you want to be fancy, you can run buildozer with an input file that has all the commands to run, but I don't think it will be any easier to print here.

Issue #51, opened 2022-03-24
Mirror releases to bcr


kormide:

alexeagle:
Seems like we need a real app with a server backend at this point.

alexeagle:
I think this is solved, at least for this one repo we now have automation. Probably the remaining work belongs in the bazel-contrib org.

PR #108, opened 2022-05-16
feat: mirror releases to bcr

kormide:
https://github.com/aspect-build/bazel-lib/issues/51

@alexeagle Are we ready for this now?

### DIFF
--- a/.github/workflows/bcr.yml
+++ b/.github/workflows/bcr.yml
@@ -22,12 +22,13 @@ on:
 
 env:
   # Bazel central registry to post a pull request to.
-  BCR: aspect-build/bazel-central-registry
+  BCR: bazelbuild/bazel-central-registry
 
   # Fork of the bazel central registery to push the pull request branch to
-  # BCR_FORK: aspect-build/bazel-central-registry
-  # TODO: disable this for bazel-lib as we always want to push to our own fork,
-  #  but re-enable for later plugin
+  BCR_FORK: aspect-build/bazel-central-registry
+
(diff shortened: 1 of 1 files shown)


.github/workflows/bcr.yml
```diff
@@ -64,15 +65,18 @@ jobs:
...
           body: |
-            [Automated] Publish ${{ github.repository }}@${{ steps.get_tag.outputs.TAG }} to bcr by [bazel-contrib](https://github.com/bazel-contrib).
+            [Automated] Publish ${{ github.repository }}@${{ steps.get_tag.outputs.TAG }}.
 
             Manual checks:
             - [ ] I have checked that the version is correct.
             - [ ] I have checked that the compatibility_level in MODULE.bazel is correct.
+
+            Submitters:
+            ${{ env.SUBMITTERS }}
```
kormide, line 80: I'm not sure we'll have permission to tag us as reviewers in bcr, so figured I'd just tag us.

alexeagle: I don't think we're actually going to want to perform these manual steps after the PR is generated. It will just give us a public outing when the PR is created and asks its creators for more input. I think we should just rely on merge-on-green rather than commit to doing manual checks

alexeagle (commented):
yes, totally ready to launch this :)

.github/workflows/bcr.yml
kormide: Removed the manual checks.

alexeagle (approved):

Issue #114, opened 2022-05-18
use provider to disallow wrong target in additional_update_targets


alexeagle:
users can write

```
update_source_files(name = "foo", additional_update_targets = ["//some/arbitrary-target"])
```

and the error is pretty obscure. We should ensure that the attribute is declared such that only compatible targets can be supplied.

Aghassi:
For reference, I passed in an `sh_binary` target no knowing that was "disallowed"

PR #115, opened 2022-05-18
fix: add mandatory provider for write_source_files additional_update_targets

kormide:
Fixes https://github.com/aspect-build/bazel-lib/issues/114

### DIFF
--- a/lib/private/write_source_file.bzl
+++ b/lib/private/write_source_file.bzl
@@ -5,6 +5,13 @@ load(":diff_test.bzl", _diff_test = "diff_test")
 load(":fail_with_message_test.bzl", "fail_with_message_test")
 load(":utils.bzl", "utils")
 
+WriteSourceFileInfo = provider(
+    "Provider for write_source_file targets",
+    fields = {
+        "executable": "Executable that updates the source files",
+    },
+)
+
 def write_source_file(
         name,
         in_file = None,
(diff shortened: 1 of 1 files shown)


lib/private/write_source_file.bzl
```diff
@@ -138,10 +145,7 @@ def _write_source_file_sh(ctx, paths):
 
     additional_update_scripts = []
     for target in ctx.attr.additional_update_targets:
-        if target[DefaultInfo].files_to_run and target[DefaultInfo].files_to_run.executable:
-            additional_update_scripts.append(target[DefaultInfo].files_to_run.executable)
-        else:
-            fail("additional_update_targets target %s does not provide an executable")
+        additional_update_scripts.append(target[WriteSourceFileInfo].executable)
```
alexeagle, line 148: oh nice, it even simplified a bit :)

lib/private/write_source_file.bzl
```diff
@@ -5,6 +5,13 @@ load(":diff_test.bzl", _diff_test = "diff_test")
 load(":fail_with_message_test.bzl", "fail_with_message_test")
 load(":utils.bzl", "utils")
 
+WriteSourceFileInfo = provider(
```
alexeagle, line 8: technically we could keep this symbol private? (leading underscore)
not sure if we should though, perhaps it's nice that someone could write a rule that interops

alexeagle (approved):

lib/private/write_source_file.bzl
kormide: Yeah I think I'm in favour of keeping it public :man_shrugging:

PR #117, opened 2022-05-20
fix: fix a bug where copy_to_directory wouldn't copy hidden files inside of folders

kormide:
Turns out `*` doesn't match hidden files.


lib/private/copy_to_directory.bzl
```diff
@@ -146,7 +146,7 @@ if [[ -f "{src}" ]]; then
     cp -f "{src}" "{dst}"
 else
     mkdir -p "{dst}"
-    cp -fR "{src}"/* "{dst}"
+    cp -fR "{src}/." "{dst}"
```
alexeagle, line 149: Do we need a fix for windows too?

alexeagle (approved):

lib/private/copy_to_directory.bzl
kormide: Nope the test passes on windows.

PR #118, opened 2022-05-20
chore: improve the readme

kormide:
https://github.com/aspect-build/bazel-lib/issues/116

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,13 +1,56 @@
 # Aspect's Bazel helpers library
 
-This is code we would contribute to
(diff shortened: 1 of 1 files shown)


alexeagle (approved):
Take a look at my readme branch for ideas of things you'd like to reuse. I'm out of time for today 👍🏻

kormide:
I tried to combine our approaches. I don't love the table headers but github markdown can't format tables without headers properly.

kormide:
I added an alternative without tables in the second commit. Which do you prefer?

alexeagle (approved):
Yeah it's better without the table I think

PR #126, opened 2022-06-03
fix: buildifier lint errors

kormide:
After this we should be green and ready for a 1.0 release if we want to do that.

https://buildkite.com/bazel/bazel-lib


gregmagolan (approved):

kormide:
Though we may want to wait until I figure out the BCR workflow before we go to 1.0 as they may be another release or two to test it.

PR #131, opened 2022-06-06
chore: remove hardcoded toolchain versions from workspace snippet

kormide:


.github/workflows/workspace_snippet.sh
```diff
@@ -48,12 +48,12 @@ Optional toolchains:
 
 load("@aspect_bazel_lib//lib:repositories.bzl", "register_jq_toolchains")
 
-register_jq_toolchains(version = "1.6")
+register_jq_toolchains()
```
kormide, line 51: No args defaults to latest.

alexeagle (approved):

PR #134, opened 2022-06-08
fix: copy_to_directory repairs read-only file permissions

alexeagle:
When two srcs have the same nested folder, we need to correct the permissions like bazel's sandbox would.
Also remove the -f on the copy operations, as those would hide mistakes where one input (source or generated) is overwritten by another

### DIFF
--- a/lib/private/copy_to_directory.bzl
+++ b/lib/private/copy_to_directory.bzl
@@ -145,6 +145,12 @@ if [[ -f "{src}" ]]; then
     mkdir -p "{dst_dir}"
     cp -f "{src}" "{dst}"
 else
+    if [[ -d "{dst}" ]]; then
+        # When running outside the sandbox, then an earlier copy will create the dst folder
+        # with nested read-only folders, so our copy operation will fail to write there.
+        # Make sure the output folders are writeable.
+        find "{dst}" -type d -print0 | xargs -0 chmod a+w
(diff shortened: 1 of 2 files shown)


kormide:
I think the `-f` is feature. You might have cases where you want files from the second entry in `srcs` to overwrite that in the first. I'm making use of this in `aio` to override boilerplate files. Of course that also requires telling buildifier not to change the order of `srcs`.

alexeagle:
okay I'll put back `-f` for now, it's orthogonal to this fix

kormide:
> okay I'll put back `-f` for now, it's orthogonal to this fix

I should add a test for the overwriting feature so that it doesn't accidentally regress.

kormide (approved):

gregmagolan:
current behaviour with `-f` is inconsistent between source files/folders and output files/folders. Since output files/folders are write protected, even with `-f` you'd fail to overwrite but with source files/folders you can.

To make it consistent we'd have to `chmod a+w` and `chmod -R a+w` before `cp -f` if the thing we are copying to exists.

gregmagolan:
Perhaps we fix this in a follow-up by making the `-f` an opt-in via `allow_overwrites` attribute so users don't accidentally stumble over it.

gregmagolan (approved):

PR #151, opened 2022-06-14
chore: warn when copy_to_bin reads from bazel-out

alexeagle:


lib/private/copy_to_bin.bzl
```diff
@@ -55,6 +55,23 @@ target to {file_basename}'s package using:
...
+            # buildifier: disable=print
+            print("""
+WARNING: sources are being copied from {bin}. This is probably not what you want.
+
+Add these lines to .bazelignore:
+bazel-out
+bazel-bin
+bazel-testlogs
+
+and/or correct the `glob` patterns that are including these files in the sources.
```
kormide, line 72: I thought that globs can only find source files. Do I have that wrong? Or is this a case where the glob is finding sources but there already is a bin copy and the bin label has has higher precedence than the source label?

lib/private/copy_to_bin.bzl
```diff
@@ -55,6 +55,23 @@ target to {file_basename}'s package using:
                 package = "%s//%s" % (ctx.label.workspace_name, ctx.label.package),
             ),
         )
+
+    if file.path.startswith("bazel-"):
```
kormide, line 59: If the file is generated, wouldn't the method exit early above on line 38?

lib/private/copy_to_bin.bzl
alexeagle: no it's even worse, Bazel is just descending into the `bazel-bin` convenience symlink and thinks it is still walking the source tree.

lib/private/copy_to_bin.bzl
alexeagle: Bazel thinks they are sources because incredibly, `bazel-bin` is not bazel-ignored by default

lib/private/copy_to_bin.bzl
kormide: :scream_cat:

lib/private/copy_to_bin.bzl
kormide: :scream_cat:

kormide (approved):

PR #153, opened 2022-06-14
feat: add allow_overwrites attribute to copy_to_directory and disable overwrites as the default behavior on linux and macos

gregmagolan:
Sadly this creates divergence on Windows vs Linux & MacOS on Windows overwrites are always allowed with robocopy


kormide:
> Sadly this creates divergence on Windows vs Linux & MacOS on Windows overwrites are always allowed with robocopy

What about the `/XO` option? https://stackoverflow.com/a/4364417

gregmagolan:
> What about the `/XO` option? https://stackoverflow.com/a/4364417

It looks like `/XO` will exclude but it won't fail so the behaviour will still be different.

kormide:
Couple of questions:
1. Do you think it's worth adding additional scripting to make up for the difference in behaviour between windows and osx/linux?
2. Should we consider this a breaking change? I am making use of the overwrite behaviour elsewhere, but it's also not documented as part of our public api, so 🤷

kormide (approved):

gregmagolan:
> 1. Do you think it's worth adding additional scripting to make up for the difference in behaviour between windows and osx/linux?

I do think its worth fixing; perhaps the fix is to use a `.bat` wrapper around the bash script and get away from using Windows native copy commands. Fixing robocopy seems difficult. `copy_directory` can still use `robocopy` since its not doing anything complicated but a `.bat` wrapper for `copy_to_directory` would make things simpler for that rule and consistent between platforms.

> 2. Should we consider this a breaking change? I am making use of the overwrite behaviour elsewhere, but it's also not documented as part of our public api, so 🤷

It seems like a fix to me since the overwrite behaviour currently works if you're overwriting a file copied from the source tree but not if your overwriting something that was copied from the output tree. This makes it consistent.

> 3. Should we enable it by default to get around ^?

Failing by default seems better since a user may accidentally overwrite & it's better to fail and have them be explicit about it.

PR #155, opened 2022-06-15
fix: fix a bug where write_source_files diff_test fails on windows when a source is from an external repository

kormide:
Turns out `diff_test` (skylib's and our fork's) relies on legacy external runfiles on Windows. We can either remove this flag, or we can try to fix the batch script to support regular runfiles trees.


kormide:
Feels like a step backwards. I want to take a closer look at the batch script.

kormide:
Removed the flag only for windows.

gregmagolan (approved):
🌮

PR #157, opened 2022-06-16
fix: fix a bug where two copy_to_bin actions couldn't copy the same file

kormide:
For example, if you have two `js_test` targets in the same package that use the same entry point, it will fail on Windows.

The problem is that `.bat` scripts with different names are being produced for the same copy.


lib/private/copy_file.bzl
```diff
@@ -35,7 +35,7 @@ def _copy_cmd(ctx, src, src_path, dst):
     # quoting or escaping is required.
     # Put a hash of the file name into the name of the generated batch file to
     # make it unique within the package, so that users can define multiple copy_file's.
```
gregmagolan, line 37: nit: mention that the label of the target is intentionally _not_ included in the `.bat` filename so that two different targets can copy the same source file to the output tree

gregmagolan (approved):
🌮

lib/private/copy_file.bzl
kormide: Added comment

PR #160, opened 2022-06-16
feat: add out_dirs to run_binary rule and deprecate output_dir

gregmagolan:
Corresponding change in rules_js is https://github.com/aspect-build/rules_js/pull/216


kormide (approved):

lib/private/expand_make_vars.bzl
```diff
@@ -59,41 +65,47 @@ def expand_variables(ctx, s, outs = [], output_dir = False, attribute_name = "ar
...
         output_dir: whether the rule is expected to output a directory (TreeArtifact)
+            Deprecated. For backward compatability with @aspect_bazel_lib 1.x. Pass
+            output tree artifacts to outs instead.
         attribute_name: name of the attribute containing the expression
 
     Returns:
         `s` with the variables expanded
     """
-    rule_dir = [f for f in [
+    rule_dir = "/".join([f for f in [
```
alexeagle, line 75: nit: same `_join` helper we have everywhere...

alexeagle (approved):

lib/private/expand_make_vars.bzl
gregmagolan: right; I should remember to use `@bazel_skylib//lib:paths.bzl` `join` function

Issue #162, opened 2022-06-17
Handle non-json (ie "raw") sources in `jq` macro


phlax:
`jq` itself will handle non json files in raw mode, but the macro doesnt

handling raw files would be particularly helpful when dealing with genquery output as `bazel query` doesnt support formatting results as json.

one workaround is to name the source files with a `.json` suffix but this means you have artefacts that are not valid json with json names

kormide:
Released in https://github.com/aspect-build/bazel-lib/releases/tag/v1.3.1

PR #163, opened 2022-06-17
fix: support non-json input files in jq

kormide:
Fixes https://github.com/aspect-build/bazel-lib/issues/162


alexeagle (approved):
would be neat to add a motivating example that gets input from a genquery like the linked issue's use case

kormide:
Added example to docs

PR #164, opened 2022-06-17
fix: relative_file fix when to path is itself relative starting with ../

gregmagolan:
Fix would be useful in https://github.com/aspect-build/rules_jest/pull/12

Should check what https://www.geeksforgeeks.org/python-os-path-relpath-method/ and https://nodejs.org/api/path.html#pathrelativefrom-to return in this case where the the `to` is a relative path starting with `../` itself


gregmagolan:
Added the test case, now to fix the impl....

gregmagolan:
@JesseTatasciore going to open a new PR

PR #165, opened 2022-06-18
chore: add vscode bazel container

Aghassi:
This change adds the necessary files so that if contributing via
vscodespaces you have bazel on your path via bazelisk and don't
need to do extra work to set it up

I also created an upstream issue for fixes I added to the container
to keep it ergonomic:

https://github.com/microsoft/vscode-dev-containers/pull/1507
https://github.com/microsoft/vscode-dev-containers/pull/1508

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -1,0 +1,26 @@
+# You can pick any Debian/Ubuntu-based image. 😊
+FROM mcr.microsoft.com/vscode/devcontainers/base:0-bullseye
+
+# Options for setup script
+ARG INSTALL_ZSH="true"
+ARG UPGRADE_PACKAGES="false"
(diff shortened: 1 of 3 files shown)


alexeagle:
Hm, this seems really cool to have things 'just work' when editing on GH, but this is a lot of SLOC for us to maintain that none of us understands

Aghassi:
Yeah unfortunately that's how VSCodespaces bootstraps. Not sure if there is a cleaner way... it's basically what Microsoft says to do at this point 😭

alexeagle:
Reading recent comments on your upstream issues, I get the sense it's not time for us to adopt this yet. @Aghassi as you know I'm super-keen on the idea though. If we could get this working in a more ergonomic way, I'd love to have every one of our OSS Bazel-related repos contain the codespaces solution.

PR #169, opened 2022-06-22
feat: let jq act like copy_to_bin

alexeagle:
This lets you have a file in the output tree with the same path as a source file, but modified.

Came up while translating an angular `ng-package.json` file to fix up the `dest` path. We don't want to change the source file name, as Angular CLI generates it, and we can't change the output file name.

### DIFF
--- a/lib/private/jq_toolchain.bzl
+++ b/lib/private/jq_toolchain.bzl
@@ -8,7 +8,9 @@ JQ_PLATFORMS = {
         release_platform = "osx-amd64",
         compatible_with = [
             "@platforms//os:macos",
-            "@platforms//cpu:x86_64",
+            # JQ only ships a universal binary; it should work on
+            # Apple Silicon (amd64) as well.
+            #"@platforms//cpu:x86_64",
         ],
     ),
     "linux_amd64": struct(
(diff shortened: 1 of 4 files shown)


kormide (approved):

lib/private/jq_toolchain.bzl
```diff
@@ -8,7 +8,9 @@ JQ_PLATFORMS = {
         release_platform = "osx-amd64",
         compatible_with = [
             "@platforms//os:macos",
-            "@platforms//cpu:x86_64",
+            # JQ only ships a universal binary; it should work on
+            # Apple Silicon (amd64) as well.
+            #"@platforms//cpu:x86_64",
```
kormide, line 13: What was this change for?

alexeagle: No toolchain found running on my Mac without this

jbedard: The toolchain it finds after commenting this out works though? Because the binary is "universal"?

kormide: Works fine on my M1.

alexeagle: Hm very strange let's debug maybe

PR #184, opened 2022-07-14
Add utility for checking file existence.

alexeagle:
Avoids continued copy-paste of this like https://github.com/aspect-build/rules_ts/blob/main/ts/defs.bzl#L35-L44

### DIFF
--- a/lib/private/utils.bzl
+++ b/lib/private/utils.bzl
@@ -110,7 +110,7 @@ def _file_exists(path):
     label = _to_label(path)
     file_abs = "%s/%s" % (label.package, label.name)
     file_rel = file_abs[len(native.package_name()) + 1:]
-    file_glob = native.glob([file_rel])
+    file_glob = native.glob([file_rel], exclude_directories = 1)
     return len(file_glob) > 0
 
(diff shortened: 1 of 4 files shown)


lib/private/utils.bzl
```diff
@@ -98,8 +98,24 @@ def _glob_directories(include, **kwargs):
...
+    Useful in macros to set defaults for a configuration file if it is present.
+    This can only be called during the loading phase, not from a rule implementation.
+
+    Args:
+        path: a label, or a string which is a path relative to this package
+    """
+    label = _to_label(path)
+    file_abs = "%s/%s" % (label.package, label.name)
+    file_rel = file_abs[len(native.package_name()) + 1:]
+    file_glob = native.glob([file_rel])
```
kormide, line 113: Should we set `exclude_directories=1` here?

alexeagle: yeah I guess since we called it "file exists" we wouldn't want a folder to return true

kormide (approved):

gregmagolan (approved):

PR #186, opened 2022-07-15
Basic stamping support

alexeagle:


lib/stamping.bzl
```diff
@@ -0,0 +1,128 @@
...
+def stamp_files(ctx):
+    """Provide the files that contain the result of the --workspace_status_command.
+
+    Returns:
+        If stamping is not enabled for the current build, returns None.
+        Otherwise, returns a struct containing (volatile_status, stable_status) keys
+    """
+    if is_stamping_enabled(ctx.attr):
+        return struct(
+            volatile_status = ctx.version_file,
```
gregmagolan, line 99: `volatile_status_file` and `stable_status_file` as the keys? the current values make downstream harder to read as `_status` implies some status file and not a file.
```
if stamp:
        args.add("--volatile_status_file", stamp.volatile_status_file)
        args.add("--stable_status_file", stamp.stable_status_file)

        inputs.extend([stamp.volatile_status_file, stamp.stable_status_file])
```

gregmagolan: Funny how version_file and info_file are terribly non-descriptive so we rename them here

gregmagolan:
Add a 2nd commit to use this in run_binary?

lib/stamping.bzl
```diff
@@ -0,0 +1,128 @@
...
+)
+```
+"""
+
+load("//lib/private:stamping.bzl", "is_stamping_enabled")
+
+def stamp_files(ctx):
+    """Provide the files that contain the result of the --workspace_status_command.
+
+    Returns:
```
gregmagolan, line 93:
```
Args:
    ctx: The rule context
```

lib/stamping.bzl
```diff
@@ -0,0 +1,128 @@
...
+        If stamping is not enabled for the current build, returns None.
```
gregmagolan, line 94: "If stamping is not enabled for rule in the current build, returns None."

lib/stamping.bzl
```diff
@@ -0,0 +1,128 @@
...
+        # ... my attributes ...
+    }, **STAMP_ATTRS),
+)
+```
+"""
+
+load("//lib/private:stamping.bzl", "is_stamping_enabled")
+
+def stamp_files(ctx):
+    """Provide the files that contain the result of the --workspace_status_command.
```
gregmagolan, line 91: You still get the files even if you don't specify a `--workspace_status_command` right? Just with the default values Bazel provides

lib/stamping.bzl
gregmagolan: Perhaps, elaborate here that stamping will be turned on for the rule if `stamp = 1` or if `stamp = -1` and the `--stamp` config flag is set

lib/stamping.bzl
alexeagle: buildifier didn't tell me, I'm confused when those warnings appear. feels like a bug, it starts yelling at me only later...

lib/stamping.bzl
alexeagle: yes that's true

lib/stamping.bzl
```diff
@@ -0,0 +1,128 @@
...
+    attrs = dict({
+        # ... my attributes ...
+    }, **STAMP_ATTRS),
+)
+```
+"""
+
+load("//lib/private:stamping.bzl", "is_stamping_enabled")
+
+def stamp_files(ctx):
```
gregmagolan, line 90: Pondering a better name than `stamp_files` for this function that better describes what is does,

```
stamp = check_stamp(ctx)
```

```
stamp = stamp_info(ctx)
```

```
stamp = maybe_stamp(ctx)
```

alexeagle: I like `maybe_stamp`

gregmagolan (approved):
🌮

gregmagolan:
squash commit 1 through 3 into 1 and then land with the 2 commits unsquahsed?

gregmagolan:
I already have

```
        "BAZEL_INFO_FILE": "$(INFO_FILE)",
        "BAZEL_VERSION_FILE": "$(VERSION_FILE)",
```

set always in js_run_binary. I'll remove those from there then and update to this new run_binary under the hood"

Issue #122, opened 2022-05-30
bad error message for missing directory input


alexeagle:
```
ERROR: /Users/alexeagle/Projects/rules_js/examples/npm_deps/BUILD.bazel:123:7: in directory_path rule //examples/npm_deps:actual14__entry_point:
Traceback (most recent call last):
	File "/private/var/tmp/_bazel_alexeagle/deecfabe94c39fb88a37966b779aac74/external/aspect_bazel_lib/lib/private/directory_path.bzl", line 17, column 13, in _directory_path
		fail("directory attribute must be created with Bazel declare_directory (TreeArtifact)")
```
is what you get if your entry point doesn't exist at all. the error message should indicate that possibility.

PR #192, opened 2022-07-21
fix: improve error message in directory_path when directory does not exist

kormide:
Fixes https://github.com/aspect-build/bazel-lib/issues/122.

The error message was already improved in https://github.com/aspect-build/bazel-lib/commit/128f7c63fbe3e2d3ead83ff7d6bb09728666acf2.


gregmagolan (approved):
🌮

PR #193, opened 2022-07-27
feat: add include_prefixes to copy_to_directory

gregmagolan:
This is going to be useful for the downstream npm_package which composes the copy_to_directory impl under the hood.

Follow up PR will add glob matching support to `include_prefixes`, `exclude_prefixes` and `replace_prefixes` using @jbedard glob helper from rules_js


lib/private/copy_to_directory.bzl
```diff
@@ -93,6 +136,13 @@ _copy_to_directory_attr = {
     "_windows_constraint": attr.label(default = "@platforms//os:windows"),
 }
 
+def _first_match(subject, tests, allow_partial = False):
+    for test in tests:
+        starts_with_test = test if allow_partial else test + "/"
```
jbedard, line 141: Do we trim trailing `"/"` somewhere? Do we want to support users passing directories both with + without the trailing `/`?

lib/tests/copy_to_directory/BUILD.bazel
```diff
@@ -300,3 +301,49 @@ copy_to_directory(
...
+
+# Case 13: include prefixes
+copy_to_directory(
+    name = "case_13",
+    srcs = case_srcs,
+    include_prefixes = [
+        # include a & b
+        "a",
+        "b",
+        # this partial match should include all of e
```
jbedard, line 313: Should it? Even though `e/e` does not match `e/e1` when using glob?
Or is `include_prefixes` strictly prefixes and will never be globs?

lib/private/copy_to_directory.bzl
gregmagolan: Nice catch. Logic updated.

lib/tests/copy_to_directory/BUILD.bazel
jbedard: Sounds like these are prefixes and in the future "glob prefixes" and not full globs 👍

gregmagolan: The glob equivalent would be an implied trailing `*` since we're looking for a prefix on a string

lib/private/copy_to_directory.bzl
jbedard: That always makes me wonder if we need more tests...

jbedard (approved):

lib/private/copy_to_directory.bzl
gregmagolan: We likely do. I'll update the test cases after I add proper wildcard matching in a follow up.

PR #194, opened 2022-07-27
feat: add glob_match starlark utility function that support "*", "**" and "?" expressions as well as optionally allow for "*" and "?" to match of path separators

gregmagolan:
Default matching behaviour (with `match_path_separator = False`) follows https://metacpan.org/dist/File-Globstar/view/lib/File/Globstar.pod#*

### DIFF
--- a/lib/glob_match.bzl
+++ b/lib/glob_match.bzl
@@ -1,0 +1,140 @@
+"""
+Basic glob match implementation for starlark.
+
+This was originally developed by @jbedard for use in rules_js
+(https://github.com/aspect-build/rules_js/blob/6ca32d5199ddc0bf19bd704f591030dc1468ca5f/npm/private/pkg_glob.bzl)
+to support the pnpm public-hoist-expr option (https://pnpm.io/npmrc#public-hoist-expr). The pnpm
+implementation and tests were used as a reference implementation:
+    https://github.com/pnpm/pnpm/blob/v7.4.0-2/packages/matcher/src/index.ts
+    https://github.com/pnpm/pnpm/blob/v7.4.0-2/packages/matcher/test/index.ts
+"""
+
+def _split_on(expr, splits):
+    # Splits an expression on the tokens in splits but keeps the tokens split in the result.
+    # Tokens are matched in order so a token such as `**` should come before `*`.
+    result = []
+    accumulator = ""
+    skip = 0
+    for i in range(len(expr)):
+        j = i + skip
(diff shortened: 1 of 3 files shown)


lib/string_match.bzl
```diff
@@ -0,0 +1,88 @@
...
+
+    Args:
+        exp: the wildcard expression
+        str: the string against which to match the wildcard expression.
+
+    Returns:
+        True if the string matches the wildcard expression
+    """
+
+    # Reduce all duplicate wildcards to single wildcards
```
jbedard, line 25: This is unrelated to glob `**` right? It just simplifies handling of double-wildcards which are pointless?

lib/string_match.bzl
```diff
@@ -0,0 +1,88 @@
...
+        else:
+            break
+
+    if exp.find("?") != -1:
+        fail("string_match: `?` expression not supported")
+
+    exp_i = 0
+    str_i = 0
+
+    # HACK: since "?" will never be in the expression, use it to
```
jbedard, line 39: But it soon will be! 🤷

gregmagolan: :shrug:

The principled way is to split on `*` and inject a `*` in between in element but not on the ends

lib/string_match.bzl
gregmagolan: Yeah, if you search for any sequence of characters twice you're not doing anything useful the 2nd time. Infinity * 2 is still infinity.

lib/string_match.bzl
```diff
@@ -0,0 +1,88 @@
+"""
+Basic string match implementation with wildcards for starlark.
+
+This was originally developed by @jbedard for use in rules_js (https://github.com/aspect-build/rules_js)
```
jbedard, line 4: Worth linking to the commit in the repo instead of just repo? 🤷

jbedard (approved):

lib/string_match.bzl
gregmagolan: No longer applies with latest changes

lib/string_match.bzl
gregmagolan: Good idea

lib/string_match.bzl
gregmagolan: Fixed

lib/string_match.bzl
gregmagolan: ✅

PR #195, opened 2022-07-28
chore: add glob_match to docs

gregmagolan:


jbedard (approved):
👍 if we're ready to make it public

gregmagolan:
> 👍 if we're ready to make it public

🚢 it

PR #196, opened 2022-07-28
feat: add glob support to copy_to_directory include_srcs_patterns, exclude_srcs_patterns, replace_prefixes and root_paths

gregmagolan:
This deprecates `exclude_prefixes` with docs suggesting to using `exclude_srcs_patterns` instead

BB: It renames `include_prefixes` that was added a few days ago to `include_srcs_patterns`, but we haven't had a release since then so it is not breaking

Globs will be useful for `copy_to_directory` in general. The downstream `npm_package` will benefit from `exclude_srcs_patterns` with globs in particular


lib/private/copy_to_directory.bzl
```diff
@@ -24,6 +25,9 @@ _copy_to_directory_attr = {
 
         "." values indicate the target's package path.
 
+        Globs expressions `**`, `*` and `?` are supported. See `glob_match` for more details.
```
kormide, line 28: Can `glob_match` be a hyperlink?

gregmagolan: Yes. Now that the doc for it has landed I can link to it

gregmagolan: ✅

kormide (approved):

lib/private/copy_to_directory.bzl
```diff
@@ -423,6 +497,17 @@ def copy_to_directory_action(
     # Replace "." root paths with the package name of the target
     root_paths = [p if p != "." else ctx.label.package for p in root_paths]
 
+    # Convert and append exclude_prefixes to srcs_exclude_patterns
```
jbedard, line 500: this is what will be removed in the future? Put a `TODO`?

gregmagolan: ✅

PR #203, opened 2022-07-29
feat: add include_srcs_packages and exclude_srcs_packages and improve implementation and documentation of other filters and transformations

gregmagolan:
Turns out that matching on the package name without worrying about root path transformations is very useful downstream for `npm_package`


lib/private/copy_to_directory.bzl
```diff
@@ -222,60 +317,82 @@ def _copy_paths(
     else:
         fail("Unsupported type")
 
-    # if the file is from an external repository check if that repository should
-    # be included in the output directory
-    if src_file.owner and src_file.owner.workspace_name and not src_file.owner.workspace_name in include_external_repositories:
+    if not src_file.owner:
+        msg = "Expected an owner target label for file {} but found none".format(src_file)
+        fail(msg)
```
kormide, line 322: When would this not have an owner?

gregmagolan: Hopefully never. Bazel docs say that the function "may" return None 🤷 https://docs.bazel.build/versions/main/skylark/lib/File.html#owner

kormide (approved):
:taco:

PR #204, opened 2022-07-30
fix: improve error messages to copy_to_directory and add missing aliases

gregmagolan:


lib/private/copy_to_directory.bzl
```diff
@@ -325,9 +325,15 @@ def _copy_paths(
         msg = "Expected owner target label for file {} to have a package name but found None".format(src_file)
         fail(msg)
 
+    if not include_srcs_packages:
+        fail("An empty 'include_srcs_packages' list will exclude all srcs and result in an empty directory")
+
```
kormide, line 330: Is this something that a user of the rule can cause by passing in an empty `list`? Or just users of the action macro?

If I included `srcs` and then set `include_srcs_packages = []` I wouldn't expect the directory to be empty. Maybe I'm missing something.

gregmagolan: `include_srcs_packages` is configuration both in the rule attr API in the action macro. The rule itself uses the action macro under the hood.

kormide (approved):

PR #208, opened 2022-08-03
chore: add Aspect CI configuration

mattem:
Off we jolly well go 🚀

### DIFF
--- a/.aspectci/config.yaml
+++ b/.aspectci/config.yaml
@@ -1,0 +1,8 @@
+---
+env:
+  CC: /dev/null
+workspaces:
+  - .
+  - e2e/bzlmod
+tasks:
+  test:


.aspectci/config.yaml
```diff
@@ -0,0 +1,8 @@
+---
+env:
+  CC: /dev/null
```
gregmagolan, line 3: disable local cc toolchain?

mattem: I did on bazel-examples, and it leads to new issues (see https://buildkite.com/aspect/bazel-examples/builds/19#01826408-869a-49d2-a67e-9c33673a6f7d/109-143). It then highlights the cycle dependency between bazel-lib and our gcc toolchain, both issues are outside the scope of this diff, so `/dev/null` it is for the time being.

alexeagle: is this like `build --action_env=BAZEL_DO_NOT_DETECT_CPP_TOOLCHAIN=1` ?

mattem: Not quite, `BAZEL_DO_NOT_DETECT_CPP_TOOLCHAIN` will disable the detection of the CC toolchain, meaning you'd need to then provide one in the WORKSPACE (see link above), where just setting `CC` to `/dev/null` will still cause bazel to configure the toolchain automatically, just it will be pointing to `/dev/null`

PR #213, opened 2022-08-12
fix: fix make var expansion in expand_template

gregmagolan:
The test for this was accidentally working since `BINDIR` was added to substitutions via `**ctx.var`. With the substitution updated to `"{BINDIR}": "$(BINDIR)"` it fails. This fixes the expand_template rule to use the `expand_variables` helper.

A follow-up update to this rule would add suport for substitution of stamp vars if a `stamp` attribute is set to `True`. This, however, would require an action so will leave it for a follow-up.

### DIFF
--- a/lib/private/expand_template.bzl
+++ b/lib/private/expand_template.bzl
@@ -1,0 +1,63 @@
+"expand_template rule"
+
+load(":expand_locations.bzl", _expand_locations = "expand_locations")
+load(":expand_variables.bzl", _expand_variables = "expand_variables")
+
+def _expand_template_impl(ctx):
+    template = ctx.file.template
+
+    substitutions = {}
+    for k, v in ctx.attr.substitutions.items():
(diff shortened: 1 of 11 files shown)


lib/private/expand_template.bzl
```diff
@@ -0,0 +1,64 @@
...
+expand_template_lib = struct(
+    doc = """Template expansion
+    
+This performs a simple search over the template file for the keys in substitutions,
+and replaces them with the corresponding values.
+
+Values may also use location templates as documented in [expand_locations](#expand_locations)
+as well as [configuration variables] such as `$(BINDIR)`, `$(TARGET_CPU)`, and `$(COMPILATION_MODE)`.
+
+Additional exp
```
kormide, line 29: Typo?

gregmagolan: Yup

kormide (approved):

PR #216, opened 2022-08-18
fix: fix issue in copy_to_directory where root_path is the full path of a directory input

tkoolen-bdi:
Hi @alexeagle, @gregmagolan :wave:. This adds a test case that is currently failing, but used to pass. See https://github.com/tkoolen-bdi/bazel-lib/tree/tkoolen/root-paths-glob-test-0.2.7 for the same test rebased onto v0.2.7, where it passes. We depend on the behavior in 0.2.7, which I think is what one would expect.


gregmagolan:
Found the issue. Thanks for the repro. I'll push the fix to this PR and merge it in.

gregmagolan (approved):
🌮

gregmagolan:
I cut a 1.10.1 release that will include the fix

tkoolen-bdi:
Thanks for the quick fix!

PR #218, opened 2022-08-18
ci: enable engflow RBE for github actions

alexeagle:


platforms/BUILD.bazel
```diff
@@ -0,0 +1,12 @@
+platform(
```
kormide, line 1: What does this do?

alexeagle: I copied from @f0rmiga - I'm not really sure and he'll answer better than my guess :)

.github/workflows/ci.bazelrc
```diff
@@ -2,6 +2,36 @@
 # Included with a --bazelrc option in the call to bazel
 build --announce_rc
 test --test_output=errors
-build --disk_cache=~/.cache/bazel
 build --repository_cache=~/.cache/bazel-repo
 test --test_env=XDG_CACHE_HOME
+
+# When no remote cache, use a local one
+build:local --disk_cache=~/.cache/bazel
```
kormide, line 9: Should we do this for the repository cache as well?

alexeagle:
FYI @mattem will be an interesting project to fold this config under our bk/cci runners with rosetta

.github/workflows/ci.bazelrc
alexeagle: repository cache should be unconditional, so I didn't change that line.

platforms/BUILD.bazel
f0rmiga: When we set this as `--host_platform`, it will match with the RBE we enabled. The `constraint_values` here have to match with the remote executors, and the `exec_properties` are passed to the remote executor as metadata; in this case, we are asking for it to run the actions with this specific container.

f0rmiga (approved):

kormide (approved):

Issue #220, opened 2022-08-19
Conflicting files in copy_file_action


purkhusid:
I'm using copy_file_action in a custom publishing rule and I've got an example of how it's used here:
https://github.com/purkhusid/rules_dotnet/blob/695ffff8b2dd4f4aae30372f0f3e094be52721dc/examples/publish_binary/BUILD#L21

The issue is that 2 targets try to create the exact same copy script on Windows because the inputs and outputs are the exact same files.

The culprit is this line here:
https://github.com/aspect-build/bazel-lib/blob/cd6c2f63ee380132fe8f206901de2abaa05ae385/lib/private/copy_file.bzl#L40

The comment above mentions that it's intentional to not have the label name in the file name but I'm not 100% understanding the reason for it. Would be be open up to a change that adds the label name to the bat file name?

kormide:
This sounds familiar and may have been fixed in https://github.com/aspect-build/bazel-lib/commit/80603fb7eb92ba079762c182443b13d5dc9964f5. Do you want to try updating to the latest release, `1.10.1`?

purkhusid:
@kormide I'm using the latest release. It seems like the fix fixes `copy_to_bin` but breaks `copy_file_action` in my case.
The change also doesn't make sense to me because by removing the label from the bat file name I would think that it would be more likely that the actions that create the bat output would collide.

purkhusid:
Ah, the issue I have is that Bazel complains that there  are 2 different targets that output the same script file. So it's not the outputs of the copy action that conflict but the script files. I'll paste the error when I'm back at the computer.

kormide:
> https://github.com/purkhusid/rules_dotnet/blob/695ffff8b2dd4f4aae30372f0f3e094be52721dc/examples/publish_binary/BUILD#L21

In the example you linked two, are those two targets supposed to produce two different output files with the same content, or the same output file?

If it's the former, maybe what I need to do here is name the script after a hash of the source and the destination file, rather than just the source.

kormide:
I got a PR up that hopefully fixes your case: https://github.com/aspect-build/bazel-lib/pull/221

purkhusid:
> > https://github.com/purkhusid/rules_dotnet/blob/695ffff8b2dd4f4aae30372f0f3e094be52721dc/examples/publish_binary/BUILD#L21
> 
> In the example you linked to, are those two targets supposed to produce two different output files with the same content, or the same output file?
> 
[...]

PR #221, opened 2022-08-20
fix: fix a bug where the same file couldn't be copied to the same output dir twice on Windows

kormide:
Addresses https://github.com/aspect-build/bazel-lib/issues/220/

The issue here is that the two copy scripts with the same name will be produced in the output tree for copying two different files.


kormide:
Tests are passing but the rbe workflows are failing.
```
ERROR: Failed to init TLS infrastructure using 'engflow.crt' as client certificate: File does not contain valid certificates: engflow.crt Failed to init TLS infrastructure using 'engflow.crt' as client certificate: File does not contain valid certificates: engflow.crt
```

alexeagle (approved):

PR #227, opened 2022-08-26
Add JUnit XML output to diff_test

jfirebaugh:
This improves the developer experience when a `diff_test` fails. For example, on CircleCI's "Tests" tab (the default tab that opens for a failing job), this is what it looks like before:

Screen Shot 2022-08-26 at 11 01 35 AM

And after:

Screen Shot 2022-08-26 at 11 01 21 AM


kormide (approved):
LGTM. Could you add a `TODO` comment in the windows implementation to also add xml output?

jfirebaugh:
Yes, done!

PR #230, opened 2022-08-30
Cleanup diff_test and copy_file implementations

dzbarsky:
The change to copy_file slightly modifies the progress_message to avoid creating extra strings at analysis time. These actions are so fast that I doubt a human can read the progress message anyway, and there's a ton of them, so this feels like the right tradeoff. (Does anyone rely on the exact messages?)

The change to diff_test moves the string templates into separate files so they are expanded as an action, rather than in the analysis phase. Also, having them in a separate file makes editing them easier (syntax highlighting, etc), and we could even run formatters/linters on them if we care.


jbedard:
Have you seen any measurable results from these `progress_message` changes? It does seem like they are [recommended](https://bazel.build/rules/lib/actions) so we should probably be following them either way, just curious.

lib/private/BUILD.bazel
```diff
@@ -148,6 +148,11 @@ bzl_library(
     srcs = ["diff_test.bzl"],
 )
 
+exports_files(
```
kormide, line 151: `exports_files` declarations ususally go at the top of a file. Did buildifier put it here? 🤔

kormide (approved):
LGTM. I have some slight concerns about diverging too far from skylib since that will make getting our fixes there more difficult, but moving to separate files is nice for readability so 👍 .

jbedard (approved):

dzbarsky:
> Have you seen any measurable results from these `progress_message` changes? It does seem like they are [recommended](https://bazel.build/rules/lib/actions) so we should probably be following them either way, just curious.

I don't have any big repos using these rules, but in the past I've seen similar changes reduce analysis time significantly (2-3x) in aggregate

lib/private/BUILD.bazel
dzbarsky: I put it there by hand to keep it next to the diff_test.bzl. I can combine it with the other public export instead.

dzbarsky:
The test failure was due to the bash template escaping `{`/`}`, which is no longer necessary. @jbedard mind kicking off tests again?

PR #232, opened 2022-08-31
ci: exclude rbe jobs from matrix if no engflow secret

alexeagle:


alexeagle:
@kormide something must be wrong with our GH actions setup that this PR is green there... 🤔

kormide:
> @kormide something must be wrong with our GH actions setup that this PR is green there... 🤔

That's concerning... I'll look into it.

alexeagle:
Note, this PR comes from my fork of the project, so rbe wasn't tested here, as expected.

kormide (approved):

PR #233, opened 2022-08-31
fix: checkout correct head on ci workflow

kormide:


kormide:
This shouldn't have passed :(

kormide:
I wonder if it's running the `ci.yaml` in `main` 🤔 ...

PR #235, opened 2022-09-02
fix: determine copy file action based on execution platform

kormide:
@alexeagle || @gregmagolan 

Trying to follow the suggestions in these threads [1](https://bazelbuild.slack.com/archives/CDCMRLS23/p1660894883837699), [2](https://bazelbuild.slack.com/archives/CA31HN1T3/p1660853078169879). Apparently what we're doing here is choosing the batch or the shell script based on the target platform rather than the platform that's doing the copy.

I'm stuck. Running into errors like the following:
```
While resolving toolchains for target //lib/tests/copy_file:a_copy_from_dir: Unable to find an execution platform for target platform @local_config_platform//:host from available execution platforms []
```

Seems like it isn't recognizing my host system (Linux) as an execution platform. I haven't worked much with platforms or constraints in bazel yet, so maybe I'm missing something.

[Related Slack thread](https://bazelbuild.slack.com/archives/CDCMRLS23/p1661018990795499?thread_ts=1660894883.837699&cid=CDCMRLS23)

### DIFF
--- a/lib/private/copy_file.bzl
+++ b/lib/private/copy_file.bzl
@@ -24,7 +24,7 @@ cmd.exe (on Windows). `_copy_xfile` marks the resulting file executable,
 `_copy_file` does not.
 """
 
-load(":copy_common.bzl", _COPY_EXECUTION_REQUIREMENTS = "COPY_EXECUTION_REQUIREMENTS")
+load(":copy_common.bzl", "COPY_EXEC_GROUPS", _COPY_EXECUTION_REQUIREMENTS = "COPY_EXECUTION_REQUIREMENTS")
 load(":directory_path.bzl", "DirectoryPathInfo")
 
 def _copy_cmd(ctx, src, src_path, dst):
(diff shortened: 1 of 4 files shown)


lib/private/copy_file.bzl
```diff
@@ -81,9 +82,10 @@ def _copy_bash(ctx, src, src_path, dst):
         progress_message = progress_message,
         use_default_shell_env = True,
         execution_requirements = _COPY_EXECUTION_REQUIREMENTS,
+        exec_group = "not_windows",
     )
 
-def copy_file_action(ctx, src, dst, dir_path = None, is_windows = False):
+def copy_file_action(ctx, src, dst, dir_path = None):
```
kormide, line 88: I might have to keep this parameter for now if we want to avoid a breaking change since this is public. We may also want to pass in the `exec_group` names in case a downstream user doesn't use `windows`/`not_windows`.

gregmagolan:
This approach won't work since all exec_groups in a rule must be satisfy-able from available execution platforms

gregmagolan:
We can fix the bug in copy actions using this approach however https://github.com/aspect-build/bazel-lib/pull/241

PR #236, opened 2022-09-09
copy_to_directory performance improvements

jbedard:
I noticed `copy_to_directory_action` consuming 3-5s in the angular-cli analysis phase, making the analysis phase itself often 30s.

First commit: mainly reducing the number of `depset.to_list()` calls. Seemed to improve performance by at most 5% so maybe this case of `to_list()` isn't done on large enough `depset`s to make a big difference?

Second commit: in ~~go~~ python (and go) passing arrays as params clones them, where a slice is a reference to the array. So switching from arrays to slices avoid cloning the arrays within the file loop. This made the analysis phase in my test go from 30s to 10s...

### DIFF
--- a/lib/private/copy_to_directory.bzl
+++ b/lib/private/copy_to_directory.bzl
@@ -631,6 +631,7 @@ def copy_to_directory_action(
         srcs,
         dst,
         additional_files = [],
+        additional_files_depsets = [],
         root_paths = ["."],
         include_external_repositories = [],
         include_srcs_packages = ["**"],
@@ -654,6 +655,8 @@ def copy_to_directory_action(
         dst: The directory to copy to. Must be a TreeArtifact.
 
         additional_files: Additional files to copy that are not in the DefaultInfo or DirectoryPathInfo of srcs
+
+        additional_files_depsets: Additional depsets to copy that are not in the DefaultInfo or DirectoryPathInfo of srcs
 
         root_paths: List of paths that are roots in the output directory.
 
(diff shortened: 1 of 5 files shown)


lib/private/copy_to_directory.bzl
```diff
@@ -655,6 +656,8 @@ def copy_to_directory_action(
 
         additional_files: Additional files to copy that are not in the DefaultInfo or DirectoryPathInfo of srcs
 
+        additional_files_depsets: Additional depsets to copy that are not in the DefaultInfo or DirectoryPathInfo of srcs
```
jbedard, line 659: Is this a public method? If so do we think this is a reasonable API to add?

gregmagolan:
I don't remember all of the context on this one. Lets go over it together on Monday

lib/private/glob_match.bzl
```diff
@@ -15,9 +15,10 @@ def _split_on(expr, splits):
     result = []
     accumulator = ""
     skip = 0
-    for i in range(len(expr)):
+    exp_len = len(expr)
```
jbedard, line 18: Although very small the repeated `len()` calls actually showed up when profiling, may as well drop it 🤷

jbedard: Actually I'm going to revert this for now. `len()` should be O(1) so this might be overkill, and isn't related to the main commit. Can maybe do this later...

lib/private/copy_to_directory.bzl
```diff
@@ -765,7 +753,7 @@ def copy_to_directory_action(
             replace_prefixes = replace_prefixes,
         )
         if src_path != None:
-            dst_path = skylib_paths.normalize("/".join([dst.path, output_path]))
+            dst_path = skylib_paths.normalize("%s/%s" % (dst.path, output_path))
```
gregmagolan, line 756: Using `%s` seems like it would be slower than `"/".join` since it would have to do string token replacements

gregmagolan (approved):
🌮

lib/private/copy_to_directory.bzl
gregmagolan: It is part of the public API exported via `lib/copy_to_directory.bzl`.

Would be simpler to make the existing `additional_files` mutable and accept either a list or a depset. You can check the type of it in the macro and convert it.

gregmagolan (changes requested):

gregmagolan (approved):
:taco:^2

PR #237, opened 2022-09-09
Add `shlib/lib/assertions.sh` and related tests

cgrindel:


shlib/lib/assertions.sh
```diff
@@ -0,0 +1,81 @@
+#!/usr/bin/env bash
+
+# This is used to determine if the library has been loaded
```
f0rmiga, line 3:
```suggestion
# This is used to determine if the library has been loaded.
```

shlib/lib/assertions.sh
```diff
@@ -0,0 +1,81 @@
+#!/usr/bin/env bash
+
```
f0rmiga, line 2:
```suggestion
#!/usr/bin/env bash

set -o nounset -o pipefail
```

shlib/lib/assertions.sh
```diff
@@ -0,0 +1,81 @@
...
+#
+# Args:
+#   err_msg: The message to print to stderr.
+#
+# Outputs:
+#   stdout: None.
+#   stderr: The error message.
+fail() {
+  local err_msg="${1:-}"
+  [[ -n "${err_msg}" ]] || err_msg="Unspecified error occurred."
```
f0rmiga, line 16:
```suggestion
  if [ -z "${err_msg}" ]; then err_msg="Unspecified error occurred."; fi
```

shlib/lib/assertions.sh
```diff
@@ -0,0 +1,81 @@
...
+  echo >&2 "${err_msg}"
+  exit 1
+}
+
+make_err_msg() {
+  local err_msg="${1}"
+  local prefix="${2:-}"
+  [[ -z "${prefix}" ]] || \
+    local err_msg="${prefix} ${err_msg}"
+  echo "${err_msg}"
```
f0rmiga, line 26:
```suggestion
  if [ -n "${prefix}" ]; then
    err_msg="${prefix} ${err_msg}"
  fi
  echo "${err_msg}"
```

shlib/tests/lib_tests/assertions_tests/assert_equal_test.sh
```diff
@@ -0,0 +1,43 @@
+#!/usr/bin/env bash
+
+# --- begin runfiles.bash initialization v2 ---
+# Copy-pasted from the Bazel Bash runfiles library v2.
+set -uo pipefail; f=bazel_tools/tools/bash/runfiles/runfiles.bash
```
f0rmiga, line 5: I know this is copy-paste but my OCD doesn't allow me to leave this alone.

```suggestion
set -o nounset -o pipefail; f=bazel_tools/tools/bash/runfiles/runfiles.bash
```

shlib/tests/lib_tests/assertions_tests/assert_fail.sh
```diff
@@ -0,0 +1,36 @@
+#!/usr/bin/env bash
+
+FAIL_ERR_MSGS=()
+
+fail() {
+  local err_msg="${1:-Unspecified error occurred.}"
```
f0rmiga, line 6:
```suggestion
  local err_msg="${1:-"Unspecified error occurred."}"
```

shlib/tests/lib_tests/assertions_tests/assert_fail.sh
```diff
@@ -0,0 +1,36 @@
...
+  FAIL_ERR_MSGS+=( "${err_msg}" )
+}
+
+reset_fail_err_msgs() {
+  FAIL_ERR_MSGS=()
+}
+
+new_fail(){
+  local err_msg="${1:-}"
+  [[ -n "${err_msg}" ]] || err_msg="Unspecified error occurred."
```
f0rmiga, line 16: Please, convert to an `if` statement, it reads much better.

shlib/lib/assertions.sh
cgrindel: I actually meant to remove this code. Thanks!

shlib/lib/assertions.sh
cgrindel: I don't set options in library scripts as it may thwart options that are set by the client of the library.

shlib/lib/assertions.sh
cgrindel: Why make this change? Do you see a problem with the existing logic?

shlib/lib/assertions.sh
cgrindel: Same question here as above.

shlib/tests/lib_tests/assertions_tests/assert_equal_test.sh
cgrindel: Does it violate our style guide for shell code?

shlib/tests/lib_tests/assertions_tests/assert_fail.sh
cgrindel: I don't believe the extra double quote is necessary. Can you explain what condition this may help?

shlib/tests/lib_tests/assertions_tests/assert_fail.sh
cgrindel: As you can tell, I disagree. Are we use a particular code style that I should review?

shlib/tests/lib_tests/assertions_tests/assert_equal_test.sh
cgrindel: I went ahead and changed this. I do want to note that we are not consistent in this repo about using explicit Bash options. If that is what we want, we should add it to a style guide. [Counter example](https://github.com/aspect-build/bazel-lib/blob/main/lib/paths.bzl#L16)

shlib/tests/lib_tests/assertions_tests/assert_fail.sh
cgrindel: I went back and looked at the [Google shell style guide](https://google.github.io/styleguide/shellguide.html). It is silent on a preference for boolean shortcuts vs. if-then statements. For the sake of moving the PR forward, I have switched to if-then statements.

shlib/tests/lib_tests/assertions_tests/assert_fail.sh
cgrindel: @f0rmiga Not sure if you saw my question 👆.

shlib/lib/assertions.sh
f0rmiga: Yes, readability.

shlib/tests/lib_tests/assertions_tests/assert_equal_test.sh
f0rmiga: Yes, how about we start a style guide for Bash? It's a language that can make lots of programmers mad because they don't understand the code. If we aim for rules that make the code more intelligible, maybe we can save some lost souls.

shlib/tests/lib_tests/assertions_tests/assert_fail.sh
f0rmiga: Readability

shlib/tests/lib_tests/assertions_tests/assert_fail.sh
f0rmiga: See my comment on creating a style guide.

shlib/lib/assertions.sh
cgrindel: I will change this for now, as it is blocking my other PRs.

f0rmiga (approved):

PR #238, opened 2022-09-09
ci: prepare bazel-lib for publish-to-bcr app

kormide:
The template files now go in `.bcr`. Don't need the github actions scripts anymore as that work is done by the app.


gregmagolan (approved):
🌮

PR #241, opened 2022-09-13
fix: fix host platform detection on all copy actions which always run locally

gregmagolan:
Relies on checking `HOST_CONTRAINTS` from
```
load("@local_config_platform//:constraints.bzl", "HOST_CONSTRAINTS")
```

to figure out the host platform.

Since these actions all have "local" execution requirements then we can assume the host platform is the execution platform


kormide (approved):
:taco:

gregmagolan:
Test locally on Windows and works as expected

PR #251, opened 2022-09-20
feat: chdir_binary is an sh_binary that wraps a tool and does a cd first

alexeagle:
As far as I know, java_binary still can't be directly `bazel run` so that it starts in the working directory, and Java makes it nearly impossible to change the working directory of the JVM.  @shs96c can correct me I'm sure.

### DIFF
--- a/lib/paths.bzl
+++ b/lib/paths.bzl
@@ -1,6 +1,5 @@
 "Public API"
 
-load("@bazel_skylib//rules:write_file.bzl", "write_file")
 load("//lib/private:paths.bzl", "paths")
 
 relative_file = paths.relative_file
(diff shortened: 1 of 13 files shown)


gregmagolan (approved):
Sweet. Could use a test but otherwise lgtm

lib/paths.bzl
```diff
@@ -22,3 +23,44 @@ source "$(grep -sm1 "^$f " "$0.exe.runfiles_manifest" | cut -f2- -d' ')" 2>/dev/
 { echo>&2 "ERROR: cannot find $f"; exit 1; }; f=; set -e
 # --- end runfiles.bash initialization v2 ---
 """
+
+def chdir_binary(name, binary, chdir = "$BUILD_WORKSPACE_DIRECTORY", **kwargs):
```
gregmagolan, line 27: put this in `chdir_binary.bzl` instead?

alexeagle: as a new public API entrypoint, or in private/chdir_binary.bzl and exposed in paths.bzl?

gregmagolan: prefer a new public API since paths is currently just small utility functions. more entry points the better for bazel-lib IMO since it is a utility library and it makes the docs easier to read on the left navbar

alexeagle: Ptal

gregmagolan (dismissed):
🌮

lib/paths.bzl
gregmagolan: lgtm 🧙

lib/wrap_binary.bzl
```diff
@@ -0,0 +1,62 @@
...
+- intercept exit code, e.g. to make an "expect_fail"
+- change user, e.g. to deal with containerized build running as root, but tool requires non-root
+- intercept signals, e.g. to make a tool behave as a Bazel persistent worker
+"""
+
+load(":paths.bzl", "BASH_RLOCATION_FUNCTION")
+load(":utils.bzl", "to_label")
+load("@bazel_skylib//rules:write_file.bzl", "write_file")
+
+def chdir_binary(name, binary, chdir = "$BUILD_WORKSPACE_DIRECTORY", **kwargs):
```
kormide, line 18: nit: I'm wondering if the `BUILD_WORKSPACE_DIRECTORY` is intuitive. Most Bazel users are accustomed to binaries running either in the execroot or the runfiles root. I feel like defaulting to the source tree could confuse some people.

kormide (dismissed):

lib/wrap_binary.bzl
alexeagle: yeah I agree, also at the use site it's better to make this value explicit.

alexeagle:
@kormide WDYT about naming the new rule `wrap_binary(chdir = "")` so that we'd add the other features to it, rather than make more wrapper if you need to do two shimming things?

kormide:
> @kormide WDYT about naming the new rule `wrap_binary(chdir = "")` so that we'd add the other features to it, rather than make more wrapper if you need to do two shimming things?

Yeah not having to wrap a binary several times sounds like a better DX. So long as most things we need to wrap can be accomplished by just adding one attribute to the macro and they are generally mutually exclusive.

kormide:
The more I think about it, the more I can see how an "everything" macro could get out of hand. I think I'm going to change my opinion here and vote for small functions. Users can macroize several of them together if they need to.

alexeagle:
I struggled to deal with the fact that binary `args` and `env` are dropped if the tool is run under an action. I might need some help to get this across the finish line

kormide:
> I struggled to deal with the fact that binary `args` and `env` are dropped if the tool is run under an action. I might need some help to get this across the finish line

I can take it over.

cgrindel:
Converted to a draft until this work is prioritized.

alexeagle:
Dropping this since it's so stale and evidently not important enough that I had to come back to it.

PR #265, opened 2022-10-26
fix: use bzlmod @@ in to_label if Bazel version is 6 or greater

gregmagolan:
This is needed for rules_js e2e/bzlmod to go green with Bazel 6.0.0rc1

### DIFF
--- a/lib/private/utils.bzl
+++ b/lib/private/utils.bzl
@@ -48,23 +48,16 @@ def _to_label(param):
     Returns:
         a Label
     """
+    root_repo = "@@" if _is_bazel_6_or_greater() else "@"
     param_type = type(param)
     if param_type == "string":
         if param.startswith("@"):
             return Label(param)
         if param.startswith("//"):
-            return Label("@" + param)
-
-        # resolve the relative label from the current package
-        # if 'param' is in another workspace, then this would return the label relative to that workspace, eg:
-        # `Label("@my//foo:bar").relative("@other//baz:bill") == Label("@other//baz:bill")`
+            return Label("{}{}".format(root_repo, param))
         if param.startswith(":"):
             param = param[1:]
-        if native.package_name():
-            return Label("@//" + native.package_name()).relative(param)
-        else:
-            return Label("@//:" + param)
-
+        return Label("{}//{}:{}".format(root_repo, native.package_name(), param))
     elif param_type == "Label":
         return param
     else:
--- a/lib/tests/utils_test.bzl
+++ b/lib/tests/utils_test.bzl
@@ -16,6 +16,23 @@ def _to_label_test_impl(ctx):
     # assert that to_label can convert from string to label
     asserts.true(env, utils.to_label("//hello/world") == Label("//hello/world:world"))
     asserts.true(env, utils.to_label("//hello/world:world") == Label("//hello/world:world"))
+    asserts.true(env, utils.to_label("@//hello/world") == Label("@//hello/world:world"))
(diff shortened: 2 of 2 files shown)


lib/private/utils.bzl
```diff
@@ -48,23 +48,16 @@ def _to_label(param):
...
-        # `Label("@my//foo:bar").relative("@other//baz:bill") == Label("@other//baz:bill")`
+            return Label("{}@{}".format("@" if is_bazel_6_or_greater else "", param))
         if param.startswith(":"):
             param = param[1:]
-        if native.package_name():
-            return Label("@//" + native.package_name()).relative(param)
-        else:
-            return Label("@//:" + param)
-
+        return Label("{}@//{}:{}".format("@" if is_bazel_6_or_greater else "", native.package_name(), param))
```
alexeagle, line 60: what if the bazel 6 user has `--noenable_bzlmod`?

lib/private/utils.bzl
```diff
@@ -48,23 +48,16 @@ def _to_label(param):
...
     if param_type == "string":
         if param.startswith("@"):
             return Label(param)
         if param.startswith("//"):
-            return Label("@" + param)
-
-        # resolve the relative label from the current package
-        # if 'param' is in another workspace, then this would return the label relative to that workspace, eg:
-        # `Label("@my//foo:bar").relative("@other//baz:bill") == Label("@other//baz:bill")`
+            return Label("{}@{}".format("@" if is_bazel_6_or_greater else "", param))
```
alexeagle, line 57: I think it's generally more readable for `@@` to appear in the interpolation rather than one `@` in the format string and another in the parameters

gregmagolan: Had it like that originally with an `else:` and then wasn't happy with the code duplication 🤷

lib/private/utils.bzl
gregmagolan: With Bazel 5
```
Error in Label: Illegal absolute label syntax: @@//hello/world
```
With Bazel 6 with bzlmod enabled all good
With Bazel 6 with `--noenable_bzlmod` also all good

lib/private/utils.bzl
alexeagle: `return Label("{}{}".format("@@" if is_bazel_6_or_greater else "@", param))`

is what I'm suggesting

gregmagolan: 🧙

.bazelversion
```diff
@@ -1,4 +1,4 @@
-5.3.2
+6.0.0rc1
```
alexeagle, line 1: same concern as rules_js - we should make sure bazel-lib still works with Bazel 5 and we don't accidentally introduce bazel6-only features for a long time

lib/private/utils.bzl
```diff
@@ -48,23 +48,16 @@ def _to_label(param):
     Returns:
         a Label
     """
+    is_bazel_6_or_greater = _is_bazel_6_or_greater()
```
alexeagle, line 51: nit: could be `root_repo = _is_bazel_6_or_greater() : "@@" : "@"` since that's the only thing you need this boolean for

alexeagle (approved):

.bazelversion
gregmagolan: right... guess I'll prefactor in the matrix here as well..... the Bazel major version rabbit hole is a deep one

lib/private/utils.bzl
gregmagolan: 🧙 ^2

kormide (approved):

PR #269, opened 2022-10-27
feat: add command attribute to run_binary

gregmagolan:


gregmagolan:
This was the wrong approach. Going with https://github.com/aspect-build/rules_js/pull/564 instead.

PR #270, opened 2022-10-27
feat: add maybe_http_archive

gregmagolan:


docs/BUILD.bazel
```diff
@@ -113,4 +113,9 @@ stardoc_with_diff_test(
     bzl_library_target = "//lib:testing",
 )
 
+stardoc_with_diff_test(
+    name = "maybe",
```
alexeagle, line 117: could go in utils?

alexeagle (approved):

kormide (approved):
👍

docs/BUILD.bazel
gregmagolan: that's true... it is already in utils.bzl in bazel_tools

PR #273, opened 2022-10-31
feat: add assert_outputs

alexeagle:
It's a simple way to make an executable example demonstrate what outputs a rule produces. See https://github.com/aspect-build/rules_ts/pull/214 for an example usage in the real world.


gregmagolan (approved):
Neat

kormide (approved):

PR #274, opened 2022-11-02
perf: do not execute glob for non-glob copy_to_directory root_paths/replace_prefix params

jbedard:
df09e9c720e49c54c10478636cbaf428e8cadb7a gets my basic copy test (no globs) from 60s to ~11s. The rest is some basic tests and cleanup that should make it easier to avoid more perf issues in the future (just putting the quick-returns in a common location etc).

The last commit might be a breaking change and we shouldn't do it? But an empty glob really is invalid...


lib/private/copy_to_directory.bzl
```diff
@@ -283,20 +283,26 @@ _copy_to_directory_attr = {
...
 
 def _any_globs_match(exprs, path):
+    # Exit quick in the common case of having a "**".
+    if "**" in exprs:
+        return True
+
+    # Special case: support non-standard empty glob expr.
+    # Only an empty expression matches an empty path.
+    if path == "":
+        return "" in exprs
```
gregmagolan, line 293: looks reasonable to me

gregmagolan (approved):

PR #275, opened 2022-11-03
chore: nit cleanup in copy_to_directory _longest_glob_match helper

gregmagolan:
Slightly easier to read and understand imo

### DIFF
--- a/lib/private/copy_to_directory.bzl
+++ b/lib/private/copy_to_directory.bzl
@@ -299,7 +299,7 @@ def _any_globs_match(exprs, path):
 
 def _longest_glob_match(expr, path):
     if not is_glob(expr):
-        return path[:len(expr)] if path.startswith(expr) else None
+        return expr if path.startswith(expr) else None
 
     for i in range(len(path)):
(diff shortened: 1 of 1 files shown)


lib/private/copy_to_directory.bzl
```diff
@@ -299,7 +299,7 @@ def _any_globs_match(exprs, path):
 
 def _longest_glob_match(expr, path):
     if not is_glob(expr):
-        return path[:len(expr)] if path.startswith(expr) else None
+        return expr if path.startswith(expr) else None
```
jbedard, line 302: Looking for a subpath but returning a "glob" 🤔

Sounds confusing to me 😛

jbedard (approved):
🤷‍♂️

lib/private/copy_to_directory.bzl
gregmagolan: Half a dozen of one and 6 of the other?

Reads nicely for me with the `not is_glob(expr)` right before 🤷‍♂️

PR #276, opened 2022-11-04
refactor: move bzl_library target out of @local_config_platform and just use exports_files in there instead

gregmagolan:
Emulates what the `@local_config_platform` repo will look like after https://github.com/bazelbuild/bazel/pull/16665 lands


alexeagle (approved):
yeah that seems like the targetted fix :)

PR #278, opened 2022-11-10
feat: install Aspect CLI for all users

alexeagle:
Assuming you are using Bazelisk, this causes it to fetch Aspect's CLI and use that to wrap Bazel

Note, we might want to update Renovate someday to understand this file so users are more likely to stay on current versions.


gregmagolan:
Lets wait until I finalize the Bazelisk bootstrap story today and make an official 4.2.0 release before landing this

gregmagolan:
The release up on the CDN atm wasn't built off of main and I haven't pushed the v4.2.0 tag yet upstream

gregmagolan (approved):

PR #280, opened 2022-11-12
ci: fix pattern for Bazel version CI matrix so Aspect CLI is still used

gregmagolan:


.github/workflows/ci.yaml
```diff
@@ -112,8 +115,6 @@ jobs:
         # the //:bazel_version_test uses it and it needs to match the Bazel version being run
         run: |
           echo "${{ matrix.bazelversion }}" > .bazelversion
```
alexeagle, line 117: we could remove this line too?

alexeagle (approved):

.github/workflows/ci.yaml
gregmagolan: nice catch

PR #283, opened 2022-11-14
perf: optimize glob_match if the expression does not have any glob symbols

gregmagolan:
Turns out the `_split_on(GLOB_CHARS)` function is expensive so it is better to short-circuit before calling that


kormide (approved):

jbedard:
I thought we would have done this already 🤔 

👍

PR #289, opened 2022-11-22
Add platform_transition_binary rule

kormide:
@alexeagle this was mostly copy-pasta'ed from https://github.com/bazelbuild/rules_rust/issues/276#issuecomment-1320198911.

Transitions are a new topic in Bazel for me. I'm not sure how to properly exercise or test the rule


alexeagle:
@f0rmiga and @illicitonion would be good reviewers

illicitonion:
Looks great, thanks!

Test-wise, I'd suggest an integration test in lib/tests/transitions/BUILD.bazel that looks something like:

```starlark
go_binary(
    name = "generic_go_binary",
    srcs = [...],
    tags = ["manual"],
)

platform_transition_binary(
    name = "transitioned_go_binary_x86_64",
    target_platform = ":x86_64_linux",
)

platform_transition_binary(
    name = "transitioned_go_binary_armv7_linux",
    target_platform = ":armv7_linux",
)

sh_test(
    name = "test_go_binary_is_x86_64",
    srcs = [...],
    data = [":transitioned_go_binary_x86_64"],
    args = ["$(rootpath :transitioned_go_binary_x86_64)"],
)
```

and then a `sh_test` which runs `file "$(realpath "$1")"` and asserts the output looks like it's for the right platform.

f0rmiga (changes requested):

alexeagle:
add me back as a reviewer once the test is added?

kormide:
> add me back as a reviewer once the test is added?

Yup, sorry I've been meaning to get back to this one.

kormide:
I'm getting this error when I try to declare a `platform_transition_binary` with a target platform that doesn't match my linux x64 machine:
```
ERROR: /home/derek/.cache/bazel/_bazel_derek/cb18fde45e94a264d41b0ba5b41ae079/external/io_bazel_rules_go/BUILD.bazel:86:17: While resolving toolchains for target @io_bazel_rules_go//:cgo_context_data: No matching toolchains found for types @bazel_tools//tools/cpp:toolchain_type.
```
Is there something I need to do to register a cpp toolchain?

f0rmiga:
It depends. You may be able to disable CGO if no C code is being used. What target needs it?

illicitonion:
Yeah, I suspect setting `pure = "on"` will make this go away, but if you push a WIP I'm happy to play around with it locally!

lib/tests/transitions/BUILD.bazel
```diff
@@ -18,6 +19,14 @@ platform(
     ],
 )
 
+platform(
+    name = "arm64_linux",
+    constraint_values = [
+        "@platforms//os:linux",
+        "@platforms//cpu:arm64",
+    ],
```
illicitonion, line 27:
```suggestion
        "@platforms//cpu:arm64",
        "@io_bazel_rules_go//go/toolchain:cgo_off",
    ],
```

illicitonion (commented):
This should fix the issue :)

I've also filed https://github.com/bazelbuild/rules_go/pull/3390 to make this work slightly more by default

lib/tests/transitions/BUILD.bazel
kormide: Thanks for your help! Got it working.

alexeagle (approved):

f0rmiga (changes requested):

lib/transitions.bzl
```diff
@@ -42,3 +44,60 @@ platform_transition_filegroup = rule(
...
+        # In order for the symlink to have the same basename as the original
+        # executable (important in the case of proto plugins), put it in a
+        # subdirectory named after the label to prevent collisions.
+        new_executable = ctx.actions.declare_file(paths.join(ctx.label.name, new_executable_name))
+        ctx.actions.symlink(
+            output = new_executable,
+            target_file = original_executable,
+            is_executable = True,
+        )
+        files = depset(direct = [new_executable])
```
f0rmiga, line 75: Why doesn't this include the files?

kormide: @illicitonion Was there a reason for not including the files from DefaultInfo in your original snippet?

illicitonion: Either it's an accidental omission, or it's forbidden. Feel free to give it a try! If it works, great! If it doesn't, I suspect there may be a restriction where `DefaultInfo.files` is only allowed to contain files generated by the current rule.

kormide: Right, makes sense that if the output it just a transitioned binary then that's all we should have in `files`.

It looks like if the `binary` is not executable, we just forward the default info, but passing in a non executable label isn't a valid case, right? I added a fixup commit to just fail early instead.

illicitonion (approved):
Wonderful, thanks so much!!

lib/transitions.bzl
f0rmiga: Look at my implementation above in https://github.com/aspect-build/bazel-lib/blob/2c66c1c778d1050bf147a9a65fcc6f53ee47f938/lib/transitions.bzl#L15-L24 for an example of how I did it.

Only forwarding the symlink to the binary may work in most cases, but you would miss files, e.g., when you transition it to add to a tarball. It won't necessarily pick the runfiles. This can include shared objects required by the binary.

kormide: Added the binary's files as transitive deps.

f0rmiga: Perfect, thank you!

f0rmiga (approved):

lib/transitions.bzl
illicitonion: Ooh, interesting... I just tried to apply this everywhere I was using this pattern before, and found that as a side-effect of this change, we can't use a `platform_transition_binary` as a value of an attribute which is set to `allow_single_file = True`, which makes it a little less useful...

f0rmiga: @illicitonion interesting. Does Bazel accept the original binary you are trying to transition as a single file? (I know it won't run, but at least pass the analysis phase)

illicitonion: > @illicitonion interesting. Does Bazel accept the original binary you are trying to transition as a single file? (I know it won't run, but at least pass the analysis phase)

Yeah, the original `go_binary` in my example passes.

When using `platform_transition_binary`, the issue is because `DefaultInfo.files` on the `platform_transition_binary` target has two files. If I remove the `transitive` arg, everything starts working again.

f0rmiga: This might be similar to this bug: https://github.com/bazelbuild/bazel/pull/16381.

nickdesaulniers:
Nice! I found this via: https://github.com/bazelbuild/bazel/discussions/25810. Any chance to get something similar for transitioning `cc_library`, too?  https://github.com/bazel-contrib/bazel-lib/issues/1065

PR #295, opened 2022-12-03
feat: add write_source_file to public API

gregmagolan:


alexeagle (commented):
Why is it better than one rule with an attribute which is a singleton list?

gregmagolan:
> Why is it better than one rule with an attribute which is a singleton list?

For the single file case I think it reads better, especially in the `executable = True` case. This is more similar to the `write_file` rule from skylib too so if users are familiar with that they will find this for recognizable:

```
write_source_file(
    name = "write_a",
    in_file = "a_desired,
    out_file = "a",
    executable = True,
)
```

vs

```
    write_source_files(
        name = "write_a",
        files = {
            "a": "a_desired",
        },
        executable = True,
    )
```

lib/write_source_files.bzl
```diff
@@ -113,9 +110,9 @@ def write_source_files(
 
             To set different executable permissions on different source tree files use multiple `write_source_files` targets.
 
-        additional_update_targets: List of other `write_source_files` or other executable updater targets to call in the same run.
```
jbedard, line 116: are "other executable updater targets" not supported?

jbedard (approved):

lib/write_source_files.bzl
gregmagolan: they are but only with an undocumented provider so easier to remove this line than to document the provider

PR #296, opened 2022-12-03
fix: add diff_test to docs

gregmagolan:
Also adds one missing gitignore


alexeagle (approved):
I don't see the gitignore change

gregmagolan:
> I don't see the gitignore change

Oh. I must have done something strange locally and the .gitignore was out of date. It is correct at HEAD

PR #297, opened 2022-12-03
fix: write_source_files should use target platform for additional_update_targets

gregmagolan:


lib/private/write_source_file.bzl
```diff
@@ -158,7 +158,14 @@ _write_source_file_attrs = {
     "out_file": attr.string(mandatory = False),
     "executable": attr.bool(),
     # buildifier: disable=attr-cfg
-    "additional_update_targets": attr.label_list(cfg = "host", mandatory = False, providers = [WriteSourceFileInfo]),
+    "additional_update_targets": attr.label_list(
+        # Intentionally the target platform since the target is always meant to be `bazel run` on the host machine
```
jbedard, line 162: Intentionally *use ?

jbedard (approved):

lib/private/write_source_file.bzl
gregmagolan: 🤦

PR #306, opened 2023-01-01
chore: bring in multi-platform releases rules for copy_to_directory golang binary

gregmagolan:
We've done this in silo & in the OSS aspect-cli repo. In theory we could replace the duplication in those repositories using `release.bzl` from here but I'm not making release.bzl or hashes.bzl part of the public API yet here. Could be done in the future.

### DIFF
--- a/tools/BUILD.bazel
+++ b/tools/BUILD.bazel
@@ -1,0 +1,19 @@
+load(":release.bzl", "multi_platform_go_binaries", "release")
+
+exports_files([
+    "create_release.sh",
+])
+
+multi_platform_go_binaries(
+    name = "copy_to_directory",
+    embed = ["//tools/copy_to_directory:copy_to_directory_lib"],
+    tags = ["manual"],
+)
+
+release(
+    name = "release",
+    tags = ["manual"],
+    targets = [
+        ":copy_to_directory",
+    ],
+)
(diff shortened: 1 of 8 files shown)


alexeagle:
Make a ticket for public API? (maybe `multi_platform_go_binaries` belongs in rules_go?)

lib/private/BUILD.bazel
```diff
@@ -11,3 +14,17 @@ exports_files(
...
     visibility = ["//lib/private/docs:__pkg__"],
 )
+
+multi_platform_go_binaries(
+    name = "copy_to_directory",
+    embed = ["//lib/private/copy_to_directory:copy_to_directory_lib"],
+    tags = ["manual"],
+)
+
+release(
```
alexeagle, line 24: where does this get released to? artifacts on the github release?

alexeagle (approved):

lib/private/BUILD.bazel
gregmagolan: it will be in a follow-up; I'll release the tools with each bazel-lib release and copy_to_directory will get a toolchain to use them

gregmagolan:
> Make a ticket for public API? (maybe `multi_platform_go_binaries` belongs in rules_go?)

That's probably a better place for it yeah

Issue #301, opened 2022-12-22
[Bug]: copy_to_directory execution-time performance issues


matthewjh:
### What happened?

Hi,

This is a follow-up to https://github.com/aspect-build/bazel-lib/issues/258. The analysis-time performance issues were addressed by https://github.com/aspect-build/bazel-lib/commit/350408b93d3d4f714f287beebdb4581a7d3ef13e (thank you!), but the execution-time performance is still problematic. I wanted to raise this to track that.

The problem is that the copy_to_directory is generally slow and particularly slow on MacOS. The wall time on our build went up after moving from rules_nodejs to rules_js, mainly because of copy_to_directory. It seems that any non-trivial folder (i.e. a library) is going to block for anything from 2s to 10s on this action (and more on MacOS). This seems problematic for rules_js users in general as copy_to_directory is in the critical path of any change in a linked npm_package, which seems like a path that needs to be as fast as possible for the UX around a developer making local changes and ibazel rebuilding those to work.

`copy_to_directory` generates a bash script which is invoked through `ctx.actions.run_shell`. The script, in my case looks like:

```shell
if [[ ! -e "bazel-out/darwin_arm64-fastbuild/bin/libs/generic/src/lib/user-pref-service/user-pref-service.js" ]]; then echo "file 'bazel-out/darwin_arm64-fastbuild/bin/libs/generic/src/lib/user-pref-service/user-pref-service.js' does not exist"; exit 1; fi
if [[ -f "bazel-out/darwin_arm64-fastbuild/bin/libs/generic/src/lib/user-pref-service/user-pref-service.js" ]]; then
    mkdir -p "bazel-out/darwin_arm64-fastbuild/bin/libs/generic/generic/src/lib/user-pref-service"
    cp -v -n "bazel-out/darwin_arm64-fastbuild/bin/libs/generic/src/lib/user-pref-service/user-pref-service.js" "bazel-out/darwin_arm64-fastbuild/bin/libs/generic/generic/src/lib/user-pref-service/user-pref-service.js" >> "$OUT_CAPTURE" 2>>"$OUT_CAPTURE"
else
    if [[ -d "bazel-out/darwin_arm64-fastbuild/bin/libs/generic/generic/src/lib/user-pref-service/user-pref-service.js" ]]; then
        # When running outside the sandbox, then an earlier copy will create the dst folder
        # with nested read-only folders, so our copy operation will fail to write there.
        # Make sure the output folders are writeable.
[...]

PR #308, opened 2023-01-02
feat: add copy_to_directory_bin_action with improved performance by moving glob match & copying to a golang binary

gregmagolan:
This is a pre-factor for improved performance on `copy_to_directory` and `npm_package` (in rules_js) which will address https://github.com/aspect-build/bazel-lib/issues/301.

Not yet load bearing on `copy_to_directory` but ready for review.

In a follow-up I will add release artifacts for the copy_to_directory golang binary in https://github.com/aspect-build/bazel-lib/pull/309 & then another with one with a toolchain definition. Those should land first to get a release of the binaries out before the `copy_to_directory` flip can happen.


gregmagolan:
BREAKING CHANGES that will come:
```
$ gd lib/tests/copy_to_directory/BUILD.bazel
diff --git a/lib/tests/copy_to_directory/BUILD.bazel b/lib/tests/copy_to_directory/BUILD.bazel
index 66c289d..21b4577 100644
--- a/lib/tests/copy_to_directory/BUILD.bazel
+++ b/lib/tests/copy_to_directory/BUILD.bazel
@@ -82,10 +82,8 @@ copy_to_directory(
         # flatten test_a & test_b to the root
         "test_a": "",
         "test_b": "",
-        # some paths that won't match
-        "a/": "wont_match_a_is_terminal_path",
-        "a/a2": "wont_match_since_a2_is_in_a_tree_artifact",
     },
+    verbose = True,
 )
 
 diff_test(
@@ -178,8 +176,8 @@ copy_to_directory(
     name = "case_8",
     srcs = case_srcs,
     exclude_srcs_patterns = [
-        "a",
-        "b",
+        "a/*",
+        "b/*",
         "e/e*",
     ],
 )
@@ -224,8 +222,6 @@ copy_to_directory(
         # flatten test_a & test_b to the root
         "test_a": "",
         "test_b": "",
-        # some paths that won't match
-        "a/a2": "wont_match_since_a2_is_not_included_in_the_directory_path_in_srcs",
     },
 )
 
@@ -313,8 +309,8 @@ copy_to_directory(
     name = "case_13",
     srcs = case_srcs,
     include_srcs_patterns = [
-        "a",
-        "b",
+        "a/*",
+        "b/*",
         "?/e*",
         "f/f2/f2",
     ],
@@ -334,8 +330,8 @@ copy_to_directory(
         "f/f2/f1",
     ],
     include_srcs_patterns = [
-        "a",
-        "b",
+        "a/*",
+        "b/*",
         "?/e*",
         "f/f?/**",
     ],
```

gregmagolan:
Note to self for the flip:
```
    """Legacy helper function to copy files to a directory.

    This helper calculates copy paths in Starlark during analysis and performs the copies in a
    bash/bat script. For improved analysis and runtime performance, it is recommended the switch
    to `copy_to_directory_bin_action` which calculates copy paths and performs copies with a tool
    binary, typically the `@aspect_bazel_lib//tools/copy_to_directory` `go_binary` either built
    from source or provided by a toolchain.

    This helper is used by copy_to_directory. It is exposed as a public API so it can be used within
    other rule implementations where additional_files can also be passed in.
```

WORKSPACE
```diff
@@ -29,11 +29,17 @@ load("//lib:host_repo.bzl", "host_repo")
 
 host_repo(name = "aspect_bazel_lib_host")
 
-load("@bazel_gazelle//:deps.bzl", "gazelle_dependencies")
+load("//:deps.bzl", "go_dependencies")
+
+# gazelle:repository_macro deps.bzl%go_dependencies
+# gazelle:repository go_repository name=org_golang_x_tools importpath=golang.org/x/tools
+# https://github.com/bazelbuild/bazel-gazelle/issues/1217#issuecomment-1152236735
+go_dependencies()
```
kormide, line 37: We'll also need to call this in the bzlmod module extension.

gregmagolan: This dep won't be passed on to users. They'll be a toolchain to setup instead similar to the yq & jq toolchains.

kormide (approved):

tools/copy_to_directory/main.go
```diff
@@ -1,7 +1,347 @@
...
+	}
+
+	// Calculate copy paths
+	copyPaths, err := calcCopyPaths(cfg)
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	// Perform copies
+	// TODO: split out into parallel go routines?
```
kzadorozhny, line 333: Yes. The `calcCopyPaths` could also benefit from parallelization. `copy` operation can be scheduled as soon as a file is found.

gregmagolan: Good idea. We can do some fancy stuff with it in golang. Tho I may wait until @f0rmiga is back from vacation in a week before tackling it as he's much better at go routines than I am.

PR #309, opened 2023-01-03
chore: update release to include copy_to_directory action binary

gregmagolan:


workspace_status.sh
```diff
@@ -0,0 +1,27 @@
...
+
+# "stable" keys, should remain constant over rebuilds, therefore changed values will cause a
+# rebuild of any stamped action that uses ctx.info_file or genrule with stamp = True
+# Note, BUILD_USER is automatically available in the stable-status.txt, it matches $USER
+echo "STABLE_BUILD_SCM_SHA $(git rev-parse HEAD)"
+echo "STABLE_BUILD_SCM_LOCAL_CHANGES $(has_local_changes)"
+
+if [ "$(git tag | wc -l)" -gt 0 ]; then
+    echo "STABLE_BUILD_SCM_TAG $(git describe --tags)"
+fi
```
kormide, line 27: Do we may use of these keys for stamping anywhere?

gregmagolan: We don't yet but I'm going to add a `--version` flag to the tool here before landing and make use of it

workspace_status.sh
```diff
@@ -0,0 +1,27 @@
...
+    if [ "$(git status --porcelain)" != "" ]; then
+        echo dirty
+    else
+        echo clean
+    fi
+}
+
+# "volatile" keys, these will not cause a re-build because they're assumed to change on every build
+# and its okay to use a stale value in a stamped binary
+echo "BUILD_TIME $(date "+%Y-%m-%d %H:%M:%S %Z")"
```
kormide, line 17: Do we need this when `BUILD_TIMESTAMP` should already exist? Was this just to change the format?

kormide (approved):

workspace_status.sh
gregmagolan: I'll0 remove since it's not being used. I copied this file over from the aspect-cli repository.

PR #311, opened 2023-01-04
feat: improved copy_to_directory performance & globbing support using copy_to_directory_bin_action

gregmagolan:
Adds the toolchain and uses the copy_to_directory golang binary release artifacts from the latest release.

Globbing performance & support is improved by using the https://github.com/bmatcuk/doublestar library and no longer doing any globbing in our limited Starlark implementation. This is the same globbing library used by [gazelle](https://github.com/bazelbuild/bazel-gazelle). See https://github.com/bmatcuk/doublestar#patterns on supported globbing patterns.

Globs are also now applied to full file paths for all files with TreeArtifacts since they are done at runtime and not in Starlark at analysis time where Tree Artifacts are opaque. This is slightly BREAKING for use cases that relied on matching on the opaque directory artifact path.

This will be a chicken & egg problem for changes to the go program in the future since you need a release to update the toolchain.

In a future PR, the copy file calls in the copy_to_directory golang binary program will be parallelized with go routines to further improve copy_to_directory performance.


gregmagolan:
I'll do some downstream testing of this commit in rules_js npm_package before landing

gregmagolan:
Hmmm. This is more breaking that I thought for downstream rule sets that want to use `copy_to_directory_bin_action` such as rules_js since they will now have to register the copy_to_directory toolchain. 🤔

lib/private/copy_to_directory_toolchain.bzl
```diff
@@ -0,0 +1,186 @@
...
+
+    return [default_info, toolchain_info]
+
+copy_to_directory_toolchain = rule(
+    implementation = _copy_to_directory_toolchain_impl,
+    attrs = {
+        "bin": attr.label(
+            mandatory = True,
+            allow_single_file = True,
+        ),
```
kormide, line 85: `executable = True`?

kormide:
> Will adopting this warrant a bump to rules_js 2.0 downstream?!? 🤔

Do we have other useful breaking changes that can go into a 2.0?

gregmagolan:
> > Will adopting this warrant a bump to rules_js 2.0 downstream?!? 🤔
> 
> Do we have other useful breaking changes that can go into a 2.0?

None that are ready to go. There are lots of `TODO(2.0)` comments of things that can be removed or changed.

kormide:
> > > Will adopting this warrant a bump to rules_js 2.0 downstream?!? 🤔
> > 
> > 
> > Do we have other useful breaking changes that can go into a 2.0?
> 
> None that are ready to go. There are lots of `TODO(2.0)` comments of things that can be removed or changed.

Kind of sketchy, but you could sneak it into the `npm_translate_lock`/`npm_import` macros, at least until `2.0`?

lib/private/copy_to_directory.bzl
```diff
@@ -27,6 +27,18 @@ For more information each filters / transformations applied, see
 the documentation for the specific filter / transformation attribute.
 """
 
+_glob_support_docstring = """Glob patterns are supported. Standard wildcards (globbing patterns) plus the `**` super-asterisk
+are supported via the underlying globbing library used, https://github.com/gobwas/glob:
```
jbedard, line 31: Can we link to a glob standard instead of the library we are using internally? Is the library adding anything beyond standard globs that needs to be mentioned?

lib/private/copy_to_directory.bzl
```diff
@@ -867,7 +843,13 @@ def copy_to_directory_action(
         replace_prefixes = {},
         allow_overwrites = False,
         is_windows = None):
-    """Helper function to copy files to a directory.
+    """Legacy helper function to copy files to a directory.
```
jbedard, line 846: Should we mark this as deprecated and for removal in 2.0?

lib/private/copy_to_directory.bzl
```diff
@@ -597,12 +587,17 @@ if exist "{src}\\*" (
...
 
-    copy_to_directory_action(
+    copy_to_directory_bin_action(
         ctx,
-        srcs = ctx.attr.srcs,
+        name = ctx.attr.name,
         dst = dst,
+        copy_to_directory_bin = copy_to_directory_bin,
+        files = ctx.files.srcs,
+        targets = [t for t in ctx.attr.srcs if DirectoryPathInfo in t],
```
jbedard, line 600: What if `t` also has files? I think we probably would want to copy both? 🤔

lib/private/copy_to_directory.bzl
gregmagolan: Yes. Will update

lib/private/copy_to_directory.bzl
gregmagolan: I suppose it adds context incase there are issues with globs. This particular library says they are inspired by the "Standard wildcards" but I don't know if it matches the glob standard exactly.

jbedard: I'm also wondering if we should change glob libraries though, so probably best if we don't emphasis which library we're using internally?

gregmagolan: I should switch it out before I land if we're using the wrong one. Docs will be versioned as well so if the library switches again in the future then it will be clear based on the which version of the docs you're looking at.

jbedard: https://github.com/gobwas/glob seems inactive? Last updated 2018 but with 2022 bug reports of crashes.

https://github.com/bmatcuk/doublestar is quite active.

I'm not sure which is faster though...

lib/private/copy_to_directory_toolchain.bzl
gregmagolan: 👍

tools/copy_to_directory_mirror_release.sh
```diff
@@ -0,0 +1,31 @@
+#!/usr/bin/env bash
+# Produce a dictionary for the current yq release,
```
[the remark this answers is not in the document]
gregmagolan: 👍

lib/private/copy_to_directory.bzl
gregmagolan: switching over to https://github.com/bmatcuk/doublestar . it is better maintained and handles one case correctly that the other doesn't: `a` is matched against `a/**`

tools/copy_to_directory_mirror_release.sh
```diff
@@ -0,0 +1,31 @@
...
+    fi
+    integrity="$integrity.sha256"
+    curl --silent --location $integrity -o "/tmp/$release.sha256"
+
+    echo "    \"$release\": \"sha256-$(cat /tmp/$release.sha256 | awk '{ print $1 }' | xxd -r -p | base64)\","
+done
+echo "}"
+
+printf "\n"
+echo "Paste the above into VERSIONS in yq_toolchain.bzl."
```
kormide, line 31: `yq` reference

e2e/copy_to_directory/BUILD.bazel
```diff
@@ -84,7 +83,7 @@ copy_to_directory(
         "//subpkg:b",
     ],
     include_srcs_packages = [
-        "./**",
+        "./*/**",
```
jbedard, line 86: This was required? What is this change?

lib/tests/copy_to_directory/BUILD.bazel
```diff
@@ -82,9 +82,6 @@ copy_to_directory(
         # flatten test_a & test_b to the root
         "test_a": "",
         "test_b": "",
-        # some paths that won't match
```
jbedard, line 85: This was a fix by switching to doublestar?

e2e/copy_to_directory/BUILD.bazel
gregmagolan: this is because `a/**` now matches against `a` so if you want to match against `a/b` and not `a` you now need `a/*/**`

lib/tests/copy_to_directory/BUILD.bazel
gregmagolan: this is because we now glob against the full paths of files even when they are in tree artifacts; previously in Starlark tree artifacts were opaque

tools/copy_to_directory_mirror_release.sh
gregmagolan: 🤦

PR #314, opened 2023-01-06
fix: handle directory symlinks in copy_to_directory_bin_action tool binary

gregmagolan:
Fixes issue reported in Bazel slack thread https://bazelbuild.slack.com/archives/CEZUUKQ6P/p1672965585053839?thread_ts=1671716358.901389&cid=CEZUUKQ6P

Pre-req to landing https://github.com/aspect-build/bazel-lib/pull/311

TL;DR here is that `filepath.Walk` doesn't recurse directory symlinks so if there is a directory symlink in srcs then there was a bug where that was passed to the `copy` function expecting it to be a regular file


tools/copy_to_directory/main.go
```diff
@@ -108,19 +111,87 @@ func longestGlobMatch(g string, test string) (string, error) {
...
-			files = append(files, path)
+func calcCopyDir(cfg *config, copyPaths copyMap, srcPaths pathSet, file fileInfo) error {
+	if srcPaths == nil {
+		srcPaths = pathSet{}
+	}
+	srcPaths[file.Path] = true
+	return filepath.Walk(file.Path, func(p string, info os.FileInfo, err error) error {
+		if info.IsDir() {
+			// filepath.Walk will recurse into directories and return files in subsequent calls so we can
+			// ignore directory entries besides remember them to prevent infinite recursive symlink loops
```
kormide, line 122: I didn't quite follow "ignore directory entries besides remember them to". Typo? Or consider rephrasing?

kormide (approved):

tools/copy_to_directory/main.go
gregmagolan: Just badly worded. Its referring to the `return nil` short circuit below for entries that are directories. I'll tweak it a bit to be more clear.

tools/copy_to_directory/main.go
```diff
@@ -108,19 +111,91 @@ func longestGlobMatch(g string, test string) (string, error) {
...
+			if !path.IsAbs(linkPath) {
+				linkPath = path.Join(path.Dir(p), linkPath)
+			}
+			if srcPaths[linkPath] {
+				// recursive symlink; silently ignore
+				return nil
+			}
+			stat, err := os.Stat(linkPath)
+			if err != nil {
+				return fmt.Errorf("failed to stat file %s pointed to by symlink %s: %w", file.Path, p, err)
```
jbedard, line 146: Quotes around the file paths?

gregmagolan: It will never be an empty path so no quotes is fine here

tools/copy_to_directory/main.go
```diff
@@ -211,11 +286,9 @@ func calcCopyPath(cfg *config, copyPaths copyPaths, file fileInfo) error {
 	dup, exists := copyPaths[outputPath]
 	if exists {
 		if dup.ShortPath == file.ShortPath {
-			// this is likely the same file listed twice: the original in the source
-			// tree and the copy in the output tree
-			// TODO: stat the two files to double check that they are the same
-			if file.RootPath == "" {
-				// when this happens prefer the output tree copy.
+			if file.FileInfo.Size() == dup.FileInfo.Size() && file.RootPath == "" {
```
jbedard, line 289: Why add the size check?

gregmagolan: That is incase the output file is actually different than the source file if there was some transformation it

gregmagolan: Not a checksum but best we can do for cheap

PR #315, opened 2023-01-06
Remove the use of realpath in tests

kzadorozhny:
The `//lib/tests/transitions` tests fail on macOS unless `coreutils` package is installed. Even if the `realpath` is not necessary in this particular case, we should avoid relying on optional packages in tests.

### DIFF
--- a/lib/tests/transitions/test_file_type_contains.sh
+++ b/lib/tests/transitions/test_file_type_contains.sh
@@ -6,4 +6,4 @@
 
 set -o errexit -o nounset -o pipefail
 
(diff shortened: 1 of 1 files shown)


kormide:
Thanks! Looks like this doesn't work on linux unless we pass `-L` to `file`. Otherwise it won't deference the symlink. I hope file on osx has the same option?

kormide:
Looks like `-L` is supported on both. Could you make that change? Actually, the long name `--deference` would be more readable.

kormide (approved):

PR #317, opened 2023-01-07
feat(jq): add a diff_test helper

alexeagle:
This is useful in rules_swc where we want to check that tsconfig.json and .swcrc have matching paths, and most users will want that too

### DIFF
--- a/lib/jq.bzl
+++ b/lib/jq.bzl
@@ -1,6 +1,45 @@
 """Public API for jq"""
 
 load("//lib/private:jq.bzl", _jq_lib = "jq_lib")
+load("@bazel_skylib//rules:diff_test.bzl", "diff_test")
+
+def assert_json_matches(name, file1, filter1, file2, filter2):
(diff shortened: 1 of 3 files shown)


lib/jq.bzl
```diff
@@ -1,6 +1,45 @@
 """Public API for jq"""
 
 load("//lib/private:jq.bzl", _jq_lib = "jq_lib")
+load("@bazel_skylib//rules:diff_test.bzl", "diff_test")
+
+def assert_json_matches(name, file1, filter1, file2, filter2):
```
gregmagolan, line 6: almost seems like we should have an `assert.bzl` for all of the `assert_*` rules... but some like this one require the jq toolchain

gregmagolan (approved):

lib/jq.bzl
alexeagle: We do have testing.bzl with other asserts, but because of the extra dependency here I think it's better to group it with the matching loads

PR #318, opened 2023-01-08
fix: write_source_files incompatibility with bzlmod when writing from external repo

kormide:
This fixes an obscure bug when using `write_source_files` with bzlmod enabled. When you pass in a label originating from an external repository as the "in" file, the `write_source_files` (unnecessarily) tries to convert the string label into a `Label`. This no longer works under bzlmod as aspect_bazel_lib likely doesn't have a repo mapping for this external repository. We can't use the `@@` trick to bypass strict deps because that can only be used to reference the main workspace, not a workspace that the main repo has a mapping for.

This surfaced when converting one of the rules_js e2es to use bzlmod, trying to check in `@npm//repositories.bzl`. bazel-lib doesn't know what `@npm` is so the `Label` constructor complains.

See a related discussion here: https://bazelbuild.slack.com/archives/C014RARENH0/p1673152186072169


e2e/bzlmod_write_source_files_external/BUILD.bazel
```diff
@@ -0,0 +1,10 @@
+load("@aspect_bazel_lib//lib:write_source_files.bzl", "write_source_files")
+
+# This tests that the "in" file to write_source_files can be a
+# label to an external repository target when bzlmod is enabled.
+write_source_files(
+    name = "write_external",
+    files = {
+        "rules_go_license": "@rules_go//:LICENSE.txt",
```
kormide, line 8: TODO: Replace the `rules_go` dep with a test module extension that creates a repository.

lib/private/write_source_file.bzl
```diff
@@ -58,7 +58,6 @@ def write_source_file(
             fail("out_file must be specified if in_file is set")
 
     if in_file and out_file:
-        in_file = utils.to_label(in_file)
         out_file = utils.to_label(out_file)
```
kormide, line 61: The out_file can continue to be converted to a label because it should exist in the main workspace. No processing is done on the in_file and it can just be passed as a string into the rule.

e2e/bzlmod_write_source_files_external/test_extension.bzl
```diff
@@ -0,0 +1,20 @@
...
+
+    rctx.file("foobar.txt", "foobar\n")
+
+test_repo = repository_rule(
+    local = True,
+    implementation = _test_repo_impl,
+)
+
+def _test_ext_impl(_):
+    test_repo(name = "test")
```
kormide, line 16: `new_local_repository` doesn't work inside of a module extension, so I had to make this custom repo rule.

lib/private/write_source_file.bzl
gregmagolan: 👍

lib/private/write_source_file.bzl
```diff
@@ -58,7 +58,6 @@ def write_source_file(
             fail("out_file must be specified if in_file is set")
 
     if in_file and out_file:
```
gregmagolan, line 60: should this be just

`if outfile:`

then?

gregmagolan (approved):

e2e/bzlmod_write_source_files_external/test_extension.bzl
gregmagolan: 👍

PR #319, opened 2023-01-09
feat: modify assert_json_matches api before release

kormide:


lib/testing.bzl
```diff
@@ -70,3 +71,41 @@ def assert_outputs(name, actual, expected):
         file1 = "_expected_ " + name,
         file2 = "_actual_" + name,
     )
+
+def assert_json_matches(name, file1, file2, filter1 = ".", filter2 = "."):
```
alexeagle, line 75: I put this in the jq.bzl file because the load statements there line up. If we put it here, then we need to add the note about setting up the `jq` toolchain for `WORKSPACE` users.

lib/testing.bzl
```diff
@@ -70,3 +71,41 @@ def assert_outputs(name, actual, expected):
...
+    """Assert that the given json files have the same semantic content.
+
+    Uses jq to filter each file. Use `"."` as the filter to compare the whole file.
```
alexeagle, line 78: Should mention "the default value of `"."` means to ..."

alexeagle (approved):

PR #320, opened 2023-01-14
chore: update pre-bazel 6 work-around in utils to only apply for bazel versions < 6

gregmagolan:
Fix in Bazel here https://github.com/bazelbuild/bazel/commit/3f92232c17bf4aad51880c811c291ba4b5b42721 made it into Bazel 6.0.0 release

### DIFF
--- a/lib/private/utils.bzl
+++ b/lib/private/utils.bzl
@@ -76,11 +76,8 @@ def _is_external_label(param):
     Returns:
         a bool
     """
-
-    # Seems like a bug in Bazel that the workspace_root for a label like
-    # @@//js/private/node-patches:fs.js is "external"
-    # See https://github.com/bazelbuild/bazel/issues/16528
-    if str(param).startswith("@@//"):
+    if not _is_bazel_6_or_greater() and str(param).startswith("@@//"):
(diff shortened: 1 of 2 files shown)


lib/private/utils.bzl
```diff
@@ -76,12 +76,6 @@ def _is_external_label(param):
     Returns:
         a bool
     """
-
-    # Seems like a bug in Bazel that the workspace_root for a label like
-    # @@//js/private/node-patches:fs.js is "external"
-    # See https://github.com/bazelbuild/bazel/issues/16528
-    if str(param).startswith("@@//"):
```
alexeagle, line 83: Could you replace it with a hard fail instead, saying that bzlmod only supported with Bazel 6?
Otherwise I suspect users will just get the original confusing error

gregmagolan: Hard fail if `@@` is encountered here and the Bazel version < 6 you mean?

gregmagolan: Hmm. I'll just put the work-around back but condition it on < Bazel 6. That seems least breaking. Thanks for catching that.

kormide (approved):

PR #321, opened 2023-01-14
feat: hardlink generated files in copy_to_directory and copy_directory instead of copying

gregmagolan:
Big impact on performance when since we'll no longer copy files from one location in the output tree to another and just hardlink them instead.

Once this lands and is released, the performance improvement can be applied downstream in rules_js `npm_package` which will make a very big impact on linking performance there. We should finally be as fast as pnpm (or faster) for linking times.

--

Downstream tests:

rules_js: https://github.com/aspect-build/rules_js/pull/789
bazel-examples: https://github.com/aspect-build/bazel-examples/pull/170

Preliminary measurements show noticeable performance gain. Also being tested at Adobe by @kzadorozhny. I'll be running larger benchmarks and getting quantitative results in the near future.


thesayyn (approved):
lgtm.

lib/private/copy_directory.bzl
```diff
@@ -90,14 +98,109 @@ def copy_directory_action(ctx, src, dst, is_windows = None):
...
+        "root_paths": [
+            paths.to_workspace_path(src),
+        ],
+        "verbose": verbose,
+    }
+
+    config_file = ctx.actions.declare_file("{}_config.json".format(name))
+    ctx.actions.write(
+        output = config_file,
+        content = json.encode_indent(config),
```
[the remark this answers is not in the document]
gregmagolan: 👍

lib/private/copy_directory.bzl
```diff
@@ -174,13 +213,30 @@ def copy_directory(name, src, out, **kwargs):
...
+      hardlink: Controls when to use hardlinks to files instead of making copies.
+
+        Creating hardlinks is much faster than making copies of files with the caveat that
+        hardlinks share file permissions with their source.
+
+        Since Bazel removes write permissions on files in the output tree after an action completes,
+        hardlinks to source files within source directories is not recommended since write
+        permissions will be inadvertently removed from sources files.
+
+        - "auto": hardlinks are used is src is a tree artifact already in the output tree
```
jbedard, line 230: "is src is a"

lib/private/copy_directory.bzl
```diff
@@ -158,10 +179,28 @@ _copy_directory = rule(
         # Cannot declare out as an output here, because there's no API for declaring
         # TreeArtifact outputs.
         "out": attr.string(mandatory = True),
+        "hardlink": attr.string(
+            values = ["auto", "off", "on"],
+            default = "auto",
```
jbedard, line 184: Is this a breaking change?

lib/private/copy_directory.bzl
```diff
@@ -158,10 +179,28 @@ _copy_directory = rule(
...
+        ),
+        "verbose": attr.bool(),
+        # use '_tool' attribute for development only; do not commit with this attribute active since it
```
jbedard, line 187: I guess there's no way to check this on commit or a test that fails?

gregmagolan: e2e tests will fail since they don't have rules_go in their WORKSPACES if its let uncommented so we're covered there

lib/private/copy_directory.bzl
gregmagolan: nope; its just an optimization

lib/private/copy_directory.bzl
gregmagolan: 👍

lib/private/copy_directory.bzl
jbedard: Switching to hard links won't cause issues in some cases? 🤔

gregmagolan: Not that I can think of. It should "just work".

PR #323, opened 2023-01-14
refactor: rename copy_to_bin to output_filegroup

gregmagolan:
This naming came from recent client work where switching existing `filegroup` targets to `copy_to_bin` didn't read well and neither did switching to `js_library`. `output_filegroup` seems more descriptive and easier to read since files that are already in the output tree are no-ops and the rule is filegroup-like in that is just exposes a number of files via DefaultInfo except it happens to copy source files to the output tree first.

Speaking of which, it looks like the rule also exposes the file to runfiles which may not be desirable and we should consider removing that in 2.0.

Existing `copy_to_bin.bzl` load points are preserved and can be removed in 2.0.


alexeagle:
I think this is a pretty heavy churn for a low level library so the benefit would have to be really high. Discussing on slack...

alexeagle:
I think we decided not to do this.

PR #324, opened 2023-01-14
Add copy_directory binary tool

gregmagolan:
In order to make copy_directory as optimized as possible and support hardlinks, a binary tool similar to the copy_to_directory binary tool is needed. This follows some of the patterns that the copy_to_directory binary tool uses but it is much simpler as it just takes `src` and `dst` arguments instead of a config file and doesn't support any includes, excludes, replace prefixes, globs etc... Since we're going for maximum speed to be faster than `cp -fR` and to reach parity with pnpm linking time in rules_js, we want to do as little additional work as possible in copy_directory so using the existing `copy_to_directory` binary is not a good option.

Once landed & a binary tool is released, a toolchain will be added in https://github.com/aspect-build/bazel-lib/pull/321


tools/copy_directory/main.go
```diff
@@ -0,0 +1,202 @@
...
+type pathSet map[string]bool
+
+var copyPaths = make(copySet, 0, 10*1024)
+var dstDirs = make(dirSet, 0, 1024)
+var srcPaths = pathSet{}
+
+func scanDir(src string, dst string) error {
+	// filepath.WalkDir walks the file tree rooted at root, calling fn for each file or directory in
+	// the tree, including root. See https://pkg.go.dev/path/filepath#WalkDir for more info.
+	return filepath.WalkDir(src, func(p string, dirEntry fs.DirEntry, err error) error {
```
thesayyn, line 31: nit: should check if the err is nil at the beginning

lib/private/copy_directory.bzl
```diff
@@ -90,6 +90,57 @@ def copy_directory_action(ctx, src, dst, is_windows = None):
...
         _copy_bash(ctx, src, dst)
 
+def copy_directory_bin_action(
+        ctx,
+        src,
+        dst,
+        copy_directory_bin,
+        hardlink = "auto",
+        verbose = False):
+    """Helper function that creates an action to copy a directory from src to dst using a tool binary.
```
alexeagle, line 100: nit: helper -> factory

lib/private/copy_directory.bzl
```diff
@@ -90,6 +90,57 @@ def copy_directory_action(ctx, src, dst, is_windows = None):
...
+    elif hardlink == "auto" and not src.is_source:
+        args.append("--hardlink")
+
+    ctx.actions.run(
+        inputs = [src],
+        outputs = [dst],
+        executable = copy_directory_bin,
+        arguments = args,
+        mnemonic = "CopyDirectory",
+        progress_message = "Copying directory %s" % src.path,
```
alexeagle, line 140: nit from the docs:

>  The message may contain %{label}, %{input}, or %{output} patterns, which are substituted with label string, first input, or output's path, respectively.

so you don't have to explicitly give `src.path` here

tools/copy_directory/main.go
```diff
@@ -0,0 +1,202 @@
...
+		} else {
+			args = append(args, a)
+		}
+	}
+
+	hardlink := false
+	verbose := false
+
+	for _, f := range flags {
+		if f == "--version" || f == "--v" {
```
alexeagle, line 137: nit: short version would usually have a single hyphen so `--version` or `-v`

tools/copy_directory/main.go
```diff
@@ -0,0 +1,202 @@
...
+			flags = append(flags, a)
+		} else {
+			args = append(args, a)
+		}
+	}
+
+	hardlink := false
+	verbose := false
+
+	for _, f := range flags {
```
thesayyn, line 136: nit: since performance is desired, we should merge this loop with the one above.  it would be readable as much IMHO.

tools/copy_directory/main.go
```diff
@@ -0,0 +1,202 @@
...
+		}
+	} else {
+		versionBuilder.WriteString(NoReleaseVersion)
+	}
+	return versionBuilder.String()
+}
+
+func main() {
+	flags := make([]string, 0, len(os.Args))
+	args := make([]string, 0, len(os.Args))
```
alexeagle, line 123: if you really want to optimize, you just want to store the index of the first and second element that don't have a `-` prefix, you could avoid this allocation and copying items

alexeagle (approved):

tools/copy_directory/main.go
thesayyn, line 123: nit: since we expect there to be only two args 

```suggestion
	args := make([]string, 0, 2)
```

tools/copy_directory/main.go
gregmagolan: oh. nice catch

tools/copy_directory/main.go
thesayyn: thinking more about this; maybe this shouldn't be an array at all since there's only two args. how about if we force that `os.Args[1]` is always `src` and `os.Args[2]` is `dst` and followed by flags? then we could get rid of this loop

lib/private/copy_directory.bzl
gregmagolan: nice. will leave a TODO do update all of these in a follow-up

gregmagolan: actually there are only three; will just change here

tools/copy_directory/main.go
```diff
@@ -0,0 +1,202 @@
...
+				FileInfo: info,
+			})
+			return nil
+		}
+	})
+}
+
+// From https://opensource.com/article/18/6/copying-files-go
+func copy(src string, dst string) error {
+	source, err := os.Open(src)
```
thesayyn, line 93: TIL: `os.Open` has `O_RDONLY` mode. nice!

thesayyn (approved):

lib/private/copy_directory.bzl
gregmagolan: 👍

tools/copy_directory/main.go
gregmagolan: 🧙

tools/copy_directory/main.go
gregmagolan: 👍

tools/copy_directory/main.go
gregmagolan: good call. done.

tools/copy_directory/main.go
gregmagolan: that seems problematic especially since I want the `--version` flag to work if that is the first arg. a small array of 2 ints will do the trick

gregmagolan: on 2nd thought lets go with that. `--version` is a special case where it is checked in the first arg only

tools/copy_directory/main.go
```diff
@@ -0,0 +1,200 @@
...
+}
+
+func main() {
+	args := os.Args[1:]
+	if len(args) < 2 {
+		fmt.Println("Usage: copy_directory [src] [dst]")
+		os.Exit(1)
+	}
+
+	if args[0] == "--version" || args[0] == "-v" {
```
thesayyn, line 132: two if conditions should switch places.

thesayyn: checking length of the args would never let this condition be checked.

thesayyn: even better suggestion, how about getting rid of the --version flag and make it part usage message?

gregmagolan: nice catch

gregmagolan: I like it separate so it can be parsed programmatically if needed

PR #332, opened 2023-01-16
feat: add coreutils toolchain

thesayyn:
uutils/coreutils is a rust port of gnu coreutils that works across mac, windows, and linux. 
I discovered it while looking for coreutils alternatives to address https://github.com/bazel-contrib/rules_oci/issues/41. 
I figured since we are doing a lot of bash scripting, this might be useful for more than one ruleset therefore should be here.


thesayyn:
if it lands; this would eliminate`tools/sha256sum` completely within the repo.

alexeagle:
Did you take a look at Tweag rules_sh? Curious if they have anything similar.

kormide:
Do you think it's worth creating a rule to wrap the toolchain?

kormide (approved):

thesayyn:
> Did you take a look at Tweag rules_sh? Curious if they have anything similar.

I looked into rules_sh. it has a similar semantics to the built-in xcode repository bazel creates for cc toolchain. helps binary paths to be included in the input hash if you just use rules_sh. 

However, though rules_nixpkgs, it allows cmds to be fetched hermetically.  so the combination of `rules_nixpkgs` and `rules-_sh` is equivalent to this toolchain.

> Do you think it's worth creating a rule to wrap the toolchain? I guess these typically wouldn't be used as build targets, so maybe not?

i don't see how much value that would add to the toolchain. it's really genrule friendly, so we'd be creating a genrule like rule for this toolchain. Let's not add until we need it?

PR #334, opened 2023-01-16
fix: fix pathing issue in new copy_to_directory binary tool on Windows

gregmagolan:


lib/BUILD.bazel
```diff
@@ -193,8 +193,8 @@ bzl_library(
     srcs = ["repositories.bzl"],
     deps = [
         ":utils",
-        "//lib/private/docs:coreutils_toolchain",
         "//lib/private/docs:copy_to_directory_toolchain",
+        "//lib/private/docs:coreutils_toolchain",
```
gregmagolan, line 197: @thesayyn looks like you need to install precommit hooks on your bazel-lib clone

gregmagolan: https://github.com/aspect-build/bazel-lib/blob/main/CONTRIBUTING.md#formatting

kormide (approved):

lib/BUILD.bazel
thesayyn: oh shoot.

PR #341, opened 2023-01-18
feat: add a module extension for http_archive

kormide:
There is a rules_js e2e I want to convert to bzlmod that uses `http_archive` in the WORKSPACE file. I could just use a workspace file alongside a MODULE.bazel file, but I thought of doing this. Wdyt?

Related to https://github.com/bazelbuild/bazel/issues/17141.

### DIFF
--- a/lib/extensions.bzl
+++ b/lib/extensions.bzl
@@ -6,6 +6,7 @@ load(
     "register_jq_toolchains",
     "register_yq_toolchains",
 )
+load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
 
 def _toolchain_extension(_):
     register_yq_toolchains(register = False)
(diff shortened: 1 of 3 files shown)


lib/extensions.bzl
```diff
@@ -16,3 +17,47 @@ def _toolchain_extension(_):
...
+                name = attr.name,
+                build_file_content = attr.build_file_content,
+                sha256 = attr.sha256,
+                strip_prefix = attr.strip_prefix,
+                urls = attr.urls,
+            )
+
+# Extension that can declare http repository rules, until there's a better way:
+# https://github.com/bazelbuild/bazel/issues/17141
+http = module_extension(
```
alexeagle, line 34: It seems a little gross, but comments on that issue from Keith make me think that someone has to have that code somewhere, so bazel-lib is the best place we have.

@fmeum do you have any opinion about it?

fmeum: Yes, please don't do it :-) It suggests that something works that is inherently broken: The names provided to the extension tag need to be globally unique, which they will be in small projects and then break when composed with other modules.

It's fine if someone builds this for their own module, but having it available in a general purpose skylib-like library makes it more likely these collisions will happen in practice and break the "strict deps" promise of Bzlmod.

I think that we should just accept the overhead of wrapping a bunch of `http_archive`s in a macro and making it a module extension for now and wait for a resolution to https://github.com/bazelbuild/bazel/issues/17141.

PR #342, opened 2023-01-20
feat: add assert helper for zip contents

alexeagle:
Helps to make sure that Python wheels are correctly constructed, as one example.


alexeagle:
Example failure looks like:

```
FAIL: //a/research/projects/text_embedding/inference:check_wheel_contents (see /shared/cache/bazel/user_base/bf7b6accf6f1187bd5511f3fbf7b21b9/execroot/a/bazel-out/k8-fastbuild/testlogs/a/research/projects/text_embedding/inference/check_wheel_contents/test.log)
INFO: From Testing //a/research/projects/text_embedding/inference:check_wheel_contents:
==================== Test output for //a/research/projects/text_embedding/inference:check_wheel_contents:
text_embedding-0.0.2.dist-info/METADeATA

ERROR: above line(s) appeared in expected_wheel_contents but are not present in the archive a/research/projects/text_embedding/inference/text_embedding-0.0.2-py3-none-any.whl
================================================================================
INFO: Elapsed time: 0.320s, Critical Path: 0.03s
INFO: 3 processes: 1 internal, 2 linux-sandbox.
```

kormide (approved):
Cool 👍 . Worth adding some tests? Is `zipinfo` cross-platform?

lib/testing.bzl
```diff
@@ -113,3 +113,53 @@ def assert_json_matches(name, file1, file2, filter1 = ".", filter2 = "."):
             file2,
         ),
     )
+
+def assert_archive_contains(name, archive, expected, type = "zip", **kwargs):
```
kormide, line 117: Should we add a `strip_prefix` or is that just part of `exptected`?

alexeagle: yeah expected would be created like
```
write_file(
    name = "expected_wheel_contents",
    out = "wheel.mf",
    content = [
        "text_embedding/__init__.py",
        "text_embedding/research/projects/text_embedding/inference/__init__.py",
        "text_embedding-0.0.2.dist-info/METADATA",
    ],
)
```
and I think it would be confusing if that didn't always root at the top of the archive file - usually I'd use this test specifically to make sure I didn't get the wrong root in whatever created the archive if a strip_prefix is missing

alexeagle:
I'll add a happy-path case where the test passes, good idea - but it's hard to test the sad path without making the macro complicated.

lib/testing.bzl
kormide: Could you see a use case for `expected` being a string or array of strings? I wonder if we should name it `expected_file` or something of the sort?

alexeagle: I was just adding support for it being the content instead (so you don't need to write a write_file rule in that case).
It makes the error messaging a bit less obvious, but I'll take a few more minutes on it.

kormide (approved):

PR #345, opened 2023-01-27
fix: show directory being copied to in copy_to_directory progress message

jbedard:
I found this useful when I saw a lot of the messages and they were all the exact same. Now it shows things such as:
```
Copying files to directory client/shared/shared
```
(the BUILD is in 'client/shared', with an `npm_package` named 'shared')

Note the use of `short_path` instead of `%{output}` so it doesn't show `bazel-lib/darwin .../` at the start. The second commit changes the copy_file/directory to also use `short_path` for that reason.

Unlike copy_file/directory this PR makes copy_to_directory show the destination, not the source of the copied files because there might be many. We could show the target name, either instead of or in addition to the destination, but I figured the destination is decent?

### DIFF
--- a/lib/private/copy_common.bzl
+++ b/lib/private/copy_common.bzl
@@ -57,3 +57,15 @@ COPY_EXECUTION_REQUIREMENTS = {
     "no-sandbox": "1",
     "local": "1",
 }
+
+def progress_path(f):
+    """
+    Convert a file to an appropriate string to display in an action progress message.
+
+    Args:
+        f: a file to show as a path in a progress message
+
+    Returns:
+        The path formatted for use in a progress message
+    """
+    return f.short_path.removeprefix("../")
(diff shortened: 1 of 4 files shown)


gregmagolan:
short_path will be strange for files from external repositories since it starts with `../`. I think we can improve on that.

lib/private/copy_common.bzl
```diff
@@ -57,3 +57,19 @@ COPY_EXECUTION_REQUIREMENTS = {
...
+    """
+    Convert a path to an appropriate string to display in an action proress message.
+
+    Args:
+        p: a bazel path
+
+    Returns:
+        The path formatted for use in a progress message
+    """
+    for _ in range(10):
```
gregmagolan, line 71: if this function takes a File object instead of a path you can make the impl,
```
return f.short_path.removeprefix("../")
```
since short_path will only ever have a single `../` segment

gregmagolan: function name could be shortened to `progress_path` which is sufficient to imply that it returns one

jbedard: Why will it only ever have one?

gregmagolan: It is a property of short_path that external repos are one down from the cwd

gregmagolan (approved):

PR #349, opened 2023-02-01
copy: limit the number of go routines

jbedard:
@f0rmiga were you thinking similar?

I don't think it's worth getting too complicated, at least for now.

I assume by the time the number of concurrent copies reaches this number the OS/IO operations will far outweigh the go routines, so no point spawning more io operations just to make that worse?

### DIFF
--- a/tools/common/copy.go
+++ b/tools/common/copy.go
@@ -9,8 +9,14 @@ import (
 	"sync"
 )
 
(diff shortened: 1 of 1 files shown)


jbedard:
See https://github.com/aspect-build/bazel-lib/pull/351

PR #350, opened 2023-02-01
release: publish release artifact

alexeagle:
GitHub's stability guarantee for the archive is iffy, and we want metrics on downloads. See https://github.com/bazel-contrib/SIG-rules-authors/issues/11#issuecomment-1412241948

### DIFF
--- a/.bcr/source.template.json
+++ b/.bcr/source.template.json
@@ -1,5 +1,5 @@
 {
   "integrity": "**leave this alone**",
   "strip_prefix": "{REPO}-{VERSION}",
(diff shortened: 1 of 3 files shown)


kormide (commented):
Looks good, but why do we no longer need the workspace snippet?

alexeagle:
oops, it's hard to remember to `git add` from the pool bar

kormide:
I think you also need to change the path to the release snippet script in `release.yml`.

gregmagolan (approved):
🏄

PR #351, opened 2023-02-01
fix: limit copy concurrency

f0rmiga:
On very large directories, the copy functions were creating unbound concurrent copies, which could reach OS limits quickly or even exhaust hardware resources. This patch solves this issue by implementing a queue and a worker pattern.

### DIFF
--- a/tools/copy_directory/main.go
+++ b/tools/copy_directory/main.go
@@ -14,11 +14,14 @@ import (
 type pathSet map[string]bool
 
 var srcPaths = pathSet{}
-var copyWaitGroup sync.WaitGroup
 var hardlink = false
 var verbose = false
 
-func copyDir(src string, dst string) error {
+type walker struct {
+	queue chan<- common.CopyOpts
+}
+
+func (w *walker) copyDir(src string, dst string) error {
 	// filepath.WalkDir walks the file tree rooted at root, calling fn for each file or directory in
 	// the tree, including root. See https://pkg.go.dev/path/filepath#WalkDir for more info.
 	return filepath.WalkDir(src, func(p string, dirEntry fs.DirEntry, err error) error {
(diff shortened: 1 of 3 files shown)


tools/copy_directory/main.go
```diff
@@ -103,8 +104,19 @@ func main() {
...
+	for i := 0; i < numWorkers; i++ {
+		go common.NewCopyWorker(queue).Run(&wg)
+	}
+
+	walker := &walker{queue}
+	if err := walker.copyDir(src, dst); err != nil {
 		log.Fatal(err)
 	}
-	copyWaitGroup.Wait()
+	close(queue)
```
jbedard, line 120: Does closing the queue block until the queue is drained? What's the difference between invoking the `wg.Wait()` before vs after closing the queue?

tools/copy_to_directory/main.go
```diff
@@ -60,7 +59,7 @@ func parseConfig(configPath string) (*config, error) {
 	}
 	defer f.Close()
 
-	byteValue, err := ioutil.ReadAll(f)
+	byteValue, err := io.ReadAll(f)
```
jbedard, line 62: What's the difference between these two?

f0rmiga: `ioutil` is deprecated and is wrapping `io.ReadAll` in this instance, so we can just cut the extra function call.

tools/copy_directory/main.go
f0rmiga: No, closing a channel has some implications:
- No more writes are allowed.
- `for` loops reading the channel will break once drained.

In this case, we close the channel after we know the src directory walk is finished, so we need to wait for the goroutines to finish the copies until we interrupt the program.

f0rmiga: https://go.dev/ref/spec#Close

jbedard (approved):

tools/copy_directory/main.go
f0rmiga: It does. The channel needs to be closed for the `wg.Done` calls to be issued.

PR #354, opened 2023-02-03
fix: fix a bug where `bazel run @jq//:jq` cannot be run on M1

kormide:
On M1 jq will download to a `jq_darwin_amd64` toolchain repo, but the host repo tries to reference `jq_darwin_arm64`, which doesn't exist.


gregmagolan (approved):
🌮

PR #360, opened 2023-02-11
fix: always include files from the same workspace as the build target in `copy_to_directory()`

### LINKED ISSUE #359: [Bug]: `copy_to_directory_bin_action()` creates an empty directory when run in an external workspace.
### What happened?

More context: https://bazelbuild.slack.com/archives/CEZUUKQ6P/p1676010753876439

`npm_package()` from `@aspect_rules_js` doesn't include sources from its own workspace when built in an external workspace. Instead, you get an empty directory ([more context](https://github.com/dgp1130/rules_prerender/issues/48#issuecomment-1425257276)). It seems that the `include_external_repositories` filter of `copy_to_directory` is [a little overzealous](https://github.com/aspect-build/bazel-lib/blob/fe1dc9c734915adbbc6546a725b27c0dd43cd207/tools/copy_to_directory/main.go#L234) and automatically includes sources from the `__main__` workspace, not the workspace containing the target. I think the correct behavior here is to automatically include any files in the _same workspace_ as the `npm_package()` target, and I think that also applies to the underlying `copy_to_directory_bin_action()`.

I think the right fix is to just make `copy_to_directory_bin_action()` add a `--workspace_name` flag containing the name of the workspace for the current target and pass that into the `copy_to_directory` tool. Then the tool should _always_ include files within that workspace regardless of `include_external_repositories` (after applying other filters ofc).

I have a failing test case in https://github.com/dgp1130/aspect_bazel_lib/commit/33e3ed99a7107fa15a0586d90c8e4b8bfb92558e, but haven't started on a proper fix yet.

### Version

Reproducible in latest version of `@aspect_bazel_lib` (1.24.2).

### How to reproduce

```shell
# my_wksp/BUILD.bazel

npm_package(name = "pkg", srcs = ["file.txt"])
```

Then link that from another workspace:

```python
# my_other_wksp/WORKSPACE.bazel

local_repository(
    name = "my_wksp",
    path = "../my_wksp",
)
```

Then build:

```
(cd my_other_wksp/ && bazel build @my_wksp//:pkg)
```

It will output an empty directory.

### Any other information?

My current workaround is to add `include_external_repositories = ["my_wksp"]` to `npm_package()`, but it's not ideal and requires that any downstream workspaces refer to this package's workspace as `@my_wksp` when they should be free to choose any name.

### Fund our work

- [ ] Sponsor our open source work by donating a [bug bounty](https://opencollective.com/aspect-build/)

dgp1130:
Fixes #359.

This updates the `copy_to_directory` tool to accept the workspace name of the target it is executing under and automatically include all source files under that workspace (after apply other filters) regardless of `include_external_repositories`.

I _believe_ all the existing tests should pass, as they do when I manually update `copy_to_directory()` to use the version of the tool built from HEAD. However they don't do that by default, and adding this workspace name option is a breaking change to the existing contract of the `copy_to_directory` tool. It needs to be released at the same time as this fix, unfortunately I'm not sure how to do that or if I even can. @gregmagolan, any suggestions on how to move forward here?


gregmagolan:
Thanks for the PR @dgp1130 .
The sequencing here is tricky with the pre-built golang binary. The golang changes need to land first and then a release cut for those to be released and then the starlark impl updated to use the new version of the golang binary.

I can help with the sequencing after I find a bit of time to review this code. We've been really busy with [Workflows](https://www.aspect.build/workflows) sales in the last month so haven't had as much time as I wanted for OSS work.

dgp1130:
No worries @gregmagolan. Would it be helpful for me to split this into two PRs? One for the Go binary and one with the Starlark changes? That way you could merge them independently.

alexeagle:
Hey @dgp1130 sorry this got parked like this. Yeah if you're willing to make the Golang change in a PR, I'll review/merge/release with that, then the starlark will be able to rely on it being there.

Note to @gregmagolan - I have the same problem in container-structure-test now too. I know we've brainstormed some ways to make the golang library "live from HEAD" but it's probably not time to design that yet.

dgp1130:
No problem @alexeagle, I'm sure you've had more important things to worry about than this. I broke up the Go changes into a [separate PR](https://github.com/aspect-build/bazel-lib/pull/488) and made the new value optional to avoid breaking anything. Once that gets merged and released, we should be able to rebase this and merge it with the changes to the `copy_to_directory` rule.

It's not very important, but IMHO the new workspace name option in the other PR _should_ be required. Unfortunately doing that would break the `copy_to_directory` rule since we can't change both atomically. So once this PR merges and the `copy_to_directory` rule is updated, we could consider a third PR which makes the workspace name option required in the Go tool. I don't know if that tool is used in other places or is considered public API such that a change like that would be considered breaking, but if not it might be worth updating it to make the option required.

dgp1130:
Since #488 got merged, I rebased this PR so it should be mostly good to go. Unfortunately it seems like the CI setup has changed and it isn't running the new workspace properly. Not sure how this is supposed to work or why the directories seem to be missing, but hopefully someone can help point me in the right direction here?

dgp1130:
Thanks! Glad we're able to land this fix!

PR #361, opened 2023-02-11
chore: update to latest version of Aspect CLI and turn on additional workflows features

gregmagolan:
Can't turn on gazelle quite yet as the bzl_library targets are out-of-date and not sure if gazelle is generating the right things there yet

Waiting until Workflows is turned-on again in the back-end for this repo before landing this


alexeagle:
@gregmagolan is it still blocked?

gregmagolan:
This is out-of-date now

PR #366, opened 2023-02-13
fix: incorrect 'find -type' argument used on MacOS

acordiner:
Within `write_source_file()`, an incorrect `find` argument is being used on MacOS to iterate over the copied files. Instead of `find -type f`, it is using `find -type file`, which produces an error:
```
find: Must separate multiple arguments to -type using: ','
```

### DIFF
--- a/lib/private/write_source_file.bzl
+++ b/lib/private/write_source_file.bzl
@@ -196,7 +196,7 @@ fi"""]
         executable_file = "chmod -x \"$out\""
         if is_macos:
             # -x+X doesn't work on macos so we have to find files and remove the execute bits only from those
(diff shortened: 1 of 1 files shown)


gregmagolan:
Hmm. Something is fishy here. I'm able to run
```
$ find . -type file | wc -l
    2642
```
on my macbook without any issues.

gregmagolan:
Oh I see. Some implementations of file are tolerant of the additional chars in `-type` but your change is correct.

gregmagolan (approved):

acordiner:
Thanks @gregmagolan! Any chance to get this released?

PR #371, opened 2023-02-17
build: add MacOS and Windows to CI for landed commits

gregmagolan:


.github/workflows/ci.yaml
```diff
@@ -47,19 +47,44 @@ jobs:
...
+    steps:
+      - uses: actions/checkout@v3
+      - id: linux
+        run: echo "os=ubuntu-latest" >> $GITHUB_OUTPUT
+      - id: macos
+        run: echo "os=macos-latest" >> $GITHUB_OUTPUT
+        # Only run on main branch (not PRs) to minimize macOS minutes (billed at 10X)
+        # https://docs.github.com/en/billing/managing-billing-for-github-actions/about-billing-for-github-actions#included-storage-and-minutes
+        # TODO: DNL
+        # if: ${{ github.ref == 'refs/heads/main' }}
```
alexeagle, line 62: nit: you could use a magic branch name that turns this on, then push your PR from there, then you don't need to comment/uncomment a line for testing

gregmagolan: that's a good idea; will use that next time I'm working on these

.github/workflows/ci.yaml
```diff
@@ -116,19 +177,30 @@ jobs:
...
         env:
           # Bazelisk will download bazel to here
           XDG_CACHE_HOME: ~/.cache/bazel-repo
 
+      - name: bazel test //... (Windows)
+        if: ${{ matrix.os == 'windows-latest' }}
+        working-directory: ${{ matrix.folder }}
+        run: |
+          $BAZEL_MAJOR_VERSION = "${{ matrix.bazelversion }}".substring(0, 1)
+          bazel --bazelrc=${{ github.workspace }}/.aspect/bazelrc/ci.bazelrc --bazelrc=${{ github.workspace }}/.github/workflows/ci.bazelrc --bazelrc=${{ github.workspace }}/.aspect/bazelrc/bazel$BAZEL_MAJOR_VERSION.bazelrc --bazelrc=.bazelrc test --config=${{ matrix.config }} //...
```
kormide, line 198: nit: Can you split this one onto multiple lines like we do in `bazel test //...` so it's easier to parse?

kormide: Perhaps also a comment clarifying why we need a different step to build on windows.

.github/workflows/ci.yaml
```diff
@@ -116,19 +177,30 @@ jobs:
...
+        env:
+          # Bazelisk will download bazel to here
+          XDG_CACHE_HOME: ~/.cache/bazel-repo
+
       - name: integration tests
-        if: ${{ matrix.folder == '.' }}
+        if: ${{ matrix.folder == '.' && matrix.os != 'windows-latest' }}
```
kormide, line 204: Why skip integration tests on Windows?

gregmagolan: They are bash scripts and Windows is running powershell

kormide (approved):

.github/workflows/ci.yaml
gregmagolan: Yes. It looks like you can use backticks to make multiline commands on Powershell

.github/workflows/ci.yaml
gregmagolan: Add comment

PR #372, opened 2023-02-17
chore: turn on more basic precommit checks

alexeagle:
In particular this makes our bazelrc presets more compliant with client codebases, ensuring they can copy these files and not trip on their own pre-commit check.

Ran `pre-commit run -a` - will add this commit to ignore-revs


.pre-commit-config.yaml
```diff
@@ -24,3 +24,15 @@ repos:
     rev: "v2.4.0"
     hooks:
       - id: prettier
+  - repo: https://github.com/pre-commit/pre-commit-hooks
+    rev: v4.4.0
+    hooks:
+      - id: check-added-large-files
```
jbedard, line 30: Are we using bazel-lib as an example and will add these to other repos as well?

jbedard (approved):

.pre-commit-config.yaml
alexeagle: yeah as usual this needs to roll out in 15 PRs or something :(

PR #375, opened 2023-02-19
Changes needed for rules_js bzlmod support

kormide:
In support of https://github.com/aspect-build/rules_js/issues/644

### DIFF
--- a/lib/extensions.bzl
+++ b/lib/extensions.bzl
@@ -9,14 +9,22 @@ load(
 )
 load("//lib/private:host_repo.bzl", "host_repo")
 
-def _toolchain_extension(_):
+def _toolchain_extension(mctx):
     register_copy_directory_toolchains(register = False)
     register_copy_to_directory_toolchains(register = False)
     register_jq_toolchains(register = False)
     register_yq_toolchains(register = False)
-    host_repo(name = "aspect_bazel_lib_host")
+
+    create_host_repo = False
+    for module in mctx.modules:
+        if len(module.tags.host) > 0:
+            create_host_repo = True
+
+    if create_host_repo:
+        host_repo(name = "aspect_bazel_lib_host")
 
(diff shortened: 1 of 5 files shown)


lib/extensions.bzl
```diff
@@ -7,12 +7,14 @@ load(
...
     "register_yq_toolchains",
 )
+load("//lib/private:host_repo.bzl", "host_repo")
 
 def _toolchain_extension(_):
     register_copy_directory_toolchains(register = False)
     register_copy_to_directory_toolchains(register = False)
     register_jq_toolchains(register = False)
     register_yq_toolchains(register = False)
+    host_repo(name = "aspect_bazel_lib_host")
```
gregmagolan, line 17: this repo is preferably opt-in if possible. It has `local = True,` so it will invalidate on every bazel run (by-design)

lib/private/docs.bzl
```diff
@@ -27,7 +27,7 @@ def stardoc_with_diff_test(
         **kwargs
     )
 
-def update_docs(name = "update"):
+def update_docs(name = "update", **kwargs):
```
gregmagolan, line 30: this one related to bzlmod or just a correctness fix?

kormide: It's needed to ignore stardoc targets when bzlmod is enabled.

lib/extensions.bzl
kormide: Added a tag class that needs to be specified to opt in.

kormide:
Ready for another review pass.

alexeagle (approved):
got it, thanks!

lib/extensions.bzl
alexeagle: err, hold on, are we saying every rules_js user is therefore going to have to opt-in to an "invalidate on every run" repo rule?

PR #383, opened 2023-02-28
fix(bazelrc-presets): restore 60s progress rate on CI

alexeagle:
Does this match RH Matt?


gregmagolan (approved):
60s does seem like a better default for CI in general. Buildkite can handle 5s without being too spammy but GHA and CCI are not as nice.

PR #384, opened 2023-03-01
Presets cleanup

kormide:
From https://bazel.build/run/bazelrc#option-defaults

"""
The inheritance (specificity) graph is:
- Every command inherits from common
- The following commands inherit from (and are more specific than) build: test, run, clean, mobile-install, info, print_action, config, cquery, and aquery
- coverage inherits from test
"""

### DIFF
--- a/.aspect/bazelrc/performance.bazelrc
+++ b/.aspect/bazelrc/performance.bazelrc
@@ -41,8 +41,6 @@ build --experimental_reuse_sandbox_directories
 # author.
 # Docs: https://bazel.build/reference/command-line-reference#flag--legacy_external_runfiles
 build --nolegacy_external_runfiles
-run --nolegacy_external_runfiles
-test --nolegacy_external_runfiles
 
 # Some actions are always IO-intensive but require little compute. It's wasteful to put the output
 # in the remote cache, it just saturates the network and fills the cache storage causing earlier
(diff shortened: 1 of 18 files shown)


.aspect/bazelrc/performance.bazelrc
```diff
@@ -52,3 +50,8 @@ test --nolegacy_external_runfiles
 # In some cases you may need to patch rulesets to add a mnemonic to actions that don't have one.
 # https://bazel.build/reference/command-line-reference#flag--modify_execution_info
 build --modify_execution_info=PackageTar=+no-remote
+
+# Build runfiles symlink forests for binaries and tests on demand rather than at build time.
+# This reduces the overhead incurred when running a binary or test isn't required.
+# Docs: https://bazel.build/docs/user-manual#build-runfile-links
+build --nobuild_runfile_links
```
gregmagolan, line 57: @alexeagle Do you think this is a good default for everyone? Suppose it is easy enough to override with `--build_runfile_links` if a test depends on the the runfiles forest of a binary in its deps.

alexeagle: I had the sense that it's pretty breaking. Maybe we should have another file like `performance_risky` that turns on flags that are unlikely to work for everyone.

kormide: In any case, I'll remove it from this PR since only de-duping the options was important for a client.

kormide:
Updated to remove `--nobuild_runfile_links`.

gregmagolan (approved):

PR #388, opened 2023-03-07
Update link to example from angular

garymm:
Previous target was deleted or renamed. Now use a permalink.


gregmagolan (approved):
🙏

gregmagolan:
Thanks for the docs fixup. You just need to run 
```
bazel run @//docs:update
```
to update the generated docs and it should green up CI

garymm:
Done.

PR #394, opened 2023-03-11
fix: disable stardoc under bzlmod and windows by default

alexeagle:
We've been making these exceptions in downstream repos

### DIFF
--- a/lib/BUILD.bazel
+++ b/lib/BUILD.bazel
@@ -1,4 +1,5 @@
 load("@bazel_skylib//:bzl_library.bzl", "bzl_library")
+load("@bazel_skylib//rules:common_settings.bzl", "bool_flag")
 load("//lib/private:stamping.bzl", "stamp_build_setting")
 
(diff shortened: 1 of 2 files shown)


lib/BUILD.bazel
```diff
@@ -15,6 +16,23 @@ exports_files(
...
+# to skip bzlmod-incompatible targets. Note that Bazel does not currently
+# support declaring a config_setting directly on --enable_bzlmod, so we
+# must use set this flag instead.
+
+config_setting(
+    name = "bzlmod",
+    flag_values = {
+        ":flag_bzlmod": "true",
+    },
+)
```
kormide, line 29: So rules_js and other rulesets would use just this bazel-lib flag rather than declaring their own?

kormide (approved):

lib/BUILD.bazel
alexeagle: right, just put up the rules_js change for that, doing rules_esbuild too

PR #397, opened 2023-03-24
feat: detect bzlmod automatically rather than requiring the user sets…

alexeagle:
… the flag.

Thanks to @aherrmann for pointing out this is possible: https://github.com/aherrmann/demo-bazel-detect-bzlmod-config-setting

### DIFF
--- a/lib/private/utils.bzl
+++ b/lib/private/utils.bzl
@@ -202,6 +202,10 @@ def _is_bazel_6_or_greater():
     # native.bazel_version only works in repository rules.
     return "apple_binary" not in dir(native)
 
+def is_bzlmod_enabled():
+    """Detect the value of the --enable_bzlmod flag"""
+    return str(Label("@//:BUILD.bazel")).startswith("@@")
+
 def _maybe_http_archive(**kwargs):
     """Adapts a maybe(http_archive, ...) to look like an http_archive.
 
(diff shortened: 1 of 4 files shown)


lib/private/utils.bzl
```diff
@@ -202,6 +202,10 @@ def _is_bazel_6_or_greater():
     # native.bazel_version only works in repository rules.
     return "apple_binary" not in dir(native)
 
+def is_bzlmod_enabled():
+    """Detect the value of the --enable_bzlmod flag"""
+    return str(Label("@//:BUILD.bazel")).startswith("@@")
```
kormide, line 207: I thought this was a test for Bazel 6, not necessarily --enable_bzlmod?

kormide: Some people may use `BUILD` rather than `BUILD.bazel`. Perhaps we could default to `BUILD.bazel`, but allow an arbitrate label to  be passed?

aherrmann: > I thought this was a test for Bazel 6, not necessarily --enable_bzlmod?

Both branches in the [demo](https://github.com/aherrmann/demo-bazel-detect-bzlmod-config-setting) were tested with Bazel 6. See also the relevant comment [here](https://bazelbuild.slack.com/archives/C014RARENH0/p1678281355216389?thread_ts=1675796308.880269&cid=C014RARENH0).

> Some people may use BUILD rather than BUILD.bazel.

The target `Label` references doesn't actually have to exist as pointed out by @fmeum [here](https://bazelbuild.slack.com/archives/C014RARENH0/p1679653389330709?thread_ts=1679653055.032969&cid=C014RARENH0).

kormide: Neat!

kormide (approved):

lib/private/utils.bzl
alexeagle: May as well change it to //:does-not-exist so it's clear that it's not load-bearing.

PR #399, opened 2023-03-26
feat: highest compatible version selection for bzlmod toolchains

kormide:
Smarter way to select toolchain versions in a bzlmod extension. Was thinking we could use this to select our toolchains such as node.

A module extension implementation that uses this would probably want an option to override the version to what the user has in their root module, in case they want to fix it.

Still a WIP.

### DIFF
--- a/lib/private/extension_utils.bzl
+++ b/lib/private/extension_utils.bzl
@@ -1,0 +1,32 @@
+"Starlark utilities for module extensions"
+
+load("//lib:semver.bzl", "semver")
+
+def highest_compatible_toolchain_version(selected_version, all_versions):
+    """Select the highest compatible version for a toolchain.
+
+    This is useful in a module extension that registers a single toolchain
+    for a module dependency graph.
+
+    Al versions must follow semantic versioning. Preceeding 'v's are supported.
+
+    Args:
+        selected_version: Toolchain version used in the root module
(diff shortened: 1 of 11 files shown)


lib/private/extension_utils.bzl
```diff
@@ -0,0 +1,32 @@
...
+    """Select the highest compatible version for a toolchain.
+
+    This is useful in a module extension that registers a single toolchain
+    for a module dependency graph.
+
+    Al versions must follow semantic versioning. Preceeding 'v's are supported.
+
+    Args:
+        selected_version: Toolchain version used in the root module
+        all_versions: Transitive closure toolchain versions in the module dependency graph
```
kormide, line 15: Maybe I don't need `selected_version` here. Just take the highest from `all_version` but fail if there are any incompatible major versions? 🤔

lib/private/extension_utils.bzl
```diff
@@ -0,0 +1,32 @@
...
+        Most recent version compatible with the selected version.
+    """
+    selected_version = semver.parse(selected_version)
+    all_versions = sorted(
+        [semver.parse(v) for v in all_versions],
+        key = semver.key,
+    )
+
+    if selected_version.major < all_versions[-1].major:
+        fail("Incompatible version")
```
kormide, line 27: Should I not fail but just warn in this case?

PR #405, opened 2023-04-12
Chore: update YQ

henkerik:
Updated the list of available YQ versions. We would like to use a feature in a newer version.  This may be a breaking change since the value of `DEFAULT_YQ_VERSION` is also updated in this PR. 

**Type of change**

- [x] Chore (any other change that doesn't affect source or test files)

- [x] Breaking change (this change will force users to change their own code or config)
- [x] Suggested release notes are provided below:

Default YQ version change from `4.25.2` to `4.33.3`.

**Test plan**

- [x] Covered by existing test cases


gregmagolan (approved):
Thanks for the update 🙏

gregmagolan:
@henkerik Are there breaking API changes in YQ between 4.25.2 to 4.33.3? Some of the existing test cases are failing with the change.

henkerik:
@gregmagolan I made some minor changes to the testcases. You can review them here: 

https://github.com/aspect-build/bazel-lib/pull/405/commits/b040e555d519ed3c9eba0b3dd2d297a045f3fd6f

jbedard:
@henkerik are there release notes describing the breaking changes?

gregmagolan:
Thanks for updating the test cases. Looks like you just have to run `bazel run //docs:update` to green up the docs

gregmagolan:
@henkerik I rebased, updated docs one more time and force pushed. Will land and cut a minor release next.

gregmagolan (approved):
:taco:

alexeagle:
> may be a breaking change since the value of DEFAULT_YQ_VERSION is also updated

It sure was! https://github.com/bazel-contrib/rules_oci/issues/212 tripped on this, so any users of rules_oci who have a newer bazel-lib get an error from yq.

I'm not sure what to do about this yet. I'm tempted to revert the `DEFAULT_YQ_VERSION` and cut another bazel-lib, and/or yank this release from BCR.

Issue #419, opened 2023-05-04
[Bug]: glob behavior doesn't match bazel


alexeagle:
### What happened?

```
filegroup(
    name = "tf",
    srcs = glob(['**/*.tf'])
)

% bazel query --output=build :tf
# /Users/alexeagle/Projects/rules_oci/BUILD.bazel:4:10
filegroup(
  name = "tf",
  srcs = ["//:a.tf", "//:b/b.tf"],
)
```

But the https://github.com/aspect-build/bazel-lib/blob/main/docs/glob_match.md function requires adding `"*.tf"` to get a match on `a.tf` in this case.

### Version

1.29.2

PR #420, opened 2023-05-05
refactor: cleanup glob implementation, add tests

jbedard:
Prefactor to https://github.com/aspect-build/bazel-lib/issues/419

To add more test coverage and cleanup to keep the actual fix minimal.

---

### Type of change

### Test plan

- New test cases added

### DIFF
--- a/lib/private/glob_match.bzl
+++ b/lib/private/glob_match.bzl
@@ -9,37 +9,43 @@ implementation and tests were used as a reference implementation:
     https://github.com/pnpm/pnpm/blob/v7.4.0-2/packages/matcher/test/index.ts
 """
 
-def _split_on(expr, splits):
-    # Splits an expression on the tokens in splits but keeps the tokens split in the result.
-    # Tokens are matched in order so a token such as `**` should come before `*`.
-    result = []
-    accumulator = ""
-    skip = 0
-    has_splits = False
-    for i in range(len(expr)):
-        j = i + skip
-        if j >= len(expr):
-            break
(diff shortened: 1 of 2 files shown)


lib/private/glob_match.bzl
```diff
@@ -13,31 +13,34 @@ def _split_on(expr, splits):
     # Splits an expression on the tokens in splits but keeps the tokens split in the result.
     # Tokens are matched in order so a token such as `**` should come before `*`.
     result = []
-    accumulator = ""
```
jbedard, line 16: There was a bug somewhere in here when doing the `**` work. This diff fixed the issue I was having and seems nicer imo 🤷

thesayyn (approved):
LGTM

lib/tests/glob_match_test.bzl
```diff
@@ -166,6 +200,26 @@ def _mixed_wrapping_globstar(ctx):
 
 mixed_wrapper_globstar_test = unittest.make(_mixed_wrapping_globstar)
 
+def _all_of_ext(ctx):
+    return _glob_match_test(
+        ctx,
+        "**/*.tf",
+        matches = ["a/b.tf", "ab/cd/e.tf"],
```
gregmagolan, line 207: This was the one that broke the release right? It should also match a .tf file at the root: `"c.tf"`

jbedard: Yeah, and this PR doesn't fix it. This just does some prefactors before the fix.

PR #421, opened 2023-05-05
fix(yq): restore default version, reverting #405

alexeagle:
The new version of yq is quite breaking, as it requires '-oj -r' for continued operation in rules_oci. We can do a default version flip later when more downstream projects are compatible.

---

### Type of change

Anyone relying on the new version being the default will have to opt-in. This has been live only four days so I think that's okay. I'll yank the prior release from BCR.

### Test plan

None

### DIFF
--- a/lib/private/yq_toolchain.bzl
+++ b/lib/private/yq_toolchain.bzl
@@ -48,7 +48,9 @@ YQ_PLATFORMS = {
     ),
 }
 
-DEFAULT_YQ_VERSION = "4.33.3"
+# Note: this is not the latest release, because it has significant breaking changes.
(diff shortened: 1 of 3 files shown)


lib/private/yq_toolchain.bzl
```diff
@@ -48,7 +48,7 @@ YQ_PLATFORMS = {
     ),
 }
 
-DEFAULT_YQ_VERSION = "4.33.3"
+DEFAULT_YQ_VERSION = "4.25.2"
```
gregmagolan, line 51: Can you add a code comment here as to why the DEFAULT is not the latest in the list?

gregmagolan (approved):

PR #422, opened 2023-05-06
fix: rewrite glob implementation to follow standard glob specs

jbedard:
This is basically a rewrite that @thesayyn and I did using the [go doublestar](https://github.com/bmatcuk/doublestar/blob/465a339d8daa03b8620e49b8ae541f71651426ad/match.go#L74) implementation as a guide. Probably not worth looking at the diff in the second commit and just looking at the code.

### Type of change

### Test plan

- New test cases added


alexeagle (approved):
Not reviewing too carefully since two of you paired on it. Thanks!

Issue #418, opened 2023-05-03
[FR]: ord, chr, hex


aryeh-looker:
### What is the current behavior?

The canonical starlark implementation does not have these methods.

### Describe the feature

Implementing the Python functions ord, chr, hex. At the moment, I don't see a way to do this code-points (ord) without a lookup table; but it's certainly possible. `url_encode` and `url_decode` could also be neat; e.g. ` ` becomes `%20`.

alexeagle:
Related, I think: https://github.com/bazelbuild/starlark/issues/112

aryeh-looker:
```
% bazel run src/main/java/net/starlark/java/cmd:starlark
>> ord 
<stdin>:1:1: name 'ord' is not defined
>> hash('a')
97
>> hash('/')
47
% python3 -c "print(ord('a'), ord('/'))" 
97 47
```

thesayyn:
fixed by #425

PR #425, opened 2023-05-08
feat: implement chr, ord, and hex

thesayyn:
### Type of change

- New feature or functionality (change which adds functionality) #418 

### Test plan

- New test cases added


alexeagle (approved):

lib/string.bzl
```diff
@@ -0,0 +1,9 @@
+
+
+"Utilities for strings"
```
gregmagolan, line 3: nit: the comment in strings.py is now out of date since it refers to base64.bzl

gregmagolan (approved):
🌮

alexeagle:
Another nit: how did we decide on `string.bzl` as the public API rather than `strings.bzl` ?

thesayyn:
> Another nit: how did we decide on `string.bzl` as the public API rather than `strings.bzl` ?

we haven't actually. open to suggestions. `strings.bzl` sounds more appropriate to me.

thesayyn:
to get this going; assuming Alex and I agree on `strings.bzl`, should I rename this to strings.bzl?

alexeagle:
yes please, let's go with that. Unless you can find some other starlark implementation that already has a name for a module that provides these (but I expect there isn't since they're proposed as builtin keywords)

thesayyn:
> but I expect there isn't since they're proposed as builtin keywords

true. renamed to strings

PR #426, opened 2023-05-09
fix: support special chars in assert_contains expected expression

jbedard:
See https://github.com/aspect-build/rules_js/issues/1045

### Type of change

### Test plan

- New test cases added


lib/testing.bzl
```diff
@@ -22,24 +22,31 @@ def assert_contains(name, actual, expected, size = None, timeout = None, **kwarg
     """
 
     test_sh = "_{}_test.sh".format(name)
+    expression_txt = "_{}_expression.txt".format(name)
```
jbedard, line 25: Any internal file naming preferences or suggestions?

kormide: `expected`? Does it need the `.txt` extension?

kormide (approved):

lib/testing.bzl
jbedard: There's really no restrictions on the name as long as it's unique to the target. I figured an extension is nice when browsing bazel-bin, but `.txt` feels odd 🤷 

`expected` sgtm though 👍

gregmagolan: +1 to no extension

jbedard: Note that `assert_outputs` also uses .txt 🤷

gregmagolan (approved):

lib/testing.bzl
gregmagolan: Its a nit either way since it is not user facing

jbedard: I'm going to stick with .txt since the others in this file use it. We can change them all in the future if anyone wants to.

Issue #431, opened 2023-05-14
[FR]: expand_template should support replacements, stamping


alexeagle:
### What happened?

We depend on a private one here
https://github.com/aspect-build/rules_js/blob/main/js/private/expand_template.bzl#L115
in some Aspect projects and some of our clients, because it's such a useful utility.

It lives in rules_js because @gregmagolan needed to write a process to perform an action.

It ought to come from bazel-lib instead. Like copy_to_directory, we should port that JS program to a minimal Go program to perform the action, and distribute it as a pre-built toolchain.

Note, https://github.com/bazelbuild/bazel-skylib/issues/442 asks for the same thing on bazel-skylib

### Version

HEAD

thesayyn:
fixed by #435

PR #435, opened 2023-05-16
feat: support stamp substituions

thesayyn:
### Type of change

- New feature or functionality (change which adds functionality) #431 

### Test plan

- New test cases added


thesayyn:
made sure that failures are isolated to missing toolchain hashes/artifacts for expand_template.

tools/expand_template/BUILD.bazel
```diff
@@ -0,0 +1,33 @@
...
+    ],
+)
+
+go_binary(
+    name = "expand_template",
+    embed = [":expand_template_lib"],
+    visibility = ["//visibility:public"],
+)
+
+# For testing purposes.
```
thesayyn, line 21: TODO: will remove in the upcoming release. doesn't affect the toolchain resolution for consumers even if we release it as is.

tools/expand_template/main.go
```diff
@@ -0,0 +1,96 @@
...
+	"log"
+	"os"
+	"regexp"
+	"strings"
+
+	"github.com/aspect-build/bazel-lib/tools/common"
+	"golang.org/x/exp/maps"
+)
+
+var STATUS_REGEX = regexp.MustCompile(`(?m)^([^\s]+)\s+([^\n ]*)$`)
```
f0rmiga, line 15: A little comment would help people who are not proficient in regular expressions to understand what's going on here.

tools/expand_template/main.go
```diff
@@ -0,0 +1,96 @@
...
+
+func parseStatusFile(statusFilePath string) (map[string]string, error) {
```
f0rmiga, line 17: The best practice is to have the `main` function first, then it reads from top to bottom in terms of execution. This is true for pretty much any language (at least the good ones).

tools/expand_template/main.go
```diff
@@ -0,0 +1,96 @@
...
+	}
+
+	return results, nil
+
+}
+
+func main() {
+	args := os.Args[1:]
+
+	if len(args) == 1 {
```
f0rmiga, line 38: Please, use the standard library to parse flags. Here's an example: https://gobyexample.com/command-line-flags.

tools/expand_template/main.go
```diff
@@ -0,0 +1,96 @@
...
+		if args[0] == "--version" || args[0] == "-v" {
+			fmt.Printf("expand_template %s\n", common.Version())
+			return
+		}
+	}
+
+	if len(args) != 6 {
```
f0rmiga, line 45: The same applies here in regard to flag parsing. Instead of positional arguments, use flags.

tools/expand_template/main.go
```diff
@@ -0,0 +1,96 @@
...
+	executable := args[5] == "true" || args[5] == "1"
+
+	substitutions := map[string]string{}
+	substitutionJson, err := os.ReadFile(args[2])
+	if err != nil {
+		log.Fatal(err)
+	}
+	if err := json.Unmarshal(substitutionJson, &substitutions); err != nil {
+		log.Fatal(err)
+	}
```
f0rmiga, line 59: This can be a separate function:

```go
func parseSubstitutions(substitutionsFile string) (map[string]string{}, error) {
	f, err := os.Open(substitutionsFile)
	if err != nil {
		return nil, fmt.Errorf("failed to parse substitutions file: %w", err)
	}
	defer f.Close()
	decoder := json.NewDecoder(f)
	out := make(map[string]string{})
	if err := decoder.Decode(&out); err != nil {
		return nil, fmt.Errorf("failed to parse substitutions file: %w", err)
	}
	return out, nil
}
```

tools/expand_template/main.go
```diff
@@ -0,0 +1,96 @@
...
+	if err != nil {
+		log.Fatal(err)
+	}
+	content := string(contentBytes)
+
+	for key, value := range substitutions {
+		content = strings.ReplaceAll(content, key, value)
+	}
+
+	var mode os.FileMode = 0x666
```
f0rmiga, line 91: Why not `0600`?

tools/expand_template/main.go
```diff
@@ -0,0 +1,96 @@
...
+	if executable {
+		mode = 0o777
```
f0rmiga, line 93: Why not `0700`?

tools/expand_template/main.go
```diff
@@ -0,0 +1,96 @@
...
+	}
+	os.WriteFile(args[1], []byte(content), mode)
```
f0rmiga, line 95: This could fail, need to check for error.

f0rmiga (dismissed):

tools/expand_template/main.go
thesayyn: Already talked with @gregmagolan about it. we don't want to spend more CPU cycles to parse args that we control at both ends. no tool in the //tools directory uses the flag package to parse these.

tools/expand_template/main.go
thesayyn: see above.

tools/expand_template/main.go
thesayyn: that looks more complicated and harder to understand. Could you elaborate more on why?

tools/expand_template/main.go
thesayyn: This is ported from https://github.com/aspect-build/rules_js/blob/main/js/private/expand_template.js. @gregmagolan  might know the answer to that.

tools/expand_template/main.go
thesayyn: same as above.

tools/expand_template/main.go
thesayyn: thanks. will fix it.

tools/expand_template/main.go
gregmagolan: It matches the permissions that Bazel sets in the output tree

gregmagolan:
Lets create a new `expand_template.bzl` load point for this rule with the PR and a new docs file for that so it shows up separately in the docsite. The `expand_make_vars` load point can stick around for now for backward compat with a `# TODO(2.0): remove` comment on it.

thesayyn:
> Lets create a new `expand_template.bzl` load point for this rule with the PR and a new docs file for that so it shows up separately in the docsite. The `expand_make_vars` load point can stick around for now for backward compat with a `# TODO(2.0): remove` comment on it.

done

gregmagolan (approved):
:shipit:

PR #436, opened 2023-05-16
feat: introduce expand_template tool

thesayyn:
work towards #431 

### Type of change

### DIFF
--- a/tools/BUILD.bazel
+++ b/tools/BUILD.bazel
@@ -16,11 +16,18 @@ multi_platform_go_binaries(
     tags = ["manual"],
 )
 
+multi_platform_go_binaries(
+    name = "expand_template",
+    embed = ["//tools/expand_template:expand_template_lib"],
+    tags = ["manual"],
+)
+
 release(
     name = "release",
     tags = ["manual"],
     targets = [
         ":copy_directory",
         ":copy_to_directory",
+        ":expand_template"
     ],
 )
--- a/tools/expand_template/BUILD.bazel
+++ b/tools/expand_template/BUILD.bazel
@@ -1,0 +1,19 @@
+load("@io_bazel_rules_go//go:def.bzl", "go_binary", "go_library")
+load("//lib/private:expand_template_toolchain.bzl", "expand_template_toolchain")
+
+go_library(
+    name = "expand_template_lib",
+    srcs = ["main.go"],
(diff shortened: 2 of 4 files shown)


MODULE.bazel
```diff
@@ -13,12 +13,14 @@ bazel_dep(name = "stardoc", repo_name = "io_bazel_stardoc", version = "0.5.0")
 
 ext = use_extension("@aspect_bazel_lib//lib:extensions.bzl", "ext")
 
-use_repo(ext, "copy_directory_toolchains", "copy_to_directory_toolchains", "jq_toolchains", "yq_toolchains", "coreutils_toolchains")
+use_repo(ext, "copy_directory_toolchains", "copy_to_directory_toolchains", "jq_toolchains", "yq_toolchains", "coreutils_toolchains", "expand_template_toolchains")
```
gregmagolan, line 16: save the registrations for the follow-up? guessing these will be breaking since the shas binaries are not up yet

lib/extensions.bzl
```diff
@@ -16,6 +17,7 @@ def _toolchain_extension(mctx):
     register_jq_toolchains(register = False)
     register_yq_toolchains(register = False)
     register_coreutils_toolchains(register = False)
+    register_expand_template_toolchains(register = False)
```
gregmagolan, line 20: same here

lib/repositories.bzl
```diff
@@ -161,3 +165,25 @@ def register_copy_to_directory_toolchains(name = "copy_to_directory", register =
         name = "%s_toolchains" % name,
         user_repository_name = name,
     )
+
+
+def register_expand_template_toolchains(name = "expand_template", register = True):
```
gregmagolan, line 170: same this one for the follow-up as well since it makes it into the docs

MODULE.bazel
thesayyn: it doesn't break since nobody is requesting the toolchain.

lib/repositories.bzl
thesayyn: um, yeah I'll do that then.

MODULE.bazel
thesayyn: done.

thesayyn:
done. it's only about adding the tool to the release process.

gregmagolan (approved):
🌮

PR #447, opened 2023-06-05
feat: add freebsd support

rickystewart:
### Type of change

Add basic support for freebsd_amd64.

### Test plan

Ran `bazel build tools:{copy_to_directory,copy_directory}`, confirmed that a FreeBSD executable is generated


alexeagle (approved):
Thanks! I guess we'll have to verify these changes work properly when we cut the next release...

PR #464, opened 2023-06-22
Use actual newline (instead of "\n") to split strings in yq rule

### LINKED ISSUE #463: [Bug] Support stamping on yq rule not working when stamp coming from running a "non-mocked" workspace_status script
Hi guys, 
Big thanks for this feature very vey useful!!

I've tried it and i have an issue that i replicate [here](https://github.com/opicaud/bazel-lib) with a red test when running
`bazel test --config=release //lib/tests/yq/...`
It's unfortunately not working when i stamp a value from stable-status.txt when I use the release config with bazel.
For info i am running on Bazel 5.4.0

_Originally posted by @opicaud in https://github.com/aspect-build/bazel-lib/issues/223#issuecomment-1603183000_

malt3:
Fixes d79f4d4
Fixes #463

This is required since versions of yq older than 4.30.3 do not parse "\n" correctly. See https://github.com/mikefarah/yq/issues/1430 and https://github.com/mikefarah/yq/commit/e02bb7194897edb916a6787fb8f8e5a7f432dd94

---

### Type of change

### Test plan

1. Build any yq rule: `bazel build //lib/tests/yq:stamped`
2. Look at the processed stamping yaml: `cat bazel-bin/lib/tests/yq/_stamped_stamp.yaml`

Without this fix, the file only contains the first item of `bazel-out/stable-status.txt` and `bazel-out/volatile-status.txt` respectively. With the fix, every item appears in the yaml.

### DIFF
--- a/lib/private/parse_status_file.yq
+++ b/lib/private/parse_status_file.yq
@@ -1,3 +1,4 @@
-load_str(filename) | split("\n") | .[] | select(length!=0)
+load_str(filename) | split("
+") | .[] | select(length!=0)
     | [capture("(?P<key>[^\s]+)\s+(?P<value>.*)")]
     | from_entries


opicaud (approved):
Confirmed that the provided red test is now fully green with this change

alexeagle:
Wow that was silly. I was debugging this to fix it myself, and stumbled on this thread without realizing it's a PR on bazel-lib :)

alexeagle (approved):

PR #468, opened 2023-07-07
feat: add a tar toolchain

### LINKED ISSUE #470: [FR]: Provide a tar toolchain

rules_pkg has a bunch of issues with creating tar files that include binaries:

- https://github.com/bazelbuild/rules_pkg/issues/153
- https://github.com/bazelbuild/rules_pkg/issues/579
- https://github.com/bazelbuild/rules_pkg/issues/161
- https://github.com/bazelbuild/rules_pkg/issues/703
- https://github.com/bazelbuild/rules_pkg/issues/642
- https://github.com/bazelbuild/rules_pkg/issues/715

It would be great to have rules_pkg solve these, but many are old.
rules_oci users keep running into this because rules_docker did its own runfiles handling, which rules_oci "fixed the layering violation" by expecting the pkg_tar to create a correct output.

### Describe the feature

Instead of waiting for rules_pkg, we could build an alternative. It should be simple:

- We already have https://docs.aspect.build/rules/aspect_bazel_lib/docs/copy_to_directory/ - we can fix this to have correct handling of symlinks and runfiles. As a result, there should be a folder in `bazel-out` that looks exactly like what we want in the `tar` file.
- Add a toolchain that provides a `tar` program with predictable behavior and arguments - in practice any GNU tar that's "recent enough" works. https://github.com/aspect-build/bazel-lib/compare/main...tar is a start.

alexeagle:
This provides an alternative to the pkg_tar rule from rules_pkg. That has the advantage that we can handle runfiles correctly from the outset and be a better recommendation for rules_oci users.

Fixes #470 

---

### Type of change

There is now a tar rule to package files into tar archives. Users of pkg_tar may want to consider this as a simpler alternative. See tar.md

### Test plan

- New test cases added


thesayyn:
ah, https://man.freebsd.org/cgi/man.cgi?mtree(8)  
https://man.freebsd.org/cgi/man.cgi?query=bsdtar&sektion=1&format=html#EXAMPLES

alexeagle:
@kormide @thesayyn should we land this on `main` or hold it for the 2.0 release so there's a compelling new feature to go along with the major release?

lib/private/tar_toolchain.bzl
```diff
@@ -0,0 +1,241 @@
...
+    "linux_amd64": struct(
+        compatible_with = [
+            "@platforms//os:linux",
+            "@platforms//cpu:x86_64",
+        ],
+    ),
+    "linux_arm64": struct(
+        compatible_with = [
+            "@platforms//os:linux",
+            "@platforms//cpu:x86_64",
```
kormide, line 15: Should this be `cpu:aarch64`?

lib/tar.bzl
```diff
@@ -0,0 +1,87 @@
...
+  https://github.com/bazelbuild/rules_pkg/pull/238
+- Does not have any custom program to produce the output, instead
+  we rely on a well-known C++ program called "tar".
+  Specifically, we use the BSD variant of tar since it provides a means
+  of controlling mtimes, uid, symlinks, etc.
+
+We also provide full control for tar'ring binaries including their runfiles.
+
+TODO:
+- Ensure we are reproducible, see https://reproducible-builds.org/docs/archives/
```
kormide, line 17: Were you intending to leave this one as a TODO before mergin?

kormide:
> @kormide @thesayyn should we land this on `main` or hold it for the 2.0 release so there's a compelling new feature to go along with the major release?

I think that's a good idea. Unless rules_oci needs it sooner.

lib/tar.bzl
alexeagle: We already give reproducible timestamps, so I think we are okay to ship a first version, but the TODO is still here since there's a section on posix headers in the pax format that we should still research some more.

alexeagle:
rules_oci users have been waiting 9 months, so a couple more weeks is fine.

That also lets us give `tar` some testing as a non-public API while we have a 2.0-beta series of releases, only committing to it for the first RC

kormide (approved):

PR #471, opened 2023-07-13
chore: add queue to workflows config

kormide:
Needed for workflows runners to pick up tasks.

### Type of change


alexeagle (approved):

kormide:
Not needed anymore. I was working off of an older `main` locally that didn't have this.

PR #485, opened 2023-07-26
chore: document JQ_BIN make var

alexeagle:
### Type of change

### Test plan

None, this is copied from an executable (green) example


mattem (approved):
💙

PR #488, opened 2023-07-30
feat: update `copy_to_directory` tool to accept the name of its workspace and automatically include files from it

dgp1130:
Refs #359.

This allows `copy_to_directory` to be run in an external workspace and include files from that workspace automatically once the `copy_to_directory` rule is updated. The rule update needs to be separated from the tool update because the tool is pre-built and vendored rather than built at HEAD.

---

This was split off from https://github.com/aspect-build/bazel-lib/pull/360 in order to land the Go changes first and release a new version of the tool.

The new workspace name argument needs to be optional to avoid breaking the `copy_to_directory` rule. Once the rule is updated and released, we should consider making this argument required if the tool is not used elsewhere.

I didn't see any direct tests or documentation of the tool itself, so I left those for https://github.com/aspect-build/bazel-lib/pull/360 which will test/document the `copy_to_directory` rule itself. Please let me know if I missed something.

I'm also not totally sure if this should be classified as a user-visible `feat` commit or a simple `refactor`. Probably depends on whether the tool itself is considered public API independent of the `copy_to_Directory` rule?

### Type of change

### DIFF
--- a/tools/copy_to_directory/main.go
+++ b/tools/copy_to_directory/main.go
@@ -44,6 +44,7 @@ type config struct {
 	Verbose                     bool              `json:"verbose"`
 
 	ReplacePrefixesKeys []string
+	TargetWorkspace     *string
 }
 
 type copyMap map[string]fileInfo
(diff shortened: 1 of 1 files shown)


alexeagle:
Thanks! @gregmagolan is the author here so I'd like an 👁️ from him

alexeagle:
ah, after https://github.com/aspect-build/bazel-lib/pull/532 lands we won't have to do this "release in two halves" dance anymore, so your original PR could land as-is without splitting.

dgp1130:
Since this got merged, don't we still need to merge #360 as well? Otherwise the workspace name is never passed into the tool.

PR #496, opened 2023-08-15
yq: allow any filename in srcs

pjjw:
we have a number of files in our repository that are yaml, despite not having a .yaml suffix in their filename, and this filter prevents us from using the yq rule on them.

---

### Type of change

### Test plan

### DIFF
--- a/lib/private/yq.bzl
+++ b/lib/private/yq.bzl
@@ -4,7 +4,7 @@ load("//lib:stamping.bzl", "STAMP_ATTRS", "maybe_stamp")
 
 _yq_attrs = dict({
     "srcs": attr.label_list(
-        allow_files = [".yaml", ".json", ".xml"],
+        allow_files = True,
         mandatory = True,
         allow_empty = True,
     ),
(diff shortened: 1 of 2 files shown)


kormide:
I think that we automatically set some `yq` flags based on the file types seen in `srcs`. See https://github.com/aspect-build/bazel-lib/blob/main/lib/yq.bzl#L165.

It would be great if you could add some tests for different input types but without extensions to test that nothing broke there.

kormide:
> @kormide sure, added for a smattering of test cases for yaml and json with different-than-expected file extensions. seems to work fine, PTAL and see if there's some combination i missed here.

Thanks. LGTM with a minor nit: can you remove the `.whatever`/`.dealerschoice` extensions since the name of the test targets imply that they are testing no extension on the input file?

pjjw:
how about i change the test names instead? really meant wrong or different- "no" in the sense of "no hints from the extension". naming!

kormide (approved):

pjjw:
thanks! meant to circle back and fix that up.

alexeagle:
Thanks for contributing @pjjw !!

PR #504, opened 2023-08-26
chore: add test coverage for run_binary & expand_variables makevars

gregmagolan:
Related to https://bazelbuild.slack.com/archives/CEZUUKQ6P/p1693001775404789 thread on Bazel Slack

### Type of change

### Test plan

- New test cases added


lib/private/run_binary.bzl
```diff
@@ -141,11 +141,17 @@ def run_binary(
...
+            and
+            [expand_make_vars](https://docs.aspect.build/rules/aspect_bazel_lib/docs/expand_make_vars).
 
         env: Environment variables of the action.
 
-            Subject to `$(location)` and makevar expansions.
+            Subject to `$(location)` and make variable expansions via
+            [expand_location](https://docs.aspect.build/rules/aspect_bazel_lib/docs/expand_make_vars#expand_locations)
+            and
+            [expand_make_vars](https://docs.aspect.build/rules/aspect_bazel_lib/docs/expand_make_vars).
```
jbedard, line 154: Can these be relative links? Just `[expand_location](./expand_make_vars)`?

jbedard (approved):

lib/private/run_binary.bzl
gregmagolan: Yea, that will work better 👍

Issue #268, opened 2022-10-27
Support jq on arm64 linux


keith:
The last official jq release, and the one these rules are currently using, was released in 2018, it sounds like the project is still alive https://github.com/stedolan/jq/issues/2305 but changes aren't happening quickly. There are some upstream issues around providing binaries for arm64 linux https://github.com/stedolan/jq/issues/1655, but so far there aren't any hosted binaries we can easily download for this.

One option to solving this would be to build jq with bazel instead of vendoring binaries. A quick google came up with these BUILD files https://github.com/attilaolah/bazel-tools/blob/master/jq.bazel which I have not tested. Another option would just be to build an arm64 binary from the last release tag and host it somewhere (like on the releases page of this repo), just to have a usable download.

kormide:
Note that our [`yq`](https://github.com/aspect-build/bazel-lib/blob/main/docs/yq.md) toolchain has an `arm64` binary. Yq is supports json and is a superset of jq.

thesayyn:
Idea; https://wapm.io/syrusakbary/jq has jq as a wasm binary and can run anywhere `wasmer` can run which is basically every nix platform plus windows. wasmer can be introduced as a toolchain and jq ran with it.

keith:
That sounds like a bit of overhead? i think the worst case alternative is we build jq manually, or host our own binary for that case?

kormide:
Jq arm64 linux support will be in the `2.0.0` release.

rcorre:
>  Jq arm64 linux support will be in the 2.0.0 release.

Is there an issue tracking this somewhere? I found https://github.com/jqlang/jq/issues/2386, which has a 1.7 milestone, and https://github.com/jqlang/jq/issues/2459, which doesn't have any milestone.

I'm trying to figure out what's necessary for https://github.com/bazel-contrib/rules_oci/issues/253 to be fixed.

thesayyn:
https://github.com/jqlang/jq/releases/tag/jq-1.7 is the available binaries as part of 1.7 release. i believe we are waiting for 2.0 release to make this breaking change.

rcorre:
Ah, so 2.0 of bazel-lib? What makes this a breaking change?

kormide:
> Ah, so 2.0 of bazel-lib? What makes this a breaking change?

Yup. We're planning to cut a bazel-lib 2.0 release in mid-late October. There were a couple of breaking changes:
[...]

PR #510, opened 2023-09-02
JQ 1.7rc2 toolchain

gleyba:
This will fix the issue with jq not available on linux arm64:
https://github.com/aspect-build/bazel-lib/issues/268


alexeagle:
Thanks, I think we're going to solve this in #520 instead

Issue #495, opened 2023-08-12
[FR]: Add utillities from `bzlparty/lib_common`


mgred:
### Describe the feature

This is a proposal to add some/all functions from [`bzlparty/lib_common`](https://github.com/bzlparty/lib_common/blob/main/defs.bzl). The motivation comes from [this conversation](https://github.com/bazelbuild/bazel-central-registry/pull/727#issuecomment-1622386889). I follow @alexeagle advice to propose the code as a new feature to collect everybody's opinion and perspective. 

My personal motivation was to have a place to collect generic functions I found useful while writing rulesets.

@kormide  I was told that you have done a lot of work regarding the public API and maybe could help integrating the functionality.

I would love to work with you guys and hear what you have to say.

I totally appreciate any feedback and opinion on this

alexeagle:
1. Xudong has good feedback on that BCR PR that we should consider whether the syntax sugar is worth it, vs. using the starlark language directly.

> the readability of any code that potentially depends on it

For example your 

```
occurs(2, [1, 2, 3]) // True
occurs("foo", ["bar", "baz"]) // False
```

should probably be written
```
2 in [1, 2, 3] # True
"foo" in ["bar", "baz"] # False
```

And I think `len(filter(f, arr)) == 1` is easier to read than `once(f, arr)` especially since it allows for expressing "exactly two matching elements"

2. Other than `occurs` your lib is basically functional programming idioms that are possible thanks to the addition of `lambda` to the language. Maybe it could be `functions.bzl` or `lambdas.bzl`

3. We should nitpick the names from whether there's a popular idiom for python developers, so we don't accidentally introduce more differences between that and starlark

4. We should figure out a place to "dogfood" this so we have the experience of being a user of the new API. Maybe we can refactor other code in bazel-lib to use these.

mgred:
Thanks for your feedback @alexeagle 

> 1. Xudong has good feedback on that BCR PR that we should consider whether the syntax sugar is worth it, vs. using the starlark language directly.
> 
> 
> > the readability of any code that potentially depends on it
> 
> For example your
> 
> ```
> occurs(2, [1, 2, 3]) // True
> occurs("foo", ["bar", "baz"]) // False
> ```
> 
> should probably be written
> 
> ```
> 2 in [1, 2, 3] # True
> "foo" in ["bar", "baz"] # False
> ```

[...]

PR #512, opened 2023-09-05
feat: add list utils

mgred:
This adds functions to work with lists from [bzlparty/lib_common](https://github.com/bzlparty/lib_common/blob/main/defs.bzl) as discussed in #495.

I removed the `occurs` functions, this is not needed because of the `in`-operator.

Please let me know if I should add more tests/documentation.

### Type of change

### Test plan

- New test cases added

### DIFF
--- a/lib/private/lists.bzl
+++ b/lib/private/lists.bzl
@@ -4,11 +4,11 @@ def every(f, arr):
     """Check if every item of `arr` passes function `f`.
 
     Example:
-      every(lambda i: i.endswith(".js"), ["app.js", "lib.js"]) // True
+      `every(lambda i: i.endswith(".js"), ["app.js", "lib.js"]) // True`
 
     Args:
-      f: function to execute on every item,
-      arr: list to iterate over,
+      f: Function to execute on every item
+      arr: List to iterate over
 
     Returns:
       True or False
@@ -19,11 +19,11 @@ def filter(f, arr):
     """Filter a list `arr` by applying a function `f` to each item.
 
     Example:
-      filter(lambda i: i.endswith(".js"), ["app.ts", "app.js", "lib.ts", "lib.js"]) // ["app.js", "lib.js"]
+      `filter(lambda i: i.endswith(".js"), ["app.ts", "app.js", "lib.ts", "lib.js"]) // ["app.js", "lib.js"]`
 
     Args:
-      f: function to execute on every item,
-      arr: list to iterate over,
+      f: Function to execute on every item
+      arr: List to iterate over
 
     Returns:
       A new list containing items that passed the filter function.
@@ -42,8 +42,8 @@ def find(f, arr):
     In this case `(-1, None)` is returned.
 
     Args:
-      f: function to execute on every item,
-      arr: list to iterate over,
+      f: Function to execute on every item
+      arr: List to iterate over
 
     Returns:
       Tuple (index, item)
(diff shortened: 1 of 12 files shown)


alexeagle (commented):
LGTM, will let @kormide look too

lib/private/lists.bzl
```diff
@@ -0,0 +1,116 @@
...
+    """
+    i = 0
+    for a in arr:
+        if f(a):
+            return (i, a)
+        i += 1
+    return (-1, None)
+
+def map(f, arr):
+    """Apply a function `f` with each item of `arr` and return a list with all items where said funtion returns truthy.
```
kormide, line 59: Does the function _need_ to return something truthy? Could it return `None`?

lib/private/lists.bzl
```diff
@@ -0,0 +1,116 @@
...
+def find(f, arr):
+    """Find a particular item from list `arr` by a given function `f`.
+
+    Unlike `pick`, the `find` method returns a tuple of the index and the value of first item passing by `f`.
+    Furhermore `find` does not fail if no item passes `f`.
+    In this case `(-1, None)` is returned.
+
+    Args:
+      f: function to execute on every item,
+      arr: list to iterate over,
```
kormide, line 46: For arg docs in bazel-lib, we tend to use the convention where we capitalize the first word and don't add a trailing punctuation. Would you be able to change all of these to match?

lib/private/lists.bzl
```diff
@@ -0,0 +1,116 @@
...
+    Returns:
+      True or False
+    """
+    return len(filter(f, arr)) == len(arr)
+
+def filter(f, arr):
+    """Filter a list `arr` by applying a function `f` to each item.
+
+    Example:
+      filter(lambda i: i.endswith(".js"), ["app.ts", "app.js", "lib.ts", "lib.js"]) // ["app.js", "lib.js"]
```
kormide, line 22: Can the examples be enclosed in backticks so they look like code in the markdown?

lib/private/lists.bzl
```diff
@@ -0,0 +1,116 @@
...
+    Returns:
+      True or False
+    """
+    return len(filter(f, arr)) == 1
+
+def pick(f, arr):
+    """Pick a particular item in list `arr` by a given function `f`.
+
+    Unlike `filter`, the `pick` method returns the first item _found_ by `f`.
+    If no item has passed `f`, the function will _fail_.
```
kormide, line 89: Unlike `find`?

lib/private/lists.bzl
```diff
@@ -0,0 +1,116 @@
...
+      some(lambda i: i.endswith(".js"), ["app.js", "lib.ts"]) // True
+
+    Args:
+      f: function to execute on every item,
+      arr: list to iterate over,
+
+    Returns:
+      True or False
+    """
+    return len(filter(f, arr)) > 0
```
kormide, line 116: Could this method short-circuit the search by returning as soon as the first item matches rather than running a full filter?

lib/private/lists.bzl
mgred: Oh, that's a mistake in the documentation :exploding_head: Obviously a `map` function should not care for the returned value, as e.g. `filter`. I'll clean that up :+1:

lib/private/lists.bzl
mgred: Sure :+1:

lib/private/lists.bzl
mgred: Yes, great hint :+1:

lib/private/lists.bzl
mgred: That's possible, I think I wanted to reuse the `filter` here without having the performance in mind at first place.

mgred:
@alexeagle thanks for taking the time to make some extra work

@kormide thanks for your review I really like your suggestions and will adopt them :heart:

kormide (approved):
LGTM

kormide:
Looks like this just needs a rebase to fix the test errors.

PR #514, opened 2023-09-06
fix: strip 'v' prefix from tag when setting module version

alexeagle:
### Type of change

### Test plan

This was exposed by the BCR PR


alexeagle:
cc @wyverald

kormide (approved):

PR #520, opened 2023-09-07
feat: support jq 1.7 toolchain

kormide:
Fixes https://github.com/aspect-build/bazel-lib/issues/268

### Type of change

- Jq 1.7 is now supported which has a wider selection of platform binaries
- BREAKING: jq stamp variables are now of the form `$stamp[0].MY_VARIABLE` rather than `$stamp.MY_VARIABLE`.
### Test plan

### DIFF
--- a/lib/jq.bzl
+++ b/lib/jq.bzl
@@ -104,7 +104,7 @@ def jq(name, srcs, filter = None, filter_file = None, args = [], out = None, **k
             # This 'as' syntax results in $stamp being null in unstamped builds.
             "$ARGS.named.STAMP as $stamp",
             # Provide a default using the "alternative operator" in case $stamp is null.
-            ".version = ($stamp.BUILD_EMBED_LABEL // \"<unstamped>\")",
+            ".version = ($stamp[0].BUILD_EMBED_LABEL // \"<unstamped>\")",
         ]),
     )
     ```
--- a/lib/private/jq.bzl
+++ b/lib/private/jq.bzl
@@ -57,14 +57,7 @@ def _jq_impl(ctx):
         )
         inputs.append(stamp_json)
 
-        # jq says of --argfile:
-        # > Do not use. Use --slurpfile instead.
-        # > (This option is like --slurpfile, but when the file has just one text,
-        # > then that is used, else an array of texts is used as in --slurpfile.)
-        # However there's no indication that it's deprecated. Maybe it's a style convention.
-        # For our purposes, "$STAMP.BUILD_TIMESTAMP" looks a lot more sensible in a BUILD file
-        # than "$STAMP[0].BUILD_TIMESTAMP".
-        args = args + ["--argfile", "STAMP", stamp_json.path]
+        args = args + ["--slurpfile", "STAMP", stamp_json.path]
 
     cmd = "{jq} {args} {filter} {sources} > {out}".format(
         jq = jq_bin.path,
--- a/lib/private/jq_toolchain.bzl
+++ b/lib/private/jq_toolchain.bzl
@@ -5,30 +5,31 @@ load(":repo_utils.bzl", "repo_utils")
 # Platform names follow the platform naming convention in @aspect_bazel_lib//:lib/private/repo_utils.bzl
 JQ_PLATFORMS = {
     "darwin_amd64": struct(
-        release_platform = "osx-amd64",
+        release_platform = "macos-amd64",
         compatible_with = [
             "@platforms//os:macos",
             "@platforms//cpu:x86_64",
         ],
     ),
     "darwin_arm64": struct(
-        # JQ only ships a universal binary; it should work on
-        # Apple Silicon (aarch64) as well.
-        release_platform = "osx-amd64",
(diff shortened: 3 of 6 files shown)


kormide:
There are some issues with the use of `--argfile` with the stamping feature which I need to work through. It appears that flag has been deprecated in 1.7.

lib/private/jq.bzl
```diff
@@ -57,14 +57,7 @@ def _jq_impl(ctx):
...
 
-        # jq says of --argfile:
-        # > Do not use. Use --slurpfile instead.
-        # > (This option is like --slurpfile, but when the file has just one text,
-        # > then that is used, else an array of texts is used as in --slurpfile.)
-        # However there's no indication that it's deprecated. Maybe it's a style convention.
-        # For our purposes, "$STAMP.BUILD_TIMESTAMP" looks a lot more sensible in a BUILD file
-        # than "$STAMP[0].BUILD_TIMESTAMP".
-        args = args + ["--argfile", "STAMP", stamp_json.path]
+        args = args + ["--slurpfile", "STAMP", stamp_json.path]
```
kormide, line 60: @alexeagle We have to use `--slurpfile` with 1.7. Should I do a version check on jq here to avoid introducing a breaking change? (the need to use `$stamp[0]` in the filter)

kormide: To not be breaking, I think we would also have to keep the default jq version at 1.6.

kormide: Then do a version check and use `--argfile` or `--slurpfile` accordingly, and document both types of stamp filters needed depending on the version.

alexeagle: Just a thought to unblock, what if we did make it breaking and go to 2.0

kormide: I like the idea of `2.0`. I'll have a look at what other breaking changes we can make.

alexeagle: ooh, you have a conference talk where you can "do the live release" on stage as an audience engagement tactic!

alexeagle: Screenshot from 2023-09-14 11-30-55

kormide: Rebased onto `2.x` so this can be a breaking change to how stamping works.

lib/private/jq_toolchain.bzl
```diff
@@ -3,50 +3,101 @@
...
+            ],
+        ),
+    },
 }
 
-DEFAULT_JQ_VERSION = "1.6"
+def jq_platforms_for_version(version):
+    (major, minor) = version.split(".")
+    pre_1_7 = int(major) < 1 or int(major) == 1 and int(minor) < 7
+    return _JQ_PLATFORMS["<1.7"] if pre_1_7 else _JQ_PLATFORMS[">=1.7"]
```
kormide, line 85: They only use `major.minor` hence why I'm not using a semver library.

lib/jq.bzl
```diff
@@ -104,7 +104,7 @@ def jq(name, srcs, filter = None, filter_file = None, args = [], out = None, **k
             # This 'as' syntax results in $stamp being null in unstamped builds.
             "$ARGS.named.STAMP as $stamp",
             # Provide a default using the "alternative operator" in case $stamp is null.
-            ".version = ($stamp.BUILD_EMBED_LABEL // \"<unstamped>\")",
+            ".version = ($stamp[0].BUILD_EMBED_LABEL // \"<unstamped>\")",
```
alexeagle, line 107: that's unfortunate. But now that expand_template supports stamp substitutions, I think we'll make less use of `jq` for this.

lib/private/jq_toolchain.bzl
```diff
@@ -3,50 +3,101 @@
...
-    ),
-    "windows_amd64": struct(
-        release_platform = "win64",
-        compatible_with = [
-            "@platforms//os:windows",
-            "@platforms//cpu:x86_64",
-        ],
-    ),
+_JQ_PLATFORMS = {
+    "<1.7": {
```
alexeagle, line 7: if it's a breaking change, why bother preserving compat with older versions of jq?

alexeagle (approved):

lib/private/jq_toolchain.bzl
kormide: I was assuming we should still support older versions of jq with newer versions of bazel-lib. Wdyt?

kormide: The breaking change only needs to be in stamping.

alexeagle: Unless it provides some utility to users, it seems to me that we should take the opportunity to simplify our code. Unlike the yq upgrade we attempted, I think pretty much everyone will be happy to have only 1.7 now?

kormide: I can't think of any utility it provides. I'll drop the old versions, and if a need becomes apparent then we can add back what I have here for pre-1.7 support.

alexeagle (approved):

PR #524, opened 2023-09-12
docs: Add --incompatible_disallow_empty_glob to correctness.bazelrc

thirtyseven:
This is a good flag to enable by default because it prevents a common mistake and usually is what users expect.

---
### Type of change

  - When updating their bazelrc to include this, some users may need to fix broken BUILD files or specify `allow_empty=True` when they were intentionally relying on the previous behavior.
### Test plan
Not sure this applies here.


gregmagolan:
Thanks for the suggestion @thirtyseven. I think this is a good one to add. @alexeagle wdyt?

alexeagle (approved):

PR #531, opened 2023-09-20
fix: lookup source files from execroot when running diff_test on RBE

mattem:
Allows `diff_test` to run against source files on RBE.

---

### Type of change

### Test plan

Run against RBE cluster via patch.


gregmagolan (approved):
🧙

PR #539, opened 2023-09-21
release: cut v1.34.6

thesayyn:
### Type of change


alexeagle:
Oh I didn't understand this change means we have to bumpversion with a commit at each release

thesayyn:
> Oh I didn't understand this change means we have to bumpversion with a commit at each release

I don’t like it either. I’m gonna make this be automated by GitHub actions

PR #540, opened 2023-09-21
fix: allow for BUILD files in outputs of write_source_file

### LINKED ISSUE #536: write_source_file can fail if the output_dir contains a BUILD file
https://github.com/aspect-build/bazel-lib/blob/45986f000d5564507b0bbaa3e4cb0f6c693d8fb3/lib/private/write_source_file.bzl#L374

Use case: a user creates a BUILD file to register a package with PNPM in rules_js. To do this, we need to generate a BUILD file into the generated NPM package. We are trying to use `write_source_file` to copy the output of the generated package into the source tree. This worked in our old world because our internal rule doesn't check if the folder exists with a `glob`. It seems that checking with a `glob` will fail because you can't `glob` into a package under Bazel. Generating a BUILD file creates a package and therefore this check fails.

mattem:
Handles the case in `write_source_file` where the generated output directory contains `BUILD` files, which can not be globbed through, resulting in incorrect failure states for missing output files.

Closes #536

---

### Type of change

Handle `BUILD` files in the output of `write_source_file`

### Test plan

- New test cases added

### DIFF
--- a/lib/private/write_source_file.bzl
+++ b/lib/private/write_source_file.bzl
@@ -372,4 +372,11 @@ def _is_file_missing(label):
     file_abs = "%s/%s" % (label.package, label.name)
     file_rel = file_abs[len(native.package_name()) + 1:]
     file_glob = native.glob([file_rel], exclude_directories = 0, allow_empty = True)
-    return len(file_glob) == 0
+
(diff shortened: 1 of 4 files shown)


kormide (approved):
Nice :+1: . I didn't think to use subpackages.

PR #542, opened 2023-09-22
refactor: cleanup e2e

thesayyn:
### Type of change

### Test plan

Now all of our e2e tests are tested both with bzlmod and without it. Also, removes aspect-recommended bazel flags to test against the default behavior of bazel across versions.


alexeagle:
hm, I think @gregmagolan had a reason to install the recommended flags into all of these e2e ?

thesayyn:
> hm, I think @gregmagolan had a reason to install the recommended flags into all of these e2e ?

I am the impostor https://github.com/aspect-build/bazel-lib/commits/main/e2e/smoke/.aspect

jbedard (approved):

PR #543, opened 2023-09-22
refactor: consume tools from source if unstamped

thesayyn:
### Type of change

### Test plan


.github/workflows/release_prep.sh
```diff
@@ -7,6 +7,10 @@ set -o errexit -o nounset -o pipefail
 cat >.git/info/attributes <<EOF
 # Omit folders that users don't need, making the distribution artifact smaller
 lib/tests export-ignore
+e2e export-ignore
```
alexeagle, line 10: I think this will break the BCR publishing since e2e/smoke is used.

https://github.com/aspect-build/rules_js/blob/main/.github/workflows/release_prep.sh#L10

thesayyn: oh good catch. will remove now.

.github/workflows/release.yml
```diff
@@ -27,11 +27,6 @@ jobs:
           # Bazelisk will download bazel to here
           XDG_CACHE_HOME: ~/.cache/bazel-repo
         run: bazel --bazelrc=.github/workflows/ci.bazelrc --bazelrc=.bazelrc test //...
-      - name: Check Versions
-        env:
-          # Bazelisk will download bazel to here
-          XDG_CACHE_HOME: ~/.cache/bazel-repo
-        run: ./scripts/check_versions.sh ${{ env.GITHUB_REF_NAME }}
```
alexeagle, line 34: I guess we always build with `-c opt` so the existing bazel test on the hashes is good enough? I thought you'd have to add another `bazel` invocation here to check them

.pre-commit-config.yaml
```diff
@@ -38,11 +38,3 @@ repos:
         exclude: &exclude_pattern "^(docs|e2e|lib/tests)/"
       - id: mixed-line-ending
       - id: trailing-whitespace
-
-  - repo: local
```
alexeagle, line 42: revert the `default_install_hook_types` above too?

MODULE.bazel
```diff
@@ -25,6 +25,18 @@ register_toolchains(
     "@expand_template_toolchains//:all",
 )
 
+# To allow /tools to be built from source
+bazel_dep(name = "rules_go", version = "0.41.0", repo_name = "io_bazel_rules_go")
```
alexeagle, line 29: This isn't good to add them as non-dev deps.

MODULE.bazel
```diff
@@ -25,6 +25,18 @@ register_toolchains(
...
+bazel_dep(name = "gazelle", version = "0.33.0")
```
jbedard, line 30: Isn't this a transitive dep (of rules_go) and we don't need to declare it with bzlmod?

MODULE.bazel
thesayyn: yeah, I am not sure if bzlmod allows users to declare dependencies for aspect_bazel_lib. they are dependencies but only if you consume from source.

tools/sri.bzl
```diff
@@ -0,0 +1,26 @@
+"AUTO GENERATED. DO NOT EDIT"
```
alexeagle, line 1: nit: name it "integrity.bzl" maybe? The term "sub-resource integrity" is probably not familiar to most readers and doesn't appear anywhere in the code

tools/version.bzl
```diff
@@ -0,0 +1,5 @@
+"version information. replaced with stamped info with each release"
```
alexeagle, line 1: nit: mention git export-subst and/or point to release_prep.sh so someone can learn how the stamping occurs

tools/sri.bzl
thesayyn: agreed.

tools/version.bzl
```diff
@@ -0,0 +1,5 @@
...
+
+_VERSION_PRIVATE = "$Format:%(describe:tags=true)$"
```
jbedard, line 3: Add a basic comment that this is git magic?

tools/BUILD.bazel
```diff
@@ -83,8 +54,10 @@ release(
...
 )
 
-bzl_library(
-    name = "hashes",
-    srcs = ["hashes.bzl"],
-    visibility = ["//visibility:public"],
+write_source_files(
+    name = "releases_versions_check_in",
+    files = {
+        "sri.bzl": ":release_versions",
```
alexeagle, line 60: this is where I'm concerned we dropped `--compilation_mode=opt` or whatever else is in `--config=release`

thesayyn: i should add a transition here.

.github/workflows/release.yml
thesayyn: I'll add a transition instead.

thesayyn: actually, now that I think about it, `--config=release` might have more flags than `-c opt`. Will add `--config=release` to the existing test step.

tools/BUILD.bazel
thesayyn: done.

.github/workflows/release_prep.sh
```diff
@@ -7,6 +7,9 @@ set -o errexit -o nounset -o pipefail
 cat >.git/info/attributes <<EOF
 # Omit folders that users don't need, making the distribution artifact smaller
 lib/tests export-ignore
+
+# Substitution for the _VERSION_PRIVATE placeholder at the top of this file
```
alexeagle, line 11: nit: not "this file"

MODULE.bazel
alexeagle: I think we have to make this a devDep though. rules_go should be able to depend on bazel-lib so we shouldn't depend on rules_go.
It means bzlmod users with a `git_override` will need to specify a toolchain though.

I guess we can do something crazy to make it a devdep only when publishing, but crazy things also not so good

e2e/copy_to_directory/WORKSPACE
```diff
@@ -6,3 +6,24 @@ local_repository(
 load("@aspect_bazel_lib//lib:repositories.bzl", "aspect_bazel_lib_dependencies")
 
 aspect_bazel_lib_dependencies()
+
+############################################
+# rules_go is needed to consume tools from sources
+
+load("@aspect_bazel_lib//:internal_deps.bzl", "bazel_lib_internal_deps")
```
alexeagle, line 13: nit: maybe worth having a internal_deps.bzl go macro users could load to get these?

alexeagle (approved):

e2e/copy_to_directory/WORKSPACE
thesayyn: I think I can do that as a follow-up if anybody complains.

MODULE.bazel
thesayyn: I just made it so that dev-dependencies are dependencies when running our e2e tests. users can accomplish the same thing by patching it using single_version override

thesayyn:
cross-ref: https://github.com/bazel-contrib/publish-to-bcr/issues/54

PR #545, opened 2023-09-22
refactor: use list utils

mgred:
from https://github.com/aspect-build/bazel-lib/issues/495#issuecomment-1675960410:

> We should figure out a place to "dogfood" this so we have the experience of being a user of the new API. Maybe we can refactor other code in bazel-lib to use these.

Both refactorings use lambda expressions. These are available from [bazel 5.0](https://blog.bazel.build/2022/01/19/bazel-5.0.html#highlights). Is there a minimum version of bazel that bazel-lib needs to support?

---

### Type of change

### Test plan

### DIFF
--- a/lib/private/copy_to_directory.bzl
+++ b/lib/private/copy_to_directory.bzl
@@ -6,6 +6,7 @@ load(":paths.bzl", "paths")
 load(":directory_path.bzl", "DirectoryPathInfo")
 load(":glob_match.bzl", "glob_match", "is_glob")
 load(":platform_utils.bzl", _platform_utils = "platform_utils")
+load(":lists.bzl", "filter", "map")
 
 _filter_transforms_order_docstring = """Filters and transformations are applied in the following order:
 
@@ -637,6 +638,24 @@ def _expand_src_packages_patterns(patterns, package):
             result.append(pattern)
     return result
 
+def _create_struct_from_file(f):
+    return struct(
+        file = f,
+        path = f.path,
+        root_path = f.root.path,
+        short_path = f.short_path,
+        workspace_path = paths.to_workspace_path(f),
+    )
+
+def _create_struct_from_target(t):
+    return struct(
+        file = t[DirectoryPathInfo].directory,
+        path = "/".join([t[DirectoryPathInfo].directory.path, t[DirectoryPathInfo].path]),
+        root_path = t[DirectoryPathInfo].directory.root.path,
+        short_path = "/".join([t[DirectoryPathInfo].directory.short_path, t[DirectoryPathInfo].path]),
+        workspace_path = "/".join([paths.to_workspace_path(t[DirectoryPathInfo].directory), t[DirectoryPathInfo].path]),
+    )
+
 def copy_to_directory_bin_action(
         ctx,
         name,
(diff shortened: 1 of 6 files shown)


alexeagle:
We support whatever we test against: https://github.com/aspect-build/bazel-lib/blob/main/.github/workflows/ci.yaml#L37-L50

so yes, Bazel 5 is the minimum

alexeagle:
What's needed to promote this from Draft?

mgred:
> What's needed to promote this from Draft?

I just found these two places for now. Maybe there are more where we can use the list utils?

mgred:
If this is enough for now, I'm eager to make this _ready for review_ :sunglasses:

lib/private/copy_to_directory.bzl
```diff
@@ -758,24 +777,8 @@ def copy_to_directory_bin_action(
...
-            continue
-        files_and_targets.append(struct(
-            file = t[DirectoryPathInfo].directory,
-            path = "/".join([t[DirectoryPathInfo].directory.path, t[DirectoryPathInfo].path]),
-            root_path = t[DirectoryPathInfo].directory.root.path,
-            short_path = "/".join([t[DirectoryPathInfo].directory.short_path, t[DirectoryPathInfo].path]),
-            workspace_path = "/".join([paths.to_workspace_path(t[DirectoryPathInfo].directory), t[DirectoryPathInfo].path]),
-        ))
+    files_and_targets.extend(map(_create_struct_from_file, files))
+    files_and_targets.extend(map(_create_struct_from_target, filter(lambda t: DirectoryPathInfo in t, targets)))
```
mgred, line 781: I'm not sure if the `extend` function is the proper choice here. Could the `+=` operator be a better fit to concat the lists?

mgred: ok, `extend` should be the proper way to do it, since this is used throughout the code base and lots of rules. It's recommended over the operator. I'm wondering why the `+=` is good for strings but seems a _not so good_ choice for lists. Seems to be about the implementation of the interpreter that bazel is using.

mgred:
@alexeagle I found several places where some of  the list utils can be used. Could you please check, if this is going in the right direction?

Thanks :v:

lib/private/yq.bzl
```diff
@@ -18,13 +19,13 @@ _yq_attrs = dict({
...
-        if arg.startswith("-s") or arg.startswith("--split-exp"):
-            return True
-    return False
+    return some(lambda arg: arg.startswith("-s") or arg.startswith("--split-exp"), args)
+
+def _map_to(value, arr):
+    return map(lambda _: value, arr)
 
 def _escape_path(path):
-    return "/".join([".." for t in path.split("/")]) + "/"
```
alexeagle, line 27: this one seems less readable to me than before

lib/private/write_source_file.bzl
```diff
@@ -177,9 +178,7 @@ def _write_source_file_sh(ctx, paths):
         ctx.label.name + "_update.sh",
     )
 
-    additional_update_scripts = []
-    for target in ctx.attr.additional_update_targets:
-        additional_update_scripts.append(target[WriteSourceFileInfo].executable)
+    additional_update_scripts = map(lambda t: t[WriteSourceFileInfo].executable, ctx.attr.additional_update_targets)
```
alexeagle, line 181: sadness, I think we should have put the map arguments in the other order, but we shipped the lib already so that would be a breaking change for 2.0 and not worth it

mattem:
Given the wide use of bazel-lib in different areas of the build, has consideration been given to the extra memory pressure these lambda functions introduce given the captured closures, vs the syntax sugar here. I'd be interested in seeing the analysis of useage against a bigger build.

lib/private/write_source_file.bzl
mgred: Can you explain why you think the arguments should be in changed order? One argument that comes to my mind is, that it might be a bit confusing to have a too long lambda expression as the first argument and a short worded list as the second. That can theoretically bypassed by assigning the lambda to a meaningfully named variable.

I think it's kind of important to have the arguments across all function in the same order, for convenience. The intention was to have an order that makes sense to say _naturally_, like: `map` method `foo` on list `arr`.

lib/private/yq.bzl
mgred: alright, I see your point... the only thing i would change is to _blank_ the loop variable with `_`, because it's not needed anyways, ok?

alexeagle: oh this should have been `[".." * len(path.split("/"))]` I think. was never a need to loop

mgred: ahh, that looks pretty. I can change it accordingly :+1:

mgred:
> has consideration been given to the extra memory pressure these lambda functions introduce given the captured closures

No, I was not aware of this. Is there a significant performance drop when using lambda expressions?

alexeagle:
You can google "functional closures increase memory usage" to see explanations for other languages like JavaScript, PHP, etc. Every time a lambda is created, it needs to store variables that it can reference. So we shouldn't create them in tight-loop code where we might create a million of them.

mgred:
@alexeagle as far as I understand the topic, we shouldn't have a problem with this, since the lambda is _created_ outside the loop and only _called_ inside it... or do I see it wrong?

alexeagle:
The issue is when one of these new functions is called within some enclosing loop.

alexeagle:
I think we are just scared to have syntax sugar where it might impact performance or memory usage in such a low-level library, unless someone has time and motivation to understand it more

PR #547, opened 2023-09-24
yq: Fix path resolution for external targets

phlax:
Currently the `yq` target is unusable if called from an external build, this should resolve that

---

### Type of change

- Style (white-space, formatting, etc...)

### Test plan

- New test cases added

### DIFF
--- a/lib/private/yq.bzl
+++ b/lib/private/yq.bzl
@@ -71,10 +71,13 @@ def _yq_impl(ctx):
 
     # For split operations, yq outputs files in the same directory so we
     # must cd to the correct output dir before executing it
-    bin_dir = "/".join([ctx.bin_dir.path, ctx.label.package]) if ctx.label.package else ctx.bin_dir.path
+    bin_dir = ctx.bin_dir.path
+    if ctx.label.workspace_name:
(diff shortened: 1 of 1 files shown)


lib/private/yq.bzl
```diff
@@ -71,10 +71,13 @@ def _yq_impl(ctx):
 
     # For split operations, yq outputs files in the same directory so we
     # must cd to the correct output dir before executing it
-    bin_dir = "/".join([ctx.bin_dir.path, ctx.label.package]) if ctx.label.package else ctx.bin_dir.path
+    bin_dir = ctx.bin_dir.path
+    if ctx.label.workspace_name:
+        bin_dir = "%s/external/%s" % (bin_dir, ctx.label.workspace_name)
```
alexeagle, line 76: this doesn't work with `--nolegacy_external_runfiles`.

alexeagle:
It might be a good idea to file an issue for this first, in case others want to contribute a fix for it.

phlax:
> It might be a good idea to file an issue for this first, in case others want to contribute a fix for it.

i proposed a fix as this is blocking us from being able to do some external builds - so was hoping for a quick turnaround and needed something in the interim that works

#548

PR #553, opened 2023-09-26
feat: add `propagate_common_{,test_,binary_}rule_attributes` to lib/utils.

reltuk:
### Type of change

- ✅ Relevant documentation has been updated

Added helper functions `propagate_common_rule_attributes`, `propagate_common_test_rule_attributes` and `propagate_common_binary_rule_attributes` which can be used to forward common attributes to rule invocations when writing macros.

### Test plan

- ✅ New test cases added
- Manual testing in a context where I hope to use this soon (https://github.com/bazel-contrib/rules_oci/pull/370)


lib/private/utils.bzl
```diff
@@ -247,6 +247,97 @@ def _maybe_http_archive(**kwargs):
...
+    https://bazel.build/reference/be/common-definitions#common-attributes-binary
+
+    Args:
+        attrs: Dict of parameters to filter
+
+    Returns:
+        The dict of parameters, containing only common binary attributes
+    """
+
+    COMMON_BINARY_RULE_ATTRIBUTES = [
```
thesayyn, line 329: can you extract these constants to global with an underscore?

lib/private/utils.bzl
```diff
@@ -247,6 +247,97 @@ def _maybe_http_archive(**kwargs):
...
+    https://bazel.build/reference/be/common-definitions#common-attributes-tests
+
+    Args:
+        attrs: Dict of parameters to filter
+
+    Returns:
+        The dict of parameters, containing only common test attributes
+    """
+
+    COMMON_TEST_RULE_ATTRIBUTES = [
```
thesayyn, line 298: can you extract these constants to global with an underscore?

thesayyn (approved):

lib/private/utils.bzl
```diff
@@ -247,6 +247,97 @@ def _maybe_http_archive(**kwargs):
...
 
+_COMMON_RULE_ATTRIBUTES = [
+    "compatible_with",
+    "deprecation",
+    "distribs",
+    "exec_compatible_with",
+    "exec_properties",
+    "features",
+    "restricted_to",
+    "tags",
```
kormide, line 258: To what extent is `tags` already solved via https://bazel.build/reference/command-line-reference#flag--incompatible_allow_tags_propagation? And are there any interactions if the flag is enabled?

thesayyn: no, it doesn't interact with that flag directly, however, it will allow tags to be forwarded to targets expanded by a macro. It's merely a helper function that will pick well-known attributes from `**kwargs` so we can safely add them to other targets in a macro.

lib/private/utils.bzl
```diff
@@ -247,6 +247,97 @@ def _maybe_http_archive(**kwargs):
...
+        The dict of parameters, containing only common attributes
+    """
+
+    return {
+        k: attrs[k]
+        for k in attrs
+        if k in _COMMON_RULE_ATTRIBUTES
+    }
+
+def _propagate_common_test_rule_attributes(attrs = {}):
```
thesayyn, line 301: one more change; let's remote the default for the argument for all these new functions.

thesayyn: then it's good to go if @kormide is okay with it.

lib/private/utils.bzl
reltuk: I'm not familiar with `--incompatible_allow_tags_propagation`. I tried to dig into it a little bit.

It looks like its purpose is to propagate certain tags on rule *and macro* invocations into to the exec properties of the spawned actions which the rule implementations create.

I'm not familiar enough with Bazel internals to understand exactly how the macros get expanded into rule invocations and where the kwargs interpolation of things like tags would happen. I think, based on my understanding, that copying the tags which were globbed into the kwargs of the macro invocation into each rule or macro sub-invocation, for example, as per the intended usage of `propagate_common_*rule_attributes` would have the same final effect, but I can't be certain of that.

I would also say it's also possible this implementation should be a little different for `tags` in particular. For example, there is the implementation of `propagate_well_known_tags` up above, and maybe propagating `tags` through this interface should automatically filter them for well-known tags or something like that.

reltuk:
Thanks for the great feedback and support here so far everyone :).

kormide (approved):

PR #554, opened 2023-09-27
chore: rules_go is a non-dev dep at HEAD

alexeagle:
This ensures if a bzlmod user were to fetch bazel-lib with a SHA or a `git_repository`, they don't need to patch our repo (similar to how CI was doing it) in order to build the tools/ from head

### DIFF
--- a/.bcr/patches/go_dev_dep.patch
+++ b/.bcr/patches/go_dev_dep.patch
@@ -1,0 +1,16 @@
+diff --git a/MODULE.bazel b/MODULE.bazel
+index f05c57a..4cb6104 100644
+--- a/MODULE.bazel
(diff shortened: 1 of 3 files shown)


thesayyn (approved):

kormide:
Would be nice if we could generate the patch and use a `write_source_files` to keep it updated. Otherwise we'll only know it failed when the BCR release fails. If that proves difficult to do, perhaps it's a good motivation for publish-to-bcr to support a regex replace like you originally suggested :thinking:

alexeagle:
I'll just add a CI step that runs `patch -p1 < .bcr/patches/*` to catch that on CI without adding complexity

PR #556, opened 2023-09-27
ci: add assertion that bcr patches apply

alexeagle:


.github/workflows/ci.yaml
```diff
@@ -218,3 +218,7 @@ jobs:
         if: ${{ matrix.folder == '.' && matrix.os != 'windows-latest' }}
         # Find all shell scripts within e2e, echo the filename, execute, fail on error
         run: find e2e/*.sh -maxdepth 1 -type f -exec sh -c 'echo "\n\n------------------------------- $0 -------------------------------" && "$0" || kill $PPID' \{\} \;
+
+      - name: verify bcr patches
+        if: matrix.bzlmodEnabled
+        run: patch -p1 < .bcr/patches/*.patch
```
kormide, line 224: Perhaps we could use `--dry-run` here? Just in case the order of this ever gets switched.

PR #560, opened 2023-09-29
feat: add assert_directory_contains test to testing.bzl

gregmagolan:
Similar to `assert_archive_contains` but for a directory artifact. If you're testing a directory output of a target this is better than use using a `build_test` since you can also assert that expected files are outputted.

---

### Type of change

### Test plan

- New test cases added

Manually tested the negative case.


lib/testing.bzl
```diff
@@ -193,3 +193,58 @@ def assert_archive_contains(name, archive, expected, type = None, **kwargs):
         timeout = "short",
         **kwargs
     )
+
+def assert_directory_contains(name, directory, expected, **kwargs):
```
thesayyn, line 197: nit: use https://github.com/aspect-build/bazel-lib/blob/e8bbdbfda60dea64f820d89ff98b6a0c6e379f88/lib/utils.bzl#L15 to propagate well know attributes to expanded targets.

thesayyn (approved):

lib/testing.bzl
gregmagolan: vs `**kwargs` you mean? Pattern here with `**kwargs` is the same as all of the `assert_*` function above so would be better to refactor them all in a follow-up with `propagate_common_test_rule_attributes` instead.

PR #561, opened 2023-09-29
feat: add undeclared output mode to expand_template

gregmagolan:
This is useful pattern in cases where you want to do a `copy_to_bin` because the output file must be in the same location and have the same name as the source file but want to do substitutions on the file that you are copying to bin.

---

### Type of change

### Test plan

- New test cases added


kormide (approved):

alexeagle (approved):
YES! This is delightful, I think I've needed this before for tsconfig.json hacks

PR #566, opened 2023-10-01
feat: implement linux bsdtar toolchain

thesayyn:


lib/private/tar_toolchain.bzl
```diff
@@ -2,80 +2,169 @@
...
+        compatible_with = [
+            "@platforms//os:windows",
+            "@platforms//cpu:x86_64",
+        ],
+    ),
+}
+
+LIBARCHIVE_TOOLS_URLS = {
+    "linux_amd64": struct(
+        url = "https://launchpadlibrarian.net/595393330/libarchive-tools_3.6.0-1ubuntu1_amd64.deb",
```
alexeagle, line 34: hm, I wonder how reliable that site is, maybe we'll need mirrors

alexeagle (approved):

lib/private/tar_toolchain.bzl
thesayyn: i have no idea but it's ubuntu, https://launchpad.net/ubuntu/jammy/amd64/libarchive-tools/3.6.0-1ubuntu1#:~:text=libarchive%2Dtools_3.6.0%2D1ubuntu1_amd64.deb

PR #568, opened 2023-10-01
fix(ci): include libarchive13.so when running tar

alexeagle:


lib/private/tar.bzl
```diff
@@ -89,11 +94,12 @@ def _tar_impl(ctx):
     inputs.append(ctx.file.mtree)
 
     ctx.actions.run(
-        executable = ctx.toolchains["@aspect_bazel_lib//lib:tar_toolchain_type"].tarinfo.binary,
+        executable = tarinfo.binary,
         inputs = inputs,
         outputs = [out],
         arguments = [args],
         mnemonic = "Tar",
+        env = {"LD_LIBRARY_PATH": tarinfo.include_path} if tarinfo.include_path else {},
```
thesayyn, line 102: I think this should not leak into actions. maybe create a bash wrapper in the repo rule that sets the env vars then executes the binary? I would feel much more comfortable with that as that'd be unified behavior across our toolchains. libs can go into defaultinfo from the toolchain.

thesayyn: this would simply the genrule case as well. let's make ld libraries a toolchain implementation detail.

alexeagle: Yes, making a wrapper so the binary always sets the LD path sounds good. However that won't fix whatever is red about this right now...

PR #572, opened 2023-10-03
Update renovate.json

alexeagle:
Don't try to update the root MODULE.bazel, it has our user-facing minimum versions

Synced from bazel-contrib/rules-template


gregmagolan (approved):
:shipit:

PR #573, opened 2023-10-03
chore: drop GHA CI for linux

alexeagle:
We are using Aspect Workflows so it's redundant


alexeagle:
Doesn't really reduce our CI load that much for the added complexity

PR #576, opened 2023-10-05
Avoid expanding mtree spec during analysis phase

dzbarsky:
Seeing what CI thinks, I didn't see why we need the empty line at the end of the file

### Type of change

- Bug fix (change which fixes an issue)
- New feature or functionality (change which adds functionality)
- Style (white-space, formatting, etc...)
- Refactor (a code change that neither fixes a bug or adds a new feature)
**- Performance (a code change that improves performance)**
- Documentation (updates to documentation or READMEs)
- Chore (any other change that doesn't affect source or test files, such as configuration)

**For changes visible to end-users**

- Breaking change (this change will force users to change their own code or config)
- Relevant documentation has been updated
- Suggested release notes are provided below:

### Test plan

**- Covered by existing test cases**
- New test cases added
- Manual testing; please provide instructions so we can reproduce:

PR #581, opened 2023-10-05
chore: run CI on 2.x branch

kormide:
### Type of change


alexeagle (approved):
thanks! #576 @dzbarsky noticed that too

PR #582, opened 2023-10-05
chore: remove legacy copy_to_directory_action

kormide:
### Type of change

- The legacy `copy_to_directory_action` rule helper has been removed. Please use `copy_to_directory_bin_action` action instead.

### Test plan


kormide:
Sorry @gregmagolan, I should have given you the satisfaction of deleting all of this.

lib/private/BUILD.bazel
```diff
@@ -12,3 +14,299 @@ exports_files(
...
+    visibility = ["//lib:__subpackages__"],
+    deps = [":strings"],
+)
+
+bzl_library(
+    name = "copy_common",
+    srcs = ["copy_common.bzl"],
+    visibility = ["//lib:__subpackages__"],
+)
+
```
kormide, line 30: I ran gazelle which made a large diff. Perhaps I should have not for this PR.

alexeagle: Yeah gazelle is a bit broken in this repo. If we liked it then we woulda put a CI on it. Likely have to revert some of the incorrect changes it makes.

kormide: Removed the gazelle diffs.

alexeagle (approved):

PR #584, opened 2023-10-05
chore: run gazelle

alexeagle:
Also delete the local_config_platform workaround for bazel 5 which made our docs setup complex

### Type of change

bazel-lib no longer provides a workaround for Bazel 5 users to get a `@local_config_platform` repository.
If you need this, you must upgrade to Bazel 6.

### Test plan


kormide (approved):

alexeagle:
I think we should drop support for Bazel 5 altogether.

PR #591, opened 2023-10-06
refactor: remove exclude_prefixes from copy_to_directory and copy_to_directory_bin_action

gregmagolan:
Removes deprecated `exclude_prefixes` from `copy_to_directory` and `copy_to_directory_bin_action`. Users should use `exclude_srcs_patterns` with glob patterns instead.

BREAKING CHANGES

- `exclude_prefixes` attribute is removed from `copy_to_directory`. Users should use `exclude_srcs_patterns` with glob patterns instead.
- `exclude_prefixes` attribute is removed from `copy_to_directory_bin_action`. Users should use `exclude_srcs_patterns` with glob patterns instead.


alexeagle (approved):

lib/private/copy_to_directory.bzl
```diff
@@ -164,23 +164,6 @@ Files that do not have matching output directory paths are subject to subsequent
 filters and transformations to determine if they are copied and what their path in the output
 directory will be.
 
-Globs are supported (see rule docstring above).
```
kormide, line 167: Is this line part of the previous comment block?

gregmagolan: Just an artifact of how github decided to show the diff. The line is repeated at the end of the diff block.

Issue #593, opened 2023-10-06
2.0: Refactor toolchain registration


alexeagle:
Registering toolchains should be a separate call made by users, not tucked into `aspect_bazel_lib_dependencies`.
Under bzlmod, this should also allow registering your own toolchain rather than ours always taking priority, maybe with a second module extension.

Issue #592, opened 2023-10-06
[FR]: instructions to register jq toolchain in MODULE.bazel


pdeva:
### What is the current behavior?

the release page only shows examples of registering jq toolchain with workspace file. 

### Describe the feature

currently no such instructions exist

alexeagle:
There are no such instructions because there's nothing to do, our MODULE.bazel file already registers jq_toolchain for you.

See #597 for some related refactoring, but this will still be true.

We should probably add a sentence in the install section there about the toolchains being registered by default.

PR #597, opened 2023-10-08
refactor: rework toolchain registration for WORKSPACE and bzlmod users

kormide:
Stop automatically registering all of the toolchains in bazel-lib for WORKSPACE users. Make it easier to override toolchain versions under bzlmod.

Closes https://github.com/aspect-build/bazel-lib/issues/593.

### Type of change

* BREAKING: Toolchains are no longer automatically registered for WORKSPACE users. Users must call the `aspect_bazel_lib_register_toolchains` macro or call the `register_xyz_toolchain` macros individually. See [Release](https://github.com/aspect-build/bazel-lib/releases) for snippets.
* BREAKING: The host repository has been moved from the extension `ext` into a differed module extension named `host`.

### Test plan

- New test cases added

### DIFF
--- a/.github/workflows/release_prep.sh
+++ b/.github/workflows/release_prep.sh
@@ -48,25 +48,16 @@ http_archive(
     url = "https://github.com/aspect-build/bazel-lib/releases/download/${TAG}/${ARCHIVE}",
 )
 
-load("@aspect_bazel_lib//lib:repositories.bzl", "aspect_bazel_lib_dependencies")
+load("@aspect_bazel_lib//lib:repositories.bzl", "aspect_bazel_lib_dependencies", "aspect_bazel_lib_register_toolchains")
+
+# Required bazel-lib dependencies
 
 aspect_bazel_lib_dependencies()
 
+# Register bazel-lib toolchains
+
+aspect_bazel_lib_register_toolchains()
+
 \`\`\`
 
-Optional toolchains:
-
-\`\`\`starlark
-# Register the following toolchain to use jq
-
-load("@aspect_bazel_lib//lib:repositories.bzl", "register_jq_toolchains")
-
-register_jq_toolchains()
-
-# Register the following toolchain to use yq
-
-load("@aspect_bazel_lib//lib:repositories.bzl", "register_yq_toolchains")
-
-register_yq_toolchains()
-\`\`\`
 EOF
--- a/docs/repositories.md
+++ b/docs/repositories.md
@@ -14,6 +14,21 @@ Load dependencies required by aspect rules
 
 
 
+<a id="aspect_bazel_lib_register_toolchains"></a>
+
+## aspect_bazel_lib_register_toolchains
+
+<pre>
+aspect_bazel_lib_register_toolchains()
+</pre>
+
+Register all bazel-lib toolchains at their default versions.
+
+To be more selective about which toolchains and versions to register,
+call the individual toolchain registration macros.
+
+
+
 <a id="register_copy_directory_toolchains"></a>
 
 ## register_copy_directory_toolchains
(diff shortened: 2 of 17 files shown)


kormide:
This is still a WIP, but wouldn't mind some preliminary feedback. Documentation still needs to be updated.

lib/private/extension_utils.bzl
```diff
@@ -0,0 +1,69 @@
...
+
+def _toolchain_repos(mctx, get_tag, toolchain_name, repos_fn, default_repository = None, get_name = None, get_version = None):
+    """Create toolchain repositories from bzlmod extension tags.
+
+    For uses of the tag class where the name isn't overridden (the `default_repository` name
+    is used), creates a single toolchain repository for the invocation of the tag closest to
+    the root module. Otherwise, resolution follows bzlmod's depth-first search through the
+    module graph.
+
+    Only a root module may declare a toolchain with a name other than the default repository name.
```
kormide, line 11: This function is confusing so I'm going to do some more iteration on these docs.

lib/private/extension_utils.bzl
```diff
@@ -0,0 +1,104 @@
+"""Utility functions for bzlmod extensions"""
+
+def _toolchain_repos_bfs(mctx, get_tag_fn, toolchain_name, toolchain_repos_fn, default_repository = None, get_name_fn = None, get_version_fn = None):
```
kormide, line 3: Calling this one "bfs" because I can envision another strategy to resolve toolchains based on semver.

lib/private/extension_utils.bzl
```diff
@@ -0,0 +1,104 @@
...
+                    registrations[name],
+                ))
+            else:
+                registrations[name] = version
+
+        for name, version in registrations.items():
+            toolchain_repos_fn(
+                name = name,
+                version = version,
+            )
```
kormide, line 100: I'm wondering if this macro is doing too much and we should return the info for which toolchain name/versions to create and not actually create them.

alexeagle (commented):
I don't think we need to completely stop registering toolchains. Our goal is just to make sure that users don't download unnecessary binaries, and are able to override our toolchain with whatever one they choose, right?

https://github.com/bazelbuild/rules_go/blob/master/docs/go/core/bzlmod.md#registering-go-sdks suggests we should be able to see how they avoid registration taking precedence over users.

.github/workflows/release_prep.sh
```diff
@@ -48,10 +73,24 @@ http_archive(
     url = "https://github.com/aspect-build/bazel-lib/releases/download/${TAG}/${ARCHIVE}",
 )
 
-load("@aspect_bazel_lib//lib:repositories.bzl", "aspect_bazel_lib_dependencies")
+load("//lib:repositories.bzl", "aspect_bazel_lib_dependencies", "register_copy_directory_toolchains", "register_copy_to_directory_toolchains", "register_coreutils_toolchains", "register_expand_template_toolchains", "register_tar_toolchains")
```
alexeagle, line 76: in the snippet we should solve for #592 by mentioning which ones you get by default and **don't** have to register.

kormide: It there a reason why we shouldn't just automatically register all of them? They should only be downloaded if used by the corresponding rules, right?

kormide:
Updated. Under bzlmod, continue autoregistering toolchains as this does not interfere with the ability for someone to override a toolchain in the root module (see https://bazelbuild.slack.com/archives/C014RARENH0/p1696893626545609). Toolchains can be selectively overridden at different versions.

For WORKSPACE users, added an `aspect_bazel_lib_register_toolchains` macro which can be used to perform all of the registration. Users can optionally register them selectively with the individual toolchain macros.

Made `extension_utils` private while we work out its utility.

docs/repositories.md
```diff
@@ -14,6 +14,21 @@ Load dependencies required by aspect rules
...
 
+<a id="aspect_bazel_lib_register_toolchains"></a>
+
+## aspect_bazel_lib_register_toolchains
+
+<pre>
+aspect_bazel_lib_register_toolchains()
+</pre>
+
+Register all bazel-lib at their default versions.
```
alexeagle, line 25: all bazel-lib **toolchains**

alexeagle (approved):

PR #599, opened 2023-10-08
chore: fix tar_toolchain naming

alexeagle:
Also register them in the correct order, matching how WORKSPACE works


kormide (approved):

alexeagle:
I should add, it's surprising that no testing caught this, especially on BCR. I guess the mechanism of overriding the dependency must not have the same logic.

Issue #466, opened 2023-06-28
[Bug]: copy_to_directory fails with host!=exec RBE


anguslees:
### What happened?

I'm using bazel on a darwin/arm64 host, pointing at a linux/amd64 RBE cluster.  `copy_to_directory` fails with "cannot execute binary file" because bazel is trying to run a linux/amd64 binary locally.

Here is one example failure, where `copy_to_directory` is invoked as part of `rules_oci`'s `oci_pull`
```
ERROR: /private/var/tmp/_bazel_gus/798cdab9ca916cc0b6f69f7c876590eb/external/distroless_static_linux_arm64/BUILD.bazel:47:18: Copying files to directory distroless_static_linux_arm64/blobs/sha256 failed: (Exit 126): copy_to_directory failed: error executing command (from target @distroless_static_linux_arm64//:blobs)
  (cd /private/var/tmp/_bazel_gus/798cdab9ca916cc0b6f69f7c876590eb/execroot/com_canva_infrastructure && \
  exec env - \
  external/copy_to_directory_linux_amd64/copy_to_directory bazel-out/darwin_arm64-fastbuild/bin/external/distroless_static_linux_arm64/blobs_config.json)
# Configuration: 552dda68697b5bb17c41b4bef2af5919b102e5481624a5bb19489b4de2a07a0c
# Execution platform: //tools/build/bazel/toolchains/remote:ubuntu-act-22-04-platform
external/copy_to_directory_linux_amd64/copy_to_directory: external/copy_to_directory_linux_amd64/copy_to_directory: cannot execute binary file
```

Note bazel _thinks_ the execution platform is my RBE platform (ubuntu-act-22-04-platform), and has selected `copy_to_directory_linux_amd64` appropriately, but the action is _actually_ executed locally on my darwin/amd64 mac and fails ("Exit 126" and "cannot execute binary file").

After much head scratching, I discovered `@aspect_bazel_lib//lib/private/copy_common.bzl` `COPY_EXECUTION_REQUIREMENTS`, which forces copy commands to be performed locally.  This effectively discards all the toolchain resolution hard work, and results in the above error.

### Version

Development (host) and target OS/architectures:

host = darwin/arm64
exec = linux/amd64
target = linux/amd64

Output of `bazel --version`:

bazel 6.1.2

Version of the Aspect rules, or other relevant rules from your
`WORKSPACE` or `MODULE.bazel` file:

Aspect rules v1.32.1

[...]

PR #605, opened 2023-10-09
feat: add force_local_execution attribute to copy rules & macros for opt-out of local execution requirements

gregmagolan:
There are use cases where copy rules should run sandboxed or on remote executors. In these cases users should set `force_local_execution` to False so the rules are not opinionated. Users can set custom execution requirements in their bazelrc via mnemonics if desired. Forcing local execution still seems like the best default for best performance by default when https://github.com/aspect-build/bazel-lib/issues/466 is not an issue.

Related issues & threads,

https://github.com/aspect-build/bazel-lib/issues/466
https://bazelbuild.slack.com/archives/CEZUUKQ6P/p1694892563112649

---

### Type of change

### Test plan

- New test cases added


alexeagle (commented):
There's not a use case for setting this differently on different targets, rather you need execution requirements to be based on how you're executing Bazel.
I think it needs to be a config setting with a custom flag, so you can opt-out of our execution requirements for the entire build?

alexeagle:
replaced with #606

PR #610, opened 2023-10-10
chore: add windows binaries

alexeagle:


thesayyn (approved):

alexeagle:
Note, it's not very hermetic, the bsdtar.exe still requires some DLLs to be on the system.
https://buildkite.com/bazel/bazel-bazel-examples/builds/2166#018b1b6c-5866-46ee-a5ab-fe2dfb3c3446 fails with `STATUS_DLL_NOT_FOUND`.

Going to merge anyway as it's green for us! I worked hard on Windows today and deserve a break :)

PR #613, opened 2023-10-10
chore: up the bzlmod compatibility level

kormide:
### Type of change


alexeagle:
I think this will be highly destructive. It means no one can build a project where bazel-lib compat 1 is reachable through the graph and compat2 is also. I don't think this is necessary.

PR #616, opened 2023-10-10
fix: avoid breaking Bazel 5 users

alexeagle:
Related to #584 which removed a Bazel 5 workaround

Makes `platform_utils` harder to discover, but the generated doc for it was useless, and this is the simplest way to ease the bazel-lib 2.0 rollout.

### Type of change

### Test plan

None


kormide (approved):

alexeagle:
This plan was too simplistic, the platform_utils is loaded by some of our documented helpers, so the `bzl_library` chain can not be broken so easily 😓

PR #626, opened 2023-10-18
feat: implement path transforms

thesayyn:
Adds support for path transforming for the tar rule. Effectively making pkg_tar do everything pkg_tar can do and more. 

### Type of change

### Test plan

- New test cases added

### DIFF
--- a/lib/private/tar.bzl
+++ b/lib/private/tar.bzl
@@ -51,6 +51,7 @@ _tar_attrs = {
 
 _mtree_attrs = {
     "srcs": attr.label_list(doc = "Files that are placed into the tar", mandatory = True, allow_files = True),
+    "transform": attr.string_dict(doc = """A dict for path transforming. These are applied serially in respect to their orders."""),
     "out": attr.output(doc = "Resulting specification file to write"),
 }
 
@@ -124,11 +125,7 @@ def _tar_impl(ctx):
 
     return DefaultInfo(files = depset([out]), runfiles = ctx.runfiles([out]))
 
-def _default_mtree_line(file):
-    # Functions passed to map_each cannot take optional arguments.
-    return _mtree_line(file.short_path, file.path, "dir" if file.is_directory else "file")
-
-def _mtree_line(file, content, type, uid = "0", gid = "0", time = "1672560000", mode = "0755"):
+def _mtree_line(file, content, type, uid = "0", gid = "0", time = "1672560000.000000", mode = "0755"):
     return " ".join([
         file,
         "uid=" + uid,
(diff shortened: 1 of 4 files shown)


lib/tests/tar/BUILD.bazel
```diff
@@ -199,3 +199,30 @@ diff_test(
...
+        "pkg/debian-binary$": "debian-binary",
+    },
+)
+
+assert_tar_listing(
+    name = "test_transform",
+    actual = "transform",
+    expected = [
+        #
+        # TODO: https://github.com/aspect-build/bazel-lib/issues/625
```
jbedard, line 224: What should it look like?

lib/tests/tar/BUILD.bazel
```diff
@@ -199,3 +199,30 @@ diff_test(
...
 )
+
+tar(
+    name = "transform",
+    srcs = [
+        "pkg/debian-binary",
+        "src_file",
+    ],
+    out = "transform.tar",
+    transform = {
```
jbedard, line 210: This doesn't cover the case having no `^` or `$`?

lib/tar.bzl
```diff
@@ -23,7 +23,27 @@ load("@bazel_skylib//rules:write_file.bzl", "write_file")
 load("//lib/private:tar.bzl", "tar_lib", _tar = "tar")
 
 mtree_spec = rule(
-    doc = "Create an mtree specification to map a directory hierarchy. See https://man.freebsd.org/cgi/man.cgi?mtree(8)",
+    doc = """Create an mtree specification to map a directory hierarchy. See https://man.freebsd.org/cgi/man.cgi?mtree(8)
+
+Supports `$` and `^` RegExp tokens, which may be used together.
```
jbedard, line 28: Is this `$`+`^` a standard from somewhere? Having the starts/ends-with but no other regex/glob?

lib/tests/tar/BUILD.bazel
thesayyn: right. will add that.

lib/tests/tar/BUILD.bazel
thesayyn:
```
"drwxr-xr-x  0 0      0           0 Jan  1  2023 package",
"-rwxr-xr-x  0 0      0           0 Jan  1  2023 package/debian-binary",
"-rwxr-xr-x  0 0      0          21 Jan  1  2023 package/PKGINFO",
```

lib/tar.bzl
thesayyn: It's standard regex tokens. https://regex101.com/r/iF7PKY/1

thesayyn: oh, i can only support those tokens as they are trivial and easy to implement.

thesayyn: and since we are applying these transforms serially, anything is possible.

lib/tests/tar/BUILD.bazel
thesayyn: done

lib/tar.bzl
jbedard: I mean is the idea of only supporting those copied from somewhere else. Are we trying to support the same thing as a rule we're replacing? Or is this new functionality now only in `bazel-lib`?

jbedard: If this isn't copying from somewhere else... why regex instead of glob like everywhere else in bazel-lib?

thesayyn: This is only available in bazel-lib. This is for transforming paths, whereas globbing is used for matching paths. do we do something similar with globs here?

lib/private/tar.bzl
```diff
@@ -139,12 +136,32 @@ def _mtree_line(file, content, type, uid = "0", gid = "0", time = "1672560000",
...
+            if match.removeprefix("^").removesuffix("$") == path:
+                path = replace
+        elif match.startswith("^"):
+            if path.startswith(match.removeprefix("^")):
+                path = "".join([replace, path.removeprefix(match.removeprefix("^"))])
+        elif match.endswith("$"):
+            if path.endswith(match.removesuffix("$")):
+                path = "".join([path.removesuffix(match.removesuffix("$")), replace])
+        else:
+            path = path.replace(match, replace)
```
jbedard, line 152: Does `replace` replace all?

lib/tar.bzl
jbedard: After discussing it we figured globs is not the right solution, but I'd like this public API to be reviewed by @alexeagle and maybe others before we introduce an API like this.

lib/private/tar.bzl
thesayyn: yes

jbedard: Do we have a test for that?

lib/tests/tar/BUILD.bazel
jbedard: Should you explain that in the comment? Even should "should also contain X, see: URL"

lib/private/tar.bzl
thesayyn: hmmm, no. i am gonna add a unit test for it now

lib/tests/tar/BUILD.bazel
alexeagle, line 210: I think we should spend a couple more weeks to be very thoughtful in what public API we introduce here

alexeagle:
I think we determined that https://github.com/vbatts/go-mtree/ is the utility we want to bring for users to do this.

PR #630, opened 2023-10-19
feat: support treeartifacts

thesayyn:
### Type of change

### Test plan

- New test cases added

### DIFF
--- a/lib/private/tar.bzl
+++ b/lib/private/tar.bzl
@@ -1,7 +1,5 @@
 "Implementation of tar rule"
 
-load("@aspect_bazel_lib//lib:paths.bzl", "to_rlocation_path")
-
 _tar_attrs = {
     "args": attr.string_list(
         doc = "Additional flags permitted by BSD tar; see the man page.",
@@ -74,9 +72,6 @@ def _add_compress_options(compress, args):
     if compress == "zstd":
         args.add("--zstd")
 
-def _runfile_path(ctx, file, runfiles_dir):
-    return "/".join([runfiles_dir, to_rlocation_path(ctx, file)])
-
 def _calculate_runfiles_dir(default_info):
     manifest = default_info.files_to_run.runfiles_manifest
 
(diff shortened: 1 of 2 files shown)


lib/private/tar.bzl
```diff
@@ -124,37 +119,71 @@ def _tar_impl(ctx):
...
         "mode=" + mode,
         "type=" + type,
-        "content=" + content,
-    ])
+    ]
+    if content:
+        spec.append("content=" + content)
+    return " ".join(spec)
+
+def _to_rlocation_path(file, workspace):
```
alexeagle, line 135: nit: we already have this function?

lib/private/tar.bzl
```diff
@@ -124,37 +119,71 @@ def _tar_impl(ctx):
...
+        # to be transferred to execution phase.
+        workspace_name = str(ctx.workspace_name)
+
+        content.add(_mtree_line(runfiles_dir, type = "dir"))
+        content.add_all(
+            s.default_runfiles.files,
+            expand_directories = True,
+            uniquify = True,
+            format_each = "{}/%s".format(runfiles_dir),
+            # be careful about what you pass to _expand_for_runfiles as it will carry the data structures over to execution phase.
```
alexeagle, line 183: how did you discover this? running some memory profiler?

lib/private/tar.bzl
thesayyn: It wants the whole ctx object which is not good in terms of performance since we are using `allow_closure = True` to configure the workspace name.  So instead of transferring the whole ctx we just copy workspace_name and move to the execution phase.

lib/tests/tar/BUILD.bazel
```diff
@@ -130,7 +134,15 @@ genrule(
     # lib/tests/tar/a uid=0 gid=0 time=1672560000 mode=0755 type=file content=bazel-out/darwin_arm64-opt/bin/lib/tests/tar/a
     # ->
     # a uid=0 gid=0 time=1672560000 mode=0755 type=file content=bazel-out/darwin_arm64-opt/bin/lib/tests/tar/a
-    cmd = "sed s#^{}/## <$< >$@".format(package_name()),
+    cmd = "sed '{}' <$< | sed '/^\\ /d' > $@".format(
```
alexeagle, line 137: oh, that's gross, harder to claim that strip_prefix is doable in userland if it has to look like this 🤔

lib/private/tar.bzl
thesayyn: It is exact copy of the function, but takes workspace string instead of ctx.

thesayyn: It can't really change the shared function as it's part of public API...

alexeagle (approved):
super cool

lib/private/tar.bzl
thesayyn: https://bazel.build/rules/lib/builtins/Args#add_all `allow_closure` explains this.

lib/tests/tar/BUILD.bazel
thesayyn: yeah, I was thinking exactly the same. unfortunately, we have to further process the mtree to remove parent directories...

lib/private/tar.bzl
alexeagle: yes, that totally makes sense 👍🏻

lib/private/tar.bzl
thesayyn: the documentation, https://bazel.build/rules/lib/builtins/Args#add_all allow_closure explains this. However, i did not compare the two.

lib/tests/tar/BUILD.bazel
alexeagle: I suppose there could be an option not to specify parent directories so the mtree file is more easily manipulated, but we'll get to something principled for this soon enough.

Maybe this example wants a little program rather than a bash one-liner...

thesayyn: I was hoping i could easily use something like https://github.com/vbatts/go-mtree but did not explore it further.

Issue #528, opened 2023-09-18
[Bug]: performance slower by default on copy actions


alexeagle:
### What happened?

Some users needed https://github.com/aspect-build/bazel-lib/pull/506
but others liked it how it was, because it guaranteed better performance.

Maybe for 2.0 we should just admit that neither is a good default for everyone, and the setting is hard to discover, so we force users to put some flag in their bazelrc to opt-in to one behavior or the other (like we did with rules_ts 2.0)

### Version

HEAD

alexeagle:
Fixed with #606

PR #639, opened 2023-10-30
fix: fix execution requirements for 'build without the bytes'

gregmagolan:
With "build without the bytes" working correctly and default in Bazel 7, our copy execution requirements should allow caching so that it works with this mode out of the box.

We should also not be opinionated about remote execution since with "build without the bytes", if you're copying all generated files around then it may be more efficient to do it on a remote executor. The subset of users that use remote execution can easily set custom execution requirements for the Mnenomic if desired.

---

### Type of change

### Test plan

### DIFF
--- a/lib/private/copy_common.bzl
+++ b/lib/private/copy_common.bzl
@@ -17,63 +17,18 @@ def execution_requirements_for_copy(ctx):
     # If the rule ctx doesn't expose the CopyOptions, the default is to run locally
     return COPY_EXECUTION_REQUIREMENTS_LOCAL
 
-# When applied to execution_requirements of an action, these prevent the action from being
-# sandboxed or remotely cached, for performance of builds that don't rely on RBE and build-without-bytes.
+# When applied to execution_requirements of an action, these prevent the action from being sandboxed
+# for improved performance.
 COPY_EXECUTION_REQUIREMENTS_LOCAL = {
     # ----------------+-----------------------------------------------------------------------------
-    # no-remote       | Prevents the action or test from being executed remotely or cached remotely.
-    #                 | This is equivalent to using both `no-remote-cache` and `no-remote-exec`.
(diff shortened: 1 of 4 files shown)


alexeagle (commented):
We need to discuss this one, as we've gone back and forth several times now.

https://github.com/aspect-build/bazel-lib/pull/594 points to https://github.com/aspect-build/bazel-lib/issues/528 where we learned that some users rely on this.

We now have two modes in bazel-lib 2.0 https://github.com/aspect-build/bazel-lib/blob/main/docs/copy_file.md#choosing-execution-requirements

I don't understand whether your change here is a complete retreat from local performance opt.

gregmagolan:
It's not a retreat. It keeps the main local development optimization of no-sandbox which is the important but it adds the ability remote cache the actions so that "build without the bytes" is optimal. It shaves off minutes of copying off of a new clone that in fully cached in a large rules_js repository with "build without the bytes" enabled.

lib/private/copy_common.bzl
```diff
@@ -20,60 +20,16 @@ def execution_requirements_for_copy(ctx):
 # When applied to execution_requirements of an action, these prevent the action from being
 # sandboxed or remotely cached, for performance of builds that don't rely on RBE and build-without-bytes.
```
kormide, line 21: Comment out of date.

kormide (approved):

alexeagle:
Sure but do we understand how this new version satisfies all the problems we had with other attempts at this?

gregmagolan:
I now understand how the requirements of "build without the bytes" fit in now after a deep dive last week. I think we should support that mode by default. Not allowing cacheing & remote execution is the wrong default IMO.

kzadorozhny:
This is a very welcome change. Being able to speed up CopyFile actions is a big win.
For example, the total time spent copying 4K+ ts_priject source files is more than 220 seconds.
image

PR #606, opened 2023-10-09
feat: expose a config_setting for copy execution_requirements

alexeagle:
Alternative to #605

Fixes #604
Fixes #466

---

### Type of change

- New feature or functionality (change which adds functionality)

**For changes visible to end-users**

- Suggested release notes are provided below:

If you rely on Remote Execution and Build-without-the-bytes, you should add to your `.bazelrc`:

(with Bazel 6.4 or greater)
`common --@aspect_bazel_lib//lib:copy_use_local_execution=false`

or with Bazel 6.3 or earlier:
```
build --@aspect_bazel_lib//lib:copy_use_local_execution=false
fetch --@aspect_bazel_lib//lib:copy_use_local_execution=false
query --@aspect_bazel_lib//lib:copy_use_local_execution=false
```

### Test plan

- Manual testing; please provide instructions so we can reproduce:
Run aquery locally:

```
$ bazel aquery lib/tests/copy_file/...
action 'Copying file lib/tests/copy_file/file1'
  Mnemonic: CopyFile
  Target: //lib/tests/copy_file:copy_same_file_b
  Configuration: k8-opt
  Execution platform: @local_config_platform//:host
  ActionKey: 51a4a2f509cd84b088d8a06fdd9136b87a9dfc494dbdeb31547395af412881cf
  Inputs: [lib/tests/copy_file/file1]
  Outputs: [bazel-out/k8-opt/bin/lib/tests/copy_file/file1_copy_b]
  Environment: [PATH=/bin:/usr/bin:/usr/local/bin]
  ExecutionInfo: {local: 1, no-cache: 1, no-remote: 1, no-remote-cache: 1, no-remote-exec: 1, no-sandbox: 1}
  Command Line: (exec /bin/bash \
    -c \
    'cp -f "$1" "$2"' \
    '' \
    lib/tests/copy_file/file1 \
    bazel-out/k8-opt/bin/lib/tests/copy_file/file1_copy_b)
# Configuration: c5a436e84a36b625a7552b0800bf8d2dcbfb101629fb9f434c99131217b14a76
# Execution platform: @local_config_platform//:host
  ExecutionInfo: {local: 1, no-cache: 1, no-remote: 1, no-remote-cache: 1, no-remote-exec: 1, no-sandbox: 1}

$ bazel aquery lib/tests/copy_file/... --@aspect_bazel_lib//lib:copy_use_local_execution=false
action 'Copying file lib/tests/copy_file/file1'
  Mnemonic: CopyFile
[...]

PR #640, opened 2023-10-31
Revert: feat: expose a config_setting for copy execution_requirements (#606)

gregmagolan:
Reverts #606.

Feature added in #606 is no longer necessary after #639.


alexeagle (approved):
goodbye forever sweet code

PR #622, opened 2023-10-12
Use coreutils toolchain for copy_file action

dzbarsky:
Use the coreutils toolchain rather than bash/batch scripts to perform the copy.

### Type of change
- Refactor (a code change that neither fixes a bug or adds a new feature)**

- Relevant documentation has been updated
- Suggested release notes are provided below:
- Breaking change (this change will force users to change their own code or config)
Might require toolchain registration where previously it wasn't needed, if users did their own toolchain registration instead of calling `aspect_bazel_lib_register_toolchains` as the docs recommend

### Test plan
- Covered by existing test cases

PR #651, opened 2023-11-14
fix: don't include e2e/smoke bazeliskrc on bcr ci runs

kormide:
Regressed in https://github.com/aspect-build/bazel-lib/pull/622. Now that we create e2e/smoke/.bazeliskrc, the bazel central registry ci fails when we push a release because there we don't publish cli windows binaries. Remove the file from the release archive altogether.

### Type of change

### DIFF
--- a/.github/workflows/release_prep.sh
+++ b/.github/workflows/release_prep.sh
@@ -11,6 +11,15 @@ TAG=${GITHUB_REF_NAME}
 PREFIX="bazel-lib-${TAG:1}"
 ARCHIVE="bazel-lib-$TAG.tar.gz"
 
+# Remove the .bazeliskrc file from the smoke e2e so that the Bazel Central Registry
+# CI doesn't attempt to use the Aspect CLI, which will fail for the Windows run since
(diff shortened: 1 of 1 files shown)


.github/workflows/release_prep.sh
```diff
@@ -11,6 +11,15 @@ TAG=${GITHUB_REF_NAME}
...
 ARCHIVE="bazel-lib-$TAG.tar.gz"
 
+# Remove the .bazeliskrc file from the smoke e2e so that the Bazel Central Registry
+# CI doesn't attempt to use the Aspect CLI, which will fail for the Windows run since
+# we don't publish Windows binaries. Two alternative solutions that were attempted but
+# did NOT work:
+#   1. In .bcr/presubmit.yml, override the env vars BAZELISK_BASE_URL and USE_BAZEL_VERSION
+#   2. Add a Publish To BCR patch under .bcr/patches to remove the file. Worked on the BCR CI
+#      on all platforms except for OSX where the patch command failed.
+rm ./e2e/smoke/.bazeliskrc
```
kormide, line 21: No, this won't work. It won't be included in the archive command below...

PR #652, opened 2023-11-14
fix: don't use aspect cli on bcr ci runs

kormide:
Regressed in https://github.com/aspect-build/bazel-lib/pull/622. Now that we create e2e/smoke/.bazeliskrc, the bazel central registry ci fails when we push a release because there we don't publish cli windows binaries. Remove this file so that the bcr never uses it. This has the consequence that on our ci we will not use the aspect cli for this test on non-windows platforms, but this is closer the bcr environment for which this test was created so it makes sense.

### Type of change

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -146,7 +146,9 @@ jobs:
           if ($BAZEL_MAJOR_VERSION -ne "6") {
             rm -Force .aspect/bazelrc/bazel6.bazelrc
           }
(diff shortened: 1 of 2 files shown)


kormide:
Err wait, I see that we can exclude files in release archives using `.gitattributes`. I think that might be better than this.

kormide:
Closing in favour of https://github.com/aspect-build/bazel-lib/pull/654

PR #655, opened 2023-11-15
Revert "Use coreutils toolchain for copy_file action (#622)"

kormide:
This reverts commit 01ca8f94324f7bc0b2865173a58de06b82dccbac.

Using the coreutils toolchain for copy actions causes downstream issues in rules_js and rules_ts which use these actions. This is difficult to fix forward in rules_js and rules_ts without causing a breaking change by requiring WORKSPACE users to call the `register_coreutils_toolchains()` macro.

Reverting this should not be a breaking change for 2.0.0. If users were using the toolchain, the implementation will now just revert to bash.

### Type of change

### DIFF
--- a/lib/private/copy_file.bzl
+++ b/lib/private/copy_file.bzl
@@ -26,6 +26,63 @@ cmd.exe (on Windows). `_copy_xfile` marks the resulting file executable,
 
 load(":copy_common.bzl", _COPY_EXECUTION_REQUIREMENTS = "COPY_EXECUTION_REQUIREMENTS", _progress_path = "progress_path")
 load(":directory_path.bzl", "DirectoryPathInfo")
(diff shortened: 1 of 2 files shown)


alexeagle:
FYI @dzbarsky 

What's the path to roll this forward again?

DavidZbarsky-at:
Hmm I thought the recommended usage was the macro to register all the toolchains (including coreutils)?

kormide:
> FYI @dzbarsky
> 
> What's the path to roll this forward again?

I think the path is to cut a 2.0.1 release and yank the 2.0.0 release from the BCR.

To re-add this we need to be ready to make breaking changes in rules_js and rules_ts because it becomes necessary for WORKSPACE users to load the `register_coreutils_toolchains()` macro. If that is considered a breaking change... I'm not too sure.

PR #665, opened 2023-11-19
chore: add an e2e to test copy actions

kormide:
An e2e like this would have caught the regression where coreutils was required for copy actions.

### Type of change

### Test plan

- New test cases added


alexeagle (approved):
Nice, thanks!

PR #668, opened 2023-11-21
Add toolchain param to affected actions 

kotlaja:
This is part of migrations of Automatic Exec Groups (AEGs) where a `toolchain` parameter should be set inside actions which use executable or tools from a toolchain. 

AEGs are under an incompatible flag right now, but will be flipped soon in Bazel 7. I've added this change since I was manually fixing rules_nodejs, which depends on bazel-lib. I can't fully test this change since I need at least Bazel 7, but Bazel@Head raises non related errors to my change. That's why there might be more related updates once bazel-lib supports Bazel 7.

Still, as I already mentioned, adding a toolchain parameter is masked by the incompatible flag, so these changes don't do anything right now. But it can help you with your migration in the near future when AEGs are enabled by default and you are forced to update actions which use toolchains. 

Also note that the `toolchain` parameter is available from Bazel 6, but the full logic (and the incompatible flag) will be available from Bazel 7.

Please check [official documentation for Automatic Exec Groups](https://bazel.build/extending/auto-exec-groups) for additional info.

---

### Type of change

### Test plan

- Use (at least) Bazel 7 and add `--incompatible_auto_exec_groups` flag.

### DIFF
--- a/docs/copy_directory.md
+++ b/docs/copy_directory.md
@@ -42,7 +42,7 @@ for more context.
 ## copy_directory_bin_action
 
 <pre>
-copy_directory_bin_action(<a href="#copy_directory_bin_action-ctx">ctx</a>, <a href="#copy_directory_bin_action-src">src</a>, <a href="#copy_directory_bin_action-dst">dst</a>, <a href="#copy_directory_bin_action-copy_directory_bin">copy_directory_bin</a>, <a href="#copy_directory_bin_action-hardlink">hardlink</a>, <a href="#copy_directory_bin_action-verbose">verbose</a>)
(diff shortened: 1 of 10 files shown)


kotlaja:
Hi @alexeagle, Nevena here from Bazel team in Munich! :) Can you please take a look at this PR?

kormide:
Looks like this needs a `bazel run //docs:update` for docs tests.

kormide:
@alexeagle Do you think we want this backported to the `1.x` branch as well?

alexeagle:
We can say that bazel-lib 1.x doesn't support Bazel 7.0 so I don't think we need to backport this. If someone complains I can revisit.

kotlaja:
Sorry, I was on a vacation. Thanks for the help here! 

Hm, `allow maintainers to add commits` was already checked, not sure why it didn't work.
I've run the command now, please take another look.

thesayyn:
@kotlaja i meant to speak with somebody from bazel team about this. does adding this parameter make the action incompatible with older versions of bazel where this attribute isn't recognized or does it get simply ignored?

alexeagle:
Ah, sorry we never merged this, looks like https://github.com/bazel-contrib/bazel-lib/commit/43f8b3e5249cba7167d095d67003a1b908b5ef1d did it.

PR #679, opened 2023-12-07
feat: expose common tar variables

thesayyn:
### Type of change

### Test plan

These are useful variables for downstream users like rules_distroless and will prevent code duplication. Example https://github.com/GoogleContainerTools/rules_distroless/blob/3b7a3de91dde8c8201cb53da586696060702af0a/distroless/private/cacerts.bzl#L74

### DIFF
--- a/lib/private/tar.bzl
+++ b/lib/private/tar.bzl
@@ -1,5 +1,41 @@
 "Implementation of tar rule"
 
+TAR_TOOLCHAIN_TYPE = "@aspect_bazel_lib//lib:tar_toolchain_type"
+
+# https://www.gnu.org/software/tar/manual/html_section/Compression.html
+_ACCEPTED_EXTENSIONS = [
+    ".tar",  # uncompressed,
+    ".gz",  # gzip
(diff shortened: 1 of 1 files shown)


lib/private/tar.bzl
```diff
@@ -198,11 +234,17 @@ tar_lib = struct(
     implementation = _tar_impl,
     mtree_attrs = _mtree_attrs,
     mtree_implementation = _mtree_impl,
+    toolchain_type = TAR_TOOLCHAIN_TYPE,
+    common = struct(
```
alexeagle, line 238: nit: I don't think I've seen a nested struct like this before in our rules, why not just add three members to the outer struct?

alexeagle (approved):

lib/private/tar.bzl
thesayyn: wdyt?

alexeagle: it's just a nit, if you think it's a good pattern to borrow from somewhere else I'm okay with that

PR #691, opened 2023-12-13
fix(tar): propagate testonly attr to mtree_spec

alexeagle:
Not sure if it's worth the time to add an analysis_test for something this simple.


thesayyn (approved):
most of the rulesets that uses macros in the ecosystem suffers from this...  wish bazel would automatically adds well knowns to expansion.

PR #696, opened 2023-12-18
fix #697: use to_repository_relative_path in mtree_spec

### LINKED ISSUE #697: [Bug]: mtree_spec yield different result when built in an external repository
### What happened?

If an mtree_spec rule gets built from a different repository, the target paths will look like `../<repo name>/path/in/repo`.

If the same target gets built inside `<repo name>` the path will be `path/in/repo`.

I suspect this is due to the (unprocessed) use of `short_path` here:
https://github.com/aspect-build/bazel-lib/blob/57bfe2c1653f9c5bf0c90c04b9b661388e390ddd/lib/private/tar.bzl#L185

### Version

Reproducible in this repository.

### How to reproduce

```shell
https://github.com/aspect-build/bazel-lib/pull/696
```

### Any other information?

_No response_

gzm0:
Fixes #697.


gzm0:
I'll try and work this into a proper fix, but just wanted to get it out there to get any thoughts / feedback.

alexeagle (commented):
looks right to me. I just wonder if it's worth creating a new e2e test for this, given those get expensive on CI as they accumulate, vs just adding an analysis test that's like a unit test for the starlark code

alexeagle (approved):
Perfect!!

PR #704, opened 2023-12-21
fix: create windows launcher for bats

thesayyn:
### Type of change

### Test plan


.github/workflows/ci.bazelrc
```diff
@@ -17,3 +17,5 @@ build:rbe --bes_results_url=https://app.buildbuddy.io/invocation/
 build:rbe --bes_backend=grpcs://remote.buildbuddy.io
 build:rbe --remote_timeout=3600
 build:rbe --remote_executor=grpcs://remote.buildbuddy.io
+
+test --test_output=errors
```
kormide, line 21: Was this just for debugging?

thesayyn: Yes, but we can keep it. e2e folder don't include `.aspect` defined flags therefore doesn't have this.

kormide (approved):

jbedard (approved):

PR #708, opened 2023-12-23
Fix prerelease

alexeagle:
### Type of change

### Test plan

I checked this by changing the logic and adding a print statement

### DIFF
--- a/tools/version.bzl
+++ b/tools/version.bzl
@@ -10,4 +10,6 @@ VERSION = "0.0.0" if _VERSION_PRIVATE.startswith("$Format") else _VERSION_PRIVAT
 # NB: When GitHub runs `git archive` to serve a source archive file,
 # it honors our .gitattributes and stamps this file, e.g.
 # _VERSION_PRIVATE = "v2.0.3-7-g57bfe2c1"
+# From https://git-scm.com/docs/git-describe:
+# > The "g" prefix stands for "git"
 IS_PRERELEASE = VERSION == "0.0.0" or VERSION.find("g") >= 0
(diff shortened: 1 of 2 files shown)


tools/version.bzl
```diff
@@ -5,3 +5,9 @@
 _VERSION_PRIVATE = "$Format:%(describe:tags=true)$"
 
 VERSION = "0.0.0" if _VERSION_PRIVATE.startswith("$Format") else _VERSION_PRIVATE.replace("v", "", 1)
+
+# Whether bazel-lib is a pre-release, and therefore has no release artifacts to download.
+# NB: When GitHub runs `git archive` to serve a source archive file,
+# it honors our .gitattributes and stamps this file, e.g.
+# _VERSION_PRIVATE = "v2.0.3-7-g57bfe2c1"
+IS_PRERELEASE = VERSION == "0.0.0" or VERSION.find("g") >= 0
```
kormide, line 13: Will there always be a `g` in what I'm guessing is the shortened sha?

kormide (approved):

tools/version.bzl
thesayyn: maybe right solution is to remove this; https://github.com/aspect-build/bazel-lib/blob/081ab5a87971d9651302e48e721e5e4b711a5eec/.gitattributes#L17?

alexeagle: > The "g" prefix stands for "git" and is used to allow describing the version of a software depending on the SCM the software is managed with

https://git-scm.com/docs/git-describe

@thesayyn I moved it here at Ivo's request in https://github.com/aspect-build/bazel-lib/pull/641 (he wanted for rules_proto and I just want them to all be consistent)

However I think having the real version in version.bzl is a feature, not a bug. We just need to interpret the value which is quite easy.

PR #711, opened 2024-01-04
fix: move flags that should only be set with bazel 6 to bazel6.bazelrc recommended settings

gregmagolan:
The [issue](https://github.com/bazelbuild/bazel/issues/17366) with "build without the bytes" and `--experimental_action_cache_store_output_metadata` in Bazel 6 that was reported in this [blog post](https://meroton.com/blog/bazel-6-errors-build-without-the-bytes/) has been resolved in Bazel 7 so that flag is not needed in bazel7.bazelrc recommended presets.

A new issue with Bazel 7 with "build without the bytes" and `--noexperimental_check_output_files` was found so that flag is moved to bazel6 only. See https://github.com/aspect-build/aspect-cli/pull/594 for more info.

---

### Type of change

- `--experimental_action_cache_store_output_metadata` flag is moved to bazel6.bazelrc recommended flags as the undelrying issue was resolved in Bazel 7
`--noexperimental_check_output_files` is no longer in our general recommended performance flags as it break "build without the bytes" on Bazel 7.0.0. Instead it has been moved to bazel6.bazelrc recommended flags.

### Test plan

Resolved `--noexperimental_check_output_files` issue in aspect-cli repo: https://github.com/aspect-build/aspect-cli/pull/594

Tested `--noexperimental_check_output_files` with Bazel 7 in aspect-cli repo: https://github.com/aspect-build/aspect-cli/pull/595


.aspect/bazelrc/performance.bazelrc
```diff
@@ -1,12 +1,3 @@
-# Speed up all builds by not checking if output files have been modified. Lets you make changes to
-# the output tree without triggering a build for local debugging. For example, you can modify
-# [rules_js](https://github.com/aspect-build/rules_js) 3rd party npm packages in the output tree
-# when local debugging.
-# Docs: https://github.com/bazelbuild/bazel/blob/1af61b21df99edc2fc66939cdf14449c2661f873/src/main/java/com/google/devtools/build/lib/pkgcache/PackageOptions.java#L185
-build --noexperimental_check_output_files
-fetch --noexperimental_check_output_files
-query --noexperimental_check_output_files
```
kormide, line 8: Can the affirmative version go in the bazel6 present and the negative version go in the bazel 7 preset along with some warning comments?

gregmagolan: We could. wtdy @alexeagle ?

gregmagolan: Updated

alexeagle (approved):

PR #714, opened 2024-01-08
feat: add is_bazel_7_or_greater utility function to public API

gregmagolan:
In the future we should make detection of the Bazel version part of the public API of bazel_features and use that. 

Currently, however, the [bazel version comparison functions in bazel_features](https://github.com/bazel-contrib/bazel_features/blob/main/private/util.bzl) are private API so a change needs to be made there to make the public API. Also adding a new non-dev dep on bazel_features here would be breaking for WORKSPACE users which is undesirable atm.

---

### Type of change

### Test plan

### DIFF
--- a/lib/utils.bzl
+++ b/lib/utils.bzl
@@ -6,6 +6,7 @@ default_timeout = utils.default_timeout
 file_exists = utils.file_exists
 glob_directories = utils.glob_directories
 is_bazel_6_or_greater = utils.is_bazel_6_or_greater
+is_bazel_7_or_greater = utils.is_bazel_7_or_greater
 is_bzlmod_enabled = utils.is_bzlmod_enabled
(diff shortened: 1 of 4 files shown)


alexeagle:
Oops we probably should have added a dep on bazel_features when we did the 2.0

lib/utils.bzl
```diff
@@ -6,6 +6,7 @@ default_timeout = utils.default_timeout
 file_exists = utils.file_exists
 glob_directories = utils.glob_directories
 is_bazel_6_or_greater = utils.is_bazel_6_or_greater
+is_bazel_7_or_greater = utils.is_bazel_7_or_greater
```
alexeagle, line 9: bazel6 was here already so it's just extending an existing public API 👍🏻

alexeagle (approved):

PR #716, opened 2024-01-09
feat(list): add `unique` function

mgred:
Add a function that returns a list with unique items. I find this function quite useful and use it here and there. What do you think?

### Type of change

### Test plan

- New test cases added

### DIFF
--- a/lib/private/lists.bzl
+++ b/lib/private/lists.bzl
@@ -1,4 +1,5 @@
 "Function definitions for working with lists"
+load("@bazel_skylib//lib:sets.bzl", "sets")
 
 def every(f, arr):
     """Check if every item of `arr` passes function `f`.
@@ -130,4 +131,4 @@ def unique(arr):
     Returns:
       A new list with unique items
     """
-    return dict([[a, a] for a in arr]).keys()
+    return sets.to_list(sets.make(arr))
--- a/lib/private/lists.bzl
+++ b/lib/private/lists.bzl
@@ -1,5 +1,4 @@
 "Function definitions for working with lists"
-load("@bazel_skylib//lib:sets.bzl", "sets")
 
 def every(f, arr):
     """Check if every item of `arr` passes function `f`.
--- a/lib/private/lists.bzl
+++ b/lib/private/lists.bzl
@@ -117,3 +117,21 @@ def some(f, arr):
         if f(a):
             return True
     return False
+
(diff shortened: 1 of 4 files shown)


lib/private/lists.bzl
```diff
@@ -117,3 +117,21 @@ def some(f, arr):
         if f(a):
             return True
     return False
+
+def unique(arr):
```
alexeagle, line 121: I think most folks do this today by converting to a set. How does this compare? I'd rather have fewer ways to do things, unless there's a significant benefit

mattem: My 2 cents would be to do this with a dict, and keep this function to remove the ambiguity of what the code is doing, but its syntax sugar around the `sets` package in skylib.

```
return dict([[a, a] for a in arr]).keys()
```

alexeagle: I'm good with Matt's answer

lib/private/lists.bzl
```diff
@@ -130,8 +130,4 @@ def unique(arr):
     Returns:
       A new list with unique items
     """
-    res = []
-    for a in arr:
-        if a not in res:
-            res.append(a)
-    return res
+    return dict([[a, a] for a in arr]).keys()
```
mgred, line 133: hey @alexeagle sorry, i got sidetracked from this. I think this solution has a problem when working with structures like `dict`s or `struct`s. Because they cannot be a key in a `dict`.

mgred: I tried the following in a starlark REPL

```
$> foo=[{}, [], "foo", {}]
$> dict([[a, a] for a in foo]).keys()
  * expression:1, in <module>
      dict([[a, a] for a in foo]).keys()
error: Value of type `dict` is not hashable
 --> expression:1:1
  |
1 | dict([[a, a] for a in foo]).keys()
  | ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  |
```

alexeagle: Thanks! You didn't need to do that though, I was relying on the tests and they pointed out the same 
https://github.com/aspect-build/bazel-lib/actions/runs/8058378460/job/22011077037?pr=716

alexeagle (approved):
Okay, I don't have more time for bikeshedding on this, I think it's fine as we don't have any users of it yet.

Issue #701, opened 2023-12-21
[Bug]: coreutils toolchain depends on glibc


malt3:
### What happened?

Similar to the issues with bsdtar, the coreutils toolchain currently used (from uutils) is linking against glibc on Linux.
This fails on some platforms (including NixOS) and is not actually hermetic.
Uutils has a statically linked variant of coreutils available as an alternative release artifact. Could `bazel-lib` use this (either by default or as a configurable option)?

Happy to provide a fix as a PR.

### Version

Development (host) and target OS/architectures:

Output of `bazel --version`:

6.4.0

Version of the Aspect rules, or other relevant rules from your
`WORKSPACE` or `MODULE.bazel` file:
aspect bazel-lib 2.0.3
rules_oci 1.5.1

Language(s) and/or frameworks involved: n/a

### How to reproduce

```shell
Use the `oci_image` rule from rules_oci on a NixOS or alpine Linux host.
The rules has a new output (_index_json_cp), which fails to build:

-session
❯ bazel build //e2e/malicious-join:_malicious-join_image_index_json_cp
INFO: Analyzed target //e2e/malicious-join:_malicious-join_image_index_json_cp (0 packages loaded, 0 targets configured).
INFO: Found 1 target...
ERROR: /home/malte/github.com/edgelesssys/constellation/e2e/malicious-join/BUILD.bazel:41:10: Copying file e2e/malicious-join/malicious-join_image failed: (Exit 1): coreutils failed: error executing command (from target //e2e/malicious-join:_malicious-join_image_index_json_cp) external/coreutils_linux_amd64/coreutils cp bazel-out/k8-opt/bin/e2e/malicious-join/malicious-join_image/index.json bazel-out/k8-opt/bin/e2e/malicious-join/_malicious-join_image_index.json
src/main/tools/process-wrapper-legacy.cc:80: "execvp(external/coreutils_linux_amd64/coreutils, ...)": No such file or directory
Target //e2e/malicious-join:_malicious-join_image_index_json_cp failed to build
Use --verbose_failures to see the command lines of failed build steps.
INFO: Elapsed time: 0.394s, Critical Path: 0.05s
INFO: 2 processes: 2 internal.
FAILED: Build did NOT complete successfully
```

alexeagle:
> Uutils has a statically linked variant of coreutils available as an alternative release artifact. Could bazel-lib use this 

That sounds great! We'd rather have a larger download and get a statically-linked binary.

[...]

PR #706, opened 2023-12-22
fix: coreutils toolchain: Use statically linked linux amd64 variant

malt3:
Uutils has a musl release artifact for linux amd64 and linux arm64. <strike>In the future, it should probably also be possible to add a aarch64 musl toolchain. At the moment, this is not an upstream release artifact.
I'm in the process of [fixing this upstream](https://github.com/uutils/coreutils/issues/5690).</strike>

### Type of change

- Bug fix (fixes #701)

### Test plan

- Covered by existing test cases

PR #720, opened 2024-01-12
Add musl binaries for coreutils

mishazharov:
---

### Type of change

We support centos 7 which has glibc 2.17, which is not compatible with the latest coreutils 

No breaking changes

### Test plan

- New test cases added

* Set `musl=True` to see that registering coreutils which were built with musl still work

Closes https://github.com/aspect-build/bazel-lib/issues/701


mishazharov:
Closing in favor of https://github.com/aspect-build/bazel-lib/pull/706

lib/private/coreutils_toolchain.bzl
```diff
@@ -62,6 +62,15 @@ COREUTILS_VERSIONS = {
...
         },
+        # We casually break the naming convention here
+        "linux_amd64_musl": {
+            "filename": "coreutils-0.0.23-x86_64-unknown-linux-musl.tar.gz",
+            "sha256": "sha256-lZvDYhmmvsETKeJCJBsC4wYblJ2CRvqAVCs62UC1TdM=",
+        },
+        "linux_arm64_musl": {
+            "filename": "coreutils-0.0.23-i686-unknown-linux-musl.tar.gz",
+            "sha256": "sha256-nEQJminqRsOUYqL1dCGeZ4rFIftwdX+PCOE49JQY+ts=",
+        },
```
snakethatlovesstaticlibs, line 73: This is actually an x86 binary, not an arm binary

Issue #698, opened 2023-12-19
[Bug]: bsdtar.sh: libbz2.so.1.0: cannot open shared object file: No such file or directory


lbhdc:
### What happened?

The tar rules failed to find libbz2.so.1.0.

### Version

Development (host) and target OS/architectures:
Fedora 39 Workstation
Uname: 6.6.3-200.fc39.x86_64

Output of `bazel --version`:
aspect 5.8.5 
bazel 6.4.0

Version of the Aspect rules, or other relevant rules from your
`WORKSPACE` or `MODULE.bazel` file:
workspace

Language(s) and/or frameworks involved:
java
running `test //..`. in this repository

### How to reproduce

I first discovered this in my own repo, but was able to reproduce it by running the test suite for this repo.

Clone the repo
```shell
git clone https://github.com/aspect-build/bazel-lib.git .
```

Run tests
```shell
bazel tests //...
```

Failed tests
```shell
FAILED: Build did NOT complete successfully
//lib/tests/tar:test_change_owner                               FAILED TO BUILD
//lib/tests/tar:test_compress                                   FAILED TO BUILD
//lib/tests/tar:test_custom_mtree                               FAILED TO BUILD
//lib/tests/tar:test_dirs                                       FAILED TO BUILD
//lib/tests/tar:test_flags                                      FAILED TO BUILD
```

Error trace
```shell
ERROR: /tmp/tmp.N2umfVRb5Rwork/lib/tests/tar/BUILD.bazel:16:8: Executing genrule //lib/tests/tar:tar_genrule failed: (Exit 127): bash failed: error executing command (from target //lib/tests/tar:tar_genrule) /bin/bash -c ... (remaining 1 argument skipped)

Use --sandbox_debug to see verbose messages from the sandbox and retain the sandbox build root for debugging
external/bsd_tar_linux_amd64/usr/bin/bsdtar: error while loading shared libraries: libbz2.so.1.0: cannot open shared object file: No such file or directory
[...]

Issue #615, opened 2023-10-10
[Bug]: bsd tar toolchain isn't totally hermetic


alexeagle:
### What happened?

Currently we allow the glibc of the machine to be used, so we don't work on Ubuntu 18 for example.

And on Windows we rely on some DLLs so you might get `Exit -1073741515`
which is an integer conversion of `0xC0000135 STATUS_DLL_NOT_FOUND`

https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-erref/596a1078-e883-4972-9bbc-49e60bebca55?redirectedfrom=MSDN

### Version

2.0.0-rc0

dzbarsky:
Yep in https://github.com/dzbarsky/libarchive/blob/master/BUILD.bazel#L40-L61 I link aginst musl to avoid the glibc issue. not sure about Windows, don't have a way to test. Happy to collab on this to provide prebuilt binaries if you're interested

alexeagle:
Sahin found that Nix cache also supplies musl-linked libarchive binaries. Either is fine! I'm just a stickler about this repo not containing any code to build them because I don't want to own it. But you could put libarchive on the BCR!

alexeagle:
@dzbarsky I just got hit by this again at a client. Let's try to get libarchive/libarchive onto the BCR?

DavidZbarsky-at:
Sounds good. Xz is already there so we probably just need zstd and libarchive to get food coverage. Is the BCR CLl the best way to add an overlay to an existing repo?

Strum355:
Just to add to this, Im also facing this (admittedly Im on NixOS)
```
INFO: Analyzed target //cmd/batcheshelper:image_tarball (353 packages loaded, 17402 targets configured).
ERROR: /home/noah/Sourcegraph/sourcegraph/cmd/batcheshelper/BUILD.bazel:57:12: OCI Tarball //cmd/batcheshelper:image_tarball failed: (Exit 127): tarball.sh failed: error executing OCITarball command (from target //cmd/batcheshelper:image_tarball) bazel-out/k8-fastbuild/bin/cmd/batcheshelper/image_tarball/tarball.sh

Use --sandbox_debug to see verbose messages from the sandbox and retain the sandbox build root for debugging
external/bsd_tar_linux_amd64/usr/bin/bsdtar: error while loading shared libraries: libacl.so.1: cannot open shared object file: No such file or directory
```

[...]

PR #732, opened 2024-01-24
Making bsdtar and coreutils fully portable for amd64

manuelnaranjo:
This is kind of in between a bug fix and a new feature.

Basically I realized bazel-lib bsdtar and coreutils don't work in centos7 amd64 because of libc version and stuff.

I got the binaries to the point where those are isolated and the entire dependency tree is self-contained,
could do similar work for arm64, but first I wanted to get some feedback

I expect no visible changes to end-users other than more platforms being supported.

For testing I executed all tests in a centos7 docker image and they passed.


manuelnaranjo:
I see issues #698 and #615 related to this, and I agree the musl approach is more sane as it has less maintenance cost, I wish I opened the issue tracker earlier, on the other hand it took me under 1 hour to get this setup and it helped to unblock my use case.

If you don't mind I think it's best to close this ticket.

PR #737, opened 2024-01-28
perf: remove unnecessary calls to 'to_list'

mattem:
Removes the unnecessary calls to `to_list` to retrieve the single file, when the attr is marked as `allow_single_file`, the file object can be obtained via `ctx.file` without the need to flatten the `files` depset.

---

### Type of change

Access toolchain `bin` via `ctx.file` to remove the need to flatten the files depset.

### Test plan


alexeagle (approved):
nice, how did you find it? Curious if there's something repeatable that belongs in a buildifier lint

mattem:
Was using these as the basis for getting the toolchains working in rules_py and noticed it.

PR #757, opened 2024-02-14
jq: add data file and location expansion support

novas0x2a:
This teaches the jq rule about a `data` attribute and adds conditional support to do bazel location expansion. The expansion is disabled by default because it likely would cause compat concerns (since jq can take arbitrary text as input args).

### Type of change

jq now supports data dependencies and optional location expansion on the args attr

### Test plan

- New test cases added

### DIFF
--- a/docs/jq.md
+++ b/docs/jq.md
@@ -7,7 +7,7 @@ Public API for jq
 ## jq
 
 <pre>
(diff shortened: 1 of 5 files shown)


alexeagle (approved):
neat, thanks!

PR #762, opened 2024-02-21
chore: make it possible to override run_binary#use_default_shell_env

alexeagle:
Still not documented as public API. We think this leads to non-hermeticity and could cause cache misses, so it must be used with care.

Part of https://github.com/aspect-build/rules_js/issues/1303

---

### Type of change

### Test plan
Added a trivial usage.

### DIFF
--- a/lib/tests/run_binary_expansions/BUILD.bazel
+++ b/lib/tests/run_binary_expansions/BUILD.bazel
@@ -53,6 +53,9 @@ run_binary(
     },
     progress_message = "doing some work to make %{output}",
     tool = ":expansions_sh",
+    # May introduce non-determinism; use with care!
+    # See e.g. https://github.com/bazelbuild/bazel/issues/4912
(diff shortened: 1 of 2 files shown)


lib/tests/run_binary_expansions/BUILD.bazel
```diff
@@ -51,6 +51,9 @@ run_binary(
     execution_requirements = {
         "no-cache": "1",
     },
+    # May introduce non-determinism; use with care!
+    # See e.g. https://github.com/bazelbuild/bazel/issues/4912
+    use_default_shell_env = True,
```
jbedard, line 56: I think buildifier needs to run on this?

alexeagle: oops I need to `pre-commit install` on all my git clones again

jbedard (approved):

PR #789, opened 2024-03-14
improvement(tar_toolchains): also search for bsdtar on host platform

### LINKED ISSUE #788: [Bug]: bsdtar toolchain fails on macOS if host tar is gnutar
### What happened?

bazel-lib doesnt provide any non-host bsdtar toolchains on macOS. This causes a hard-fail when the host tar binary is a gnutar implementation instead of bsdtar, as is the case when e.g. running inside a Nix shell. 
It may be expected that in this case, one should provide their own bsdtar toolchain, in which case this is working-as-intended. It would definitely be more convenient though for bazel-lib to also provide that : )

### Version

Development (host) and target OS/architectures: macOS arm64

Output of `bazel --version`: aspect 5.8.19

`WORKSPACE` or `MODULE.bazel` file: 2.5.3

Language(s) and/or frameworks involved: Nix

### How to reproduce

_No response_

### Any other information?

_No response_

pcj:
Fixes #788

---

### Type of change

- Bug fix

### Test plan

No change in test coverage.


pcj:
In nix, `bsdtar` is provided by the `libarchive` package:

```
$ which bsdtar
/nix/store/w9m2y5kpmgs342fnmksirvhiqrmqh08j-libarchive-3.6.2/bin/bsdtar
```

alexeagle:
Thanks! But our plan has been to provide a pre-built `bsdtar` binary for darwin as well - this is replaced by https://github.com/aspect-build/bazel-lib/pull/804

PR #792, opened 2024-03-15
feat: add a helper for rules to work with resource_sets

alexeagle:
This API is poorly designed and needs some help to let rule users pick a value in cases that they aren't also the rule author

Workaround for https://github.com/bazelbuild/bazel/issues/15187

---

### Type of change

### Test plan

I'm using it downstream in rules_ts.

### DIFF
--- a/lib/resource_sets.bzl
+++ b/lib/resource_sets.bzl
@@ -2,11 +2,16 @@
 
 Workaround for https://github.com/bazelbuild/bazel/issues/15187
 
-By default, Bazel allocates 1 cpu and 250M of RAM:
-https://github.com/bazelbuild/bazel/blob/058f943037e21710837eda9ca2f85b5f8538c8c5/src/main/java/com/google/devtools/build/lib/actions/AbstractAction.java#L77
+Note, this workaround only provides some fixed values for either CPU or Memory.
+
+Rule authors who are ALSO the BUILD author might know better, and can
+write custom resource_set functions for use within their own repository.
+This seems to be the use case that Google engineers imagined.
 """
 
 resource_set_values = [
+    "cpu_2",
+    "cpu_4",
     "default",
     "mem_512m",
     "mem_1g",
@@ -16,6 +21,12 @@ resource_set_values = [
     "mem_16g",
     "mem_32g",
 ]
+
+def _resource_set_cpu_2(_, __):
+    return {"cpu": 2}
+
+def _resource_set_cpu_4(_, __):
+    return {"cpu": 4}
 
 def _resource_set_mem_512m(_, __):
     return {"memory": 512}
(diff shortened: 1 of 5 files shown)


lib/resource_sets.bzl
```diff
@@ -0,0 +1,72 @@
...
+    if attr.resource_set == "mem_8g":
+        return _resource_set_mem_8g
+    if attr.resource_set == "mem_16g":
+        return _resource_set_mem_16g
+    if attr.resource_set == "mem_32g":
+        return _resource_set_mem_32g
+    fail("unknown resource set", attr.resource_set)
+
+resource_set_attr = {
+    "resource_set": attr.string(
```
jbedard, line 62: Same comment as in the rules_ts PR. How will we add `cpu` to this?

alexeagle: we can just make a `cpu_1`, `cpu_4`, etc.

We just don't have a nice way to expose the cartesian product of all the cpu and ram combinations.

jbedard: Then how do you specify 1cpu + 1g ram?

alexeagle: I don't think there's a way for us to expose that. As a rule author you can just put that function directly in your rule, because then you're implementing a hard-coded value the way the API was designed.

jbedard: So a rule might add this attribute so a BUILD author can specify a size, but you think a rule author would be the one to add `cpu` because the rule author knows the action (but the BUILD author knows the scale of it)?

Do we want to rename this to `resource_set_mem` if that's all it's for? (idk if that's a good idea, just a thought 🤷)

alexeagle: It's a low-fidelity way to give a few "presets" but doesn't generalize to setting everything you might want.

Would you like me to add a `cpu_2` example here to flesh out the answer?

jbedard: I think that's good for now, but doesn't address the infinite combo of things like `cpu_2_mem4g` etc

jbedard (approved):

lib/resource_sets.bzl
alexeagle: nope, and I think the API was poorly designed since the user can't pass a lambda to a rule to specify this, it's required to be a top-level function. It's the best we can do from outside Bazel IIUC

lib/resource_sets.bzl
```diff
@@ -0,0 +1,88 @@
...
+    "mem_8g",
+    "mem_16g",
+    "mem_32g",
+]
+
+def _resource_set_cpu_2(_, __):
+    return {"cpu": 2}
+
+def _resource_set_cpu_4(_, __):
+    return {"cpu": 4}
```
DavidZbarsky-at, line 29: if you're using 4 cores it's Likely you are using more than 250mB as well. So maybe you want a cross-product? This bazel API is silly

PR #808, opened 2024-04-02
chore: remove buildbuddy RBE

alexeagle:
---

### Type of change

### Test plan


alexeagle:
This is premature since ours isn't ready on GCP yet.

PR #820, opened 2024-04-22
support spaces in status

com6056:
---

### Type of change

Bug fix (change which fixes an issue)

https://bazelbuild.slack.com/archives/C04281DTLH0/p1713216473704709

The old regex didn't support spaces in the workspace status entries, this should hopefully fix that 🎉 

### Test plan

Updated link to new regex101 examples


alexeagle (approved):
nice, thanks!

PR #829, opened 2024-05-01
feat(tar): add ergonomic way to strip_prefix

alexeagle:
The genrule we used was too complex to ask users to copy-paste it

This is a partial step to exposing a rules_pkg `pkg_tar` compatible API.

---

The `tar` package now provides a new `mtree_mutate` macro which allows a more ergonomic equivalent of pkg_tar attributes like `strip_prefix`, `uid`, and more.

### Test plan

- New test cases added

### DIFF
--- a/lib/tar.bzl
+++ b/lib/tar.bzl
@@ -13,13 +13,13 @@ this:
 
 We also provide full control for tar'ring binaries including their runfiles.
 
-## Modifying metadata
+## Mutating the tar contents
 
 The `mtree_spec` rule can be used to create an mtree manifest for the tar file.
-Then you can mutate that spec, as it's just a simple text file, and feed the result
+Then you can mutate that spec using `mtree_mutate` and feed the result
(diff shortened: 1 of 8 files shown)


lib/tar.bzl
```diff
@@ -130,3 +125,18 @@ def tar(name, mtree = "auto", stamp = 0, **kwargs):
         mtree = mtree_target,
         **kwargs
     )
+
+def mtree_mutate(name, mtree, awk_script = "@aspect_bazel_lib//lib/private:modify_mtree.awk", **kwargs):
+    vars = [
+        "-v {}='{}'".format(variable, value)
+        for variable, value in kwargs.items()
```
thesayyn, line 132: we should add a new attribute for variables, we probably don't want arbitrary attrs to be variables, we might want to use them.

alexeagle: made them all explicit parameters.

thesayyn (approved):

PR #834, opened 2024-05-06
fix: only check for blank space between status entries in expand_template

com6056:
---

### Type of change

Bug fix (change which fixes an issue)

https://bazelbuild.slack.com/archives/C04281DTLH0/p1713216473704709

Follow-up from https://github.com/aspect-build/bazel-lib/pull/820

The old regex was including newlines when checking for the second value, so this switches it to only look for spaces/tabs instead

### Test plan

Updated link to new regex101 examples

### DIFF
--- a/tools/expand_template/main.go
+++ b/tools/expand_template/main.go
@@ -71,8 +71,8 @@ func main() {
 }
 
 // captures every `KEY VALUE` line in the status file.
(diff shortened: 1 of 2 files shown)


alexeagle (approved):
Thanks for the fix! It seems like we ought to have a unit test here, could be in a followup.

PR #795, opened 2024-03-20
Refactor mtree_spec to vis encode filenames

pcj:
This PR refactors the `mtree_spec` rule:

Instead of emitting the `.spec` output file directly in it's final formatted form, it has been changed to emit a jsonl equivalent.  This intermediate file is then processed line-by-line in a shell script with `jq` to assist with parsing.  The filenames (and `content`) are encoded with `vis` prior to final mtree line formatting.  A golden test has been added to `e2e/smoke`.

> NOTE: this PR assumes the `vis` tool is present on the host machine (this is generally true for macos and linux).  A future improvement could bundle the vis implementation from `go-mtree`. 

Fixes #794

Issue #794, opened 2024-03-19
[Bug]: `mtree_spec` function does not encode filenames with special characters 


pcj:
### What happened?

The `mtree_spec` rule generates an invalid mtree entries for filenames with special characters.  For example:

```
server.runfiles/pypi/_lock/setuptools@69.1.1/site-packages/setuptools/command/launcher manifest.xml uid=0 gid=0 time=1672560000 mode=0755 type=file content=bazel-out/darwin_arm64-fastbuild-ST-5b66dcbb8b68/bin/external/pypi/_lock/setuptools@69.1.1/site-packages/setuptools/command/launcher manifest.xml
```

Causes:

```
bsdtar: Error reading archive bazel-out/darwin_arm64-fastbuild/bin/server/server_image.packages_tar_manifest.spec: Can't open bazel-out/darwin_arm64-fastbuild-ST-5b66dcbb8b68/bin/external/pypi/_lock/setuptools@69.1.1/site-packages/setuptools/command/launcher
```

Filenames should be encoded according to `vis` (see `man vis`) 

In this case it should be encoded as:

```
server.runfiles/pypi/_lock/setuptools@69.1.1/site-packages/setuptools/command/launcher\040manifest.xml uid=0 gid=0 time=1672560000 mode=0755 type=file content=bazel-out/darwin_arm64-fastbuild-ST-5b66dcbb8b68/bin/external/pypi/_lock/setuptools@69.1.1/site-packages/setuptools/command/launcher\040manifest.xml
```

References:

- https://github.com/vbatts/go-mtree/blob/main/testdata/source.casync-mtree#L127
- https://github.com/vbatts/go-mtree/blob/d54530a564d0beea696c2c616ac0d38ba0b4f5d9/pkg/govis/vis.go#L61
- https://github.com/libarchive/libarchive/blob/master/libarchive/archive_read_support_format_mtree.c#L1942

### Version

```bazel
http_archive(
  name = "aspect_bazel_lib",
  generator_name = "aspect_bazel_lib",
  generator_function = "fetch_dependencies",
  urls = ["https://github.com/aspect-build/bazel-lib/archive/40948111707ff0f491d14ea4bd016b78498d14cf.tar.gz"],
  sha256 = "b69d39dd191bcd6f4ac6d62b0124ea460caf530bb18ea678c9778fe4046c21f9",
  strip_prefix = "bazel-lib-40948111707ff0f491d14ea4bd016b78498d14cf",
)
```

[...]

PR #835, opened 2024-05-07
fix(tar): handle spaces in input filenames

alexeagle:
This is a partial solution to #794 that covers the most common case needing escaping. See #795 for a principled but blocked fix that requires a BSD vis executable

---

### Test plan

- New test cases added


thesayyn (approved):
it's not likely we'll be able to port vis to starlark as starlark doesn't let us know anything about MB strings, nevertheless this is better than nothing.

PR #836, opened 2024-05-10
fix: jq.md toolchains reference

kolloch:
A simple documentation-only fix.


thesayyn:
I don't think this would work, genrule's expect a target that materializes the toolchain type and returns some providers, OTOH toolchain types are not materialized if you just depend on toolchain type.

alexeagle:
Also, if there's a doc-only change, it should reflect an identical edit to an example/ or e2e/ test code change, so we know it's load-bearing and functional.

I'm closing PRs just to clean up the GH repo but please feel free to re-open if we got it wrong.

PR #839, opened 2024-05-13
fix: Set size to a default value as well as timeout.

matts1:
Currently, we are unable to run our `write_source_files` tests in our pre-upload checks, because we have `--test_size_filter=small`, and setting `size` will attempt to set it on both the run rule and the test rule, the former being invalid.

See [here](https://github.com/bazelbuild/bazel/issues/22355) for more details.

### Test plan
`bazel test //... --test_size_filters=small`

Before: 5 tests run
After: 152 tests run

### DIFF
--- a/lib/private/diff_test.bzl
+++ b/lib/private/diff_test.bzl
@@ -106,7 +106,7 @@ _diff_test = rule(
     implementation = _diff_test_impl,
 )
 
-def diff_test(name, file1, file2, size = None, timeout = None, **kwargs):
+def diff_test(name, file1, file2, size = "small", **kwargs):
     """A test that compares two files.
 
     The test succeeds if the files' contents match.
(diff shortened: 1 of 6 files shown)


lib/private/utils.bzl
```diff
@@ -136,6 +136,42 @@ def _file_exists(path):
...
+    `--test_verbose_timeout_warnings`.
+
+    This function can be used in a macro which wraps a testing rule.
+
+    Args:
+        size: the size attribute of a test target
+        timeout: the timeout attribute of a test target
+
+    Returns:
+        "short" if neither is set, otherwise timeout
```
rrbutani, line 167: Nit (`short` is a timeout, not a `size` IIUC):
```suggestion
        "small" if neither is set, otherwise timeout
```

matts1: I just realized a better way to do it, so I completely removed this method.

matts1:
I just realized a better way to do it.

Previously:
* Neither specified -> size = "medium", timeout = "short"
* Timeout only -> size = "medium", timeout = timeout
* size only -> size = size, timeout = inferred by bazel based on size
* both -> size = size, timeout = timeout

Now:
* (changed) Neither specified -> size = "small", timeout = inferred by bazel based on size (so always short)
* (changed) Timeout only -> size = "small", timeout = timeout
* size only -> size = "size", timeout = inferred by bazel based on size
* both -> size = size, timeout = timeout

rrbutani (approved):

matts1:
Any chance we can get this merged?

lib/private/diff_test.bzl
```diff
@@ -116,16 +115,15 @@ def diff_test(name, file1, file2, size = None, timeout = None, **kwargs):
...
       **kwargs: The <a href="https://docs.bazel.build/versions/main/be/common-definitions.html#common-attributes-tests">common attributes for tests</a>.
     """
-
     _diff_test(
         name = name,
         file1 = file1,
         file2 = file2,
-        size = size,
-        timeout = default_timeout(size, timeout),
+        size = size or "small",
```
alexeagle, line 126: It's better to just declare the macro parameter with `def diff_test(name, file1, file2, size = "small", **kwargs):` so the docgen can see the default value

alexeagle (approved):

PR #852, opened 2024-05-23
fix(tar): append slash to top-level directory mtree entries

### LINKED ISSUE #851: [Bug]: `mtree_spec` generates incorrect spec file if the root directory has a file and a directory
### What happened?

I'm using a vendored version of `py_layer` at work (from https://github.com/aspect-build/bazel-examples/blob/fba56194cd925b8d12b0339becbfc10bb9504c26/oci_python_image/py_layer.bzl). `py_layer` uses `mtree_spec` and `tar` from `aspect_bazel_lib` to create multiple tar files which are then passed to `oci_image`. As the input, I'm passing a `py_binary` like this:

```starlark
py_binary(
    name = "project_bin",
    srcs = ["project/__main__.py"],
    deps = [":project_lib"],
    imports = ["."],
    main = "project/__main__.py",
)

py_oci_image(
    name = "image",
    base = "@distroless_python",
    binary = ":project_bin",
    entrypoint = ["/usr/bin/python3", "/project_bin"],
)
```

However, this doesn't work properly. The reason is because the mtree spec generated by `aspect_bazel_lib` looks like this:

```
project uid=0 gid=0 time=1672560000 mode=0755 type=dir
project_bin uid=0 gid=0 time=1672560000 mode=0755 type=file content=bazel-out/k8-fastbuild/...
project/__main__.py uid=0 gid=0 time=1672560000 mode=0755 type=file content=project/__main__.py
```

This unfortunately hits some (frankly very bizarre) rules in the [mtree file format](https://man.archlinux.org/man/mtree.5), specifically this one:

> Relative
> If the first whitespace-delimited word has no / characters, it is the name of a file in the current directory. Any relative entry that describes a directory changes the current directory.

This means that when `bsdtar` processes the first line, it "switches directory" to `project/`, and creates `project/project_bin` in the tar file rather than `project_bin`.

Additionally, we can't use `/project_bin` because `mtree` treats leading slashes as "special":

> Special
> Lines beginning with / are special commands that influence the interpretation of later lines.

`./project_bin` works, but shows up as `./project_bin` in `tar tf` output as well (not sure if that breaks anything). It seems the only proper way to place files at the root directory is to make sure that files at root are always listed before the first directory entry.

### Version

Development (host) and target OS/architectures: Gentoo Linux amd64

Output of `bazel --version`: 7.1.2

`WORKSPACE` or `MODULE.bazel` file:

```
bazel_dep(name = "aspect_bazel_lib", version = "2.7.3")
```

Language(s) and/or frameworks involved: Python

### How to reproduce

```shell
Run the following commands:

$ mkdir repro
$ cd repro
$ mkdir foo
$ touch foo/__main__.py
$ touch bar
$ cat > mtree.spec <<EOF
foo uid=0 gid=0 time=1672560000 mode=0755 type=dir
bar uid=0 gid=0 time=1672560000 mode=0644 type=file content=bar
foo/__main__.py uid=0 gid=0 time=1672560000 mode=0644 type=file content=foo/__main__.py
EOF
$ bsdtar -c -f repro.tar @mtree.spec
$ tar -tf repro.tar
foo/
foo/bar
foo/__main__.py
```

### Any other information?

_No response_

sin-ack:
bsdtar's mtree format has a quirk wherein entries without "/" in their first word are treated as "relative" entries, and "relative" directories will cause tar to "change directory" into the declared directory entry. If such a directory is followed by a "relative" entry, then the file will be created within the directory, instead of at top-level as expected. To mitigate, we append a slash to top-level directory entries.

Fixes #851.

---

### Test plan

- New test cases added: `//:tar_test13`, `//lib/tests/tar:test14`

### DIFF
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -5,6 +5,7 @@ load("@bazel_skylib//rules:copy_file.bzl", "copy_file")
 load("@bazel_skylib//rules:write_file.bzl", "write_file")
 load("@buildifier_prebuilt//:rules.bzl", "buildifier")
 load("//lib:diff_test.bzl", "diff_test")
+load("//lib:tar.bzl", "mtree_spec")
 load("//lib:testing.bzl", "assert_contains")
 load("//lib:utils.bzl", "is_bazel_7_or_greater")
 load("//lib:write_source_files.bzl", "write_source_files")
@@ -135,4 +136,41 @@ bzl_library(
     srcs = ["deps.bzl"],
     visibility = ["//visibility:public"],
     deps = ["@bazel_gazelle//:deps"],
+)
+
+# Test case for mtree_spec: Ensure that multiple entries at the root directory are handled correctly (bug #851)
+# See lib/tests/tar/BUILD.bazel for why this is here.
+write_file(
+    name = "tar_test13_main",
+    out = "13project/__main__.py",
+    content = ["__main__.py"],
+)
+
+write_file(
+    name = "tar_test13_bin",
+    out = "13project_bin",
+    content = ["project_bin"],
+)
+
+mtree_spec(
+    name = "tar_test13_mtree_unsorted",
+    srcs = [
+        ":tar_test13_main",
+        ":tar_test13_bin",
+    ]
+)
+
+# NOTE: On some systems, the mtree_spec output can have a different order.
+#       In order to make the test more robust, we sort the mtree output.
+genrule(
+    name = "tar_test13_mtree",
+    srcs = [":tar_test13_mtree_unsorted"],
+    outs = ["actual13.mtree"],
+    cmd = "sort $< > $@",
+)
+
+diff_test(
+    name = "tar_test13",
+    file1 = "tar_test13_mtree",
+    file2 = "//lib/tests/tar:expected13.mtree",
 )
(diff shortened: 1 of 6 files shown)


alexeagle:
Thanks - you should be able to find somewhere in https://github.com/aspect-build/bazel-lib/blob/main/lib/tests/tar/BUILD.bazel that tickles this case, and assert that the new mtree matches some string

sin-ack:
I tried to introduce a test, however the main problem is that any file I create with `file_write` is created inside `lib/tests/tar/`, and `mtree` doesn't hit the bug. That also made me realize that we need to add this logic to `mtree_mutate` (because of `strip_prefix`), but that doesn't solve the problem that we can't test `mtree_spec` by itself. What should I do here?

sin-ack:
Could I get another look at this please? :)

sin-ack:
Interesting, it seems like the output order of `mtree_spec` is either random or differs between my machine and the CI. Perhaps I could write a genrule that orders the output of `mtree` in such a way to trigger the bug.

sin-ack:
Sorry for the delay, I was on vacation. I've made the tests run the `sort` command on the output, which should hopefully make the order consistent between systems now.

@alexeagle Could I get a re-run of the tests please?

thesayyn (approved):

sin-ack:
Thanks!

PR #853, opened 2024-05-23
fix: coreutils download path for darwin_amd64

mrmeku:
---

### Test plan
Ran `bazel run //:format` locally to completion

### DIFF
--- a/lib/private/coreutils_toolchain.bzl
+++ b/lib/private/coreutils_toolchain.bzl
@@ -51,6 +51,7 @@ COREUTILS_VERSIONS = {
             "sha256": "sha256-9zGLFOMDjUDbraDY/hrE5zFJ0O+QYrvx2wHk3Gw3q/A=",
         },
         "darwin_amd64": {
+            # TODO: remove this version_override hack once 0.0.27 is releases and contains a darwin arm binary
             "version_override": "0.0.23",
             "filename": "coreutils-0.0.23-x86_64-apple-darwin.tar.gz",
             "sha256": "sha256-SswetVAuK/hMK1r9uBvNnKj5JpSgD0bzkbsHTxOabCo=",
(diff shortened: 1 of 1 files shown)


lib/private/coreutils_toolchain.bzl
```diff
@@ -51,6 +51,7 @@ COREUTILS_VERSIONS = {
             "sha256": "sha256-9zGLFOMDjUDbraDY/hrE5zFJ0O+QYrvx2wHk3Gw3q/A=",
         },
         "darwin_amd64": {
+            "version_override": "0.0.23",
```
gregmagolan, line 54: can you add a comment that [0.0.26](https://github.com/uutils/coreutils/releases/tag/0.0.26) doesn't have a apple-darwin in the release artifacts

gregmagolan: @thesayyn do you know why the x86_64 apple-darwin isn't there??

gregmagolan (approved):

lib/private/coreutils_toolchain.bzl
alexeagle: I think it's better to be consistent and downgrade all of them to 0.0.23 rather than get different version depending on platform

thesayyn: eh, their CI is broken because of Github!  macos-latest is an arm runner not an amd64 runner as it used to be. 

https://github.blog/changelog/2024-01-30-github-actions-macos-14-sonoma-is-now-available/

https://github.com/uutils/coreutils/pull/6425

gregmagolan: For reference, this was broken in https://github.com/aspect-build/bazel-lib/pull/706

gregmagolan: That PR both bumped to 0.0.26 incorrectly at the same time as adding musl linking. Downgrading to 0.0.23 for all of the them sounds good to me.

gregmagolan (changes requested):
See my last comment

lib/private/coreutils_toolchain.bzl
gregmagolan: Looks like the bump to 0.0.26 was because 0.0.23 didn't have a -musl linked linux binary for both arm and amd. The version_override hack looks like the least bad option we have for now until 0.0.27 contains the darwin amd binary again after https://github.com/aspect-build/bazel-lib/pull/853/files#r1612212955 lands.

gregmagolan (approved):
🦐

PR #854, opened 2024-05-23
refactor(release): switch release integrity to be dynamic

alexeagle:
This matches rules_py as documented by
https://blog.aspect.build/releasing-bazel-rulesets-rust

It has the benefit that developers no longer get yelled at to vendor some updated integrity hashes into bazel-lib every time they touch the Go sources.

---

### Test plan

Validated on https://github.com/alexeagle/bazel-lib/releases
downloading the latest release and cracking it open looks right.


.bazelrc
```diff
@@ -15,7 +15,11 @@ build --embed_label=v1.2.3
 # Mock versioning command to test the --stamp behavior
 build --workspace_status_command="echo BUILD_SCM_VERSION 1.2.3"
 
-common --compilation_mode opt
+# For releasing, use --workspace_status_command and stamp
+# before adding more flags to the release config make sure it does not
+# affect the hashes of /tools. See tools/release.bzl for opt transition
+# add appropriate commandline transition there to match the configuration.
+common:release -c opt
```
alexeagle, line 22: note: we could also restore the stamp variables and let our Go programs know their own versions again, like we used to prior to september 2023

gregmagolan (approved):
🧙

PR #860, opened 2024-06-01
fix: Directory hidden files in write_source_file.

### LINKED ISSUE #667: [Bug]: write_source_files does not copy hidden files
### What happened?

Files that begin with a `.` are not copied, regardless of whether the source is in the source tree or the output tree.

### Version

Development (host) and target OS/architectures:

Output of `bazel --version`: 6.3.2

`WORKSPACE` or `MODULE.bazel` file:

Language(s) and/or frameworks involved: 2.0.1

### How to reproduce

```shell
Include a hidden file in a write_source_files targets. Observe that it does not get written.
```

### Any other information?

I have a patch in this PR to fix it: https://github.com/aspect-build/rules_js/pull/1366

However, it does not copy executable bits. That may be a separate issue related to https://github.com/aspect-build/bazel-lib/issues/526, however.

jgiles:
Copy and manage hidden files starting with "." in write_source_files.

Previously these files were not supported if they were in the top level of the directory to copy.

Refs: #667 

---

### Changes are visible to end-users: yes/no

### Test plan

- Manual testing: Used this lib version to copy a directory of generated files that included hidden files.

### DIFF
--- a/lib/private/write_source_file.bzl
+++ b/lib/private/write_source_file.bzl
@@ -209,7 +209,7 @@ fi"""]
             executable_dir = "find \"$out\" -type f | xargs chmod -x"
         else:
             # Remove execute/search bit recursively from files bit not directories: https://superuser.com/a/434418
-            executable_dir = "chmod -R -x+X \"$out\"/{{*,.[!.]*}}"
+            executable_dir = "chmod -R -x+X \"$out\""
 
     for in_path, out_path in paths:
         contents.append("""
@@ -233,8 +233,8 @@ else
     chmod -R ug+w "$out" > /dev/null 2>&1 || true
     rm -Rf "$out"/{{*,.[!.]*}}
     mkdir -p "$out"
-    cp -fRL "$in"/{{*,.[!.]*}} "$out"
-    chmod -R ug+w "$out"/{{*,.[!.]*}}
+    cp -fRL "$in"/ "$out"
+    chmod -R ug+w "$out"
     {executable_dir}
 fi
 """.format(
--- a/lib/private/write_source_file.bzl
+++ b/lib/private/write_source_file.bzl
@@ -201,7 +201,7 @@ fi"""]
 
     if ctx.attr.executable:
         executable_file = "chmod +x \"$out\""
-        executable_dir = "chmod -R +x \"$out\"/{{*,.[!.]*}}"
+        executable_dir = "chmod -R +x \"$out\""
     else:
         executable_file = "chmod -x \"$out\""
         if is_macos:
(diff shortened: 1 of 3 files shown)


lib/private/write_source_file.bzl
```diff
@@ -201,15 +201,15 @@ fi"""]
 
     if ctx.attr.executable:
         executable_file = "chmod +x \"$out\""
-        executable_dir = "chmod -R +x \"$out\"/*"
+        executable_dir = "chmod -R +x \"$out\"/{{*,.[!.]*}}"
```
jgiles, line 204: This Bash incantation means "all files, including those starting with a `.`, but not `.` or `..`".

This is necessary as opposed to the the simpler approach of `*` and `.*` because `.*` will include `.` and `..`, which in general can be dangerous with rm, cp, chmod, etc.

kormide:
Thanks! Can you add a test?

lib/private/write_source_file.bzl
```diff
@@ -201,15 +201,15 @@ fi"""]
 
     if ctx.attr.executable:
         executable_file = "chmod +x \"$out\""
-        executable_dir = "chmod -R +x \"$out\"/*"
+        executable_dir = "chmod -R +x \"$out\""
```
jgiles, line 204: For chmod, it's safe to also +x the out dir (or below, to recursively remove execute from files but not directories in the out dir)

lib/private/write_source_file.bzl
```diff
@@ -231,10 +231,10 @@ else
     echo "Copying directory $in to $out in $PWD"
     # in case `cp` from previous command was terminated midway which can result in read-only files/dirs
     chmod -R ug+w "$out" > /dev/null 2>&1 || true
-    rm -Rf "$out"/*
+    rm -Rf "$out"/{{*,.[!.]*}}
     mkdir -p "$out"
-    cp -fRL "$in"/* "$out"
-    chmod -R ug+w "$out"/*
+    cp -fRL "$in"/ "$out"
```
jgiles, line 236: cp is specifically documented that including the trailing `/` will copy the contents, not the directory itself.

jgiles:
Hey @kormide, I took a stab at adding a test, let me know if that's the right structure.

I also adjusted some of the commands - chmod and cp, without `-f`, were issuing an error when one of the glob patterns didn't resolve any files (so, if you had a directory with only hidden or only non-hidden files rather than mixed). The updated commands avoid that.

kormide (commented):
LGTM. Thanks for fixing this old issue! I'll wait for the CI run to pass before clicking approve.

kormide (approved):

Issue #848, opened 2024-05-16
[Bug]: bsdtar.exe fails to run on windows: libcrypto dll not found


peakschris:
### What happened?

With rules_js 1.42.1 (and aspect_bazel_lib 2.7.3) on Windows, running js_test targets fails due to a missing crypto DLL:

D:\workspace>bazel test //src/test --verbose_failures
WARNING: Streamed test output requested. All tests will be run locally, without sharding, one at a time
INFO: Analyzed target //src/test:test (0 packages loaded, 0 targets configured).
ERROR: D:/workspace/BUILD.bazel:6:22: Extracting npm package @types/node@20.12.12 failed: (Exit -1073741515): bsdtar.exe failed: error executing NpmPackageExtract command (from target //:.aspect_rules_js/node_modules/@types+node@20.12.12/pkg)
  cd /d D:/udu/b/nag22qmp/execroot/_main
external\aspect_bazel_lib~~toolchains~bsd_tar_windows_amd64\libarchive\bin\bsdtar.exe --extract --no-same-owner --no-same-permissions --strip-components 1 --file external/aspect_rules_js~~npm~npm__at_types_node__20.12.12/package.tgz --directory bazel-out/x64_windows-opt/bin/node_modules/.aspect_rules_js/@types+node@20.12.12/node_modules/@types/node
# Configuration: d850a7e5673f06961afdd7f68e62a46ae2620e12a4464604ad6d160b42ae6c5f
# Execution platform: @@my_toolchain~//platforms:x64_windows
Target //src/test:test failed to build
INFO: Elapsed time: 1.440s, Critical Path: 0.93s
INFO: 41 processes: 41 internal.
ERROR: Build did NOT complete successfully
//src/test:test    FAILED TO BUILD

Running the exe by double clicking is the easiest way to see what dependency is missing:
![image](https://github.com/aspect-build/rules_js/assets/77508021/87b927f5-3356-4973-bd3a-c7fcd017fc44)

The issue is that libcrypto-1_1-x64.dll is not found

### Version

Development (host) and target OS/architectures:

Output of `bazel --version`:
7.2.0rc1

Version of the Aspect rules, or other relevant rules from your
`WORKSPACE` or `MODULE.bazel` file:
rules_js 1.42.1
aspect_bazel_lib 2.7.3

Language(s) and/or frameworks involved:

peakschris:
This is the relevant open issue https://github.com/libarchive/libarchive/issues/1925

[...]

PR #866, opened 2024-06-16
chore: upgrade bsdtar to 3.7.4

alexeagle:
---

bsdtar is now version 3.7.4 and includes some improvements to how its compiled, see https://github.com/bazelbuild/bazel-central-registry/pull/1916

### Test plan

### DIFF
--- a/lib/private/tar_toolchain.bzl
+++ b/lib/private/tar_toolchain.bzl
@@ -36,24 +36,24 @@ BSDTAR_PLATFORMS = {
 
 BSDTAR_PREBUILT = {
     "darwin_amd64": (
-        "https://github.com/aspect-build/bsdtar-prebuilt/releases/download/v3.7.2-1/tar_darwin_amd64",
-        "97bec05ebdfeb313de611b66c0f4b13e0df743de9677efb93fe0c725eb44a1fd",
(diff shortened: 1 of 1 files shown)


alexeagle:
This still uses the windows .exe from the libarchive distribution, so I guess that doesn't fix #848 
@peakschris I think we need to change https://github.com/aspect-build/bsdtar-prebuilt/blob/main/.github/workflows/release.yml to also build `.exe` - not sure if we can cross-compile for that, or need to add Windows GHA runners

peakschris:
@alexeagle thanks for trying, sounds complex...

thesayyn:
> This still uses the windows .exe from the libarchive distribution, so I guess that doesn't fix #848 @peakschris I think we need to change https://github.com/aspect-build/bsdtar-prebuilt/blob/main/.github/workflows/release.yml to also build `.exe` - not sure if we can cross-compile for that, or need to add Windows GHA runners

i believe cross-compilation works now.

thesayyn (approved):
LGTM, we can fix libarchive-prebuilt to include windows binaries.

alexeagle:
Nice, filed https://github.com/aspect-build/bsdtar-prebuilt/issues/3

PR #868, opened 2024-06-19
perf: reduce concatenation in relative_path

jbedard:
The `bazel_skylib` `paths.join` contains a lot of logic unnecessary in this case that degrades performance.

See https://github.com/bazelbuild/bazel-skylib/blob/1.7.1/lib/paths.bzl#L90-L98

Also removed duplicate `len(..)` calls and removed the use of `zip` and `basename` which both construct a new array.

---

### Test plan

### DIFF
--- a/lib/private/paths.bzl
+++ b/lib/private/paths.bzl
@@ -29,33 +29,32 @@ def _relative_file(to_file, frm_file):
 
         parent_count = to_parent_count - frm_parent_count
 
-    to_segments = _spaths.normalize(_spaths.join("/", to_file)).split("/")[:-1]
-    frm_segments = _spaths.normalize(_spaths.join("/", frm_file)).split("/")[:-1]
+    to_segments = _spaths.normalize(to_file if to_file.startswith("/") else "/" + to_file).split("/")[:-1]
+    frm_segments = _spaths.normalize(frm_file if frm_file.startswith("/") else "/" + frm_file).split("/")[:-1]
 
-    if len(to_segments) == 0 and len(frm_segments) == 0:
+    to_segments_len = len(to_segments)
+    frm_segments_len = len(frm_segments)
+
+    if to_segments_len == 0 and frm_segments_len == 0:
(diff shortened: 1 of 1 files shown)


DavidZbarsky-at:
Before:
image


After:
image

DavidZbarsky-at:
image

lib/private/paths.bzl
```diff
@@ -29,33 +29,30 @@ def _relative_file(to_file, frm_file):
...
     split_point = 1
 
     # if either of the paths starts with ../ then assume that any shared paths are a coincidence
-    if to_segments[0] != ".." and frm_segments != "..":
-        longest_common = []
-        for to_seg, frm_seg in zip(to_segments, frm_segments):
-            if to_seg == frm_seg:
-                longest_common.append(to_seg)
-            else:
+    if to_segments[0] != ".." and frm_segments[0] != "..":
```
jbedard, line 45: Was the missing `[0]` on `frm_segments` a bug? Is there a test that should have caught that?

DavidZbarsky-at:
Seems about the same as the previous one
image
image

lib/private/paths.bzl
gregmagolan: There are tests but guess none caught this case

lib/private/paths.bzl
```diff
@@ -127,7 +126,7 @@ def _to_repository_relative_path(file):
     """
 
     if file.short_path.startswith("../"):
-        return "/".join(file.short_path.split("/")[2:])
+        return file.short_path[file.short_path.find("/", 3) + 1:]
```
gregmagolan, line 129: clever with the start of the search pattern at 3 👍

lib/private/paths.bzl
gregmagolan: Test cases are in `lib/tests/paths_test.bzl`

gregmagolan (approved):
🦖

Issue #603, opened 2023-10-09
[FR]: tar rule should support runfiles with symlinks


alexeagle:
### What is the current behavior?

We had to do some stuff to make js_image_layer work with rules_pkg.

### Describe the feature

Should have feature parity with those, and be able to make clean bazelbuild/examples directories showing container builds for JS, Python, Java, etc with no userland starlark code.

alexeagle:
I think it's okay for this to slip to 2.1 as it's not a breaking change.

alexeagle:
I'm on the fence about shipping 2.0 final so long as you can't replace rules_pkg in our js_image_layer example. @thesayyn WDYT?

thesayyn:
we can land that after 2.0. it's non-breaking change.

alexeagle:
IIUC, We'll probably want to wait for https://github.com/bazelbuild/bazel/pull/19944 to be available in Bazel 7.0 and direct users to upgrade so that we can handle the cases correctly under bzlmod.

liningpan:
Hi! Is there a plan for implementing this? In the absence of this feature, is the recommended way to use a custom binary like `js_image_layer`?

thesayyn:
If you don't care about symlinks tar rule should work for you already, js_image_layer is special as symlinks needs to be preserved in the resulting layers.

liningpan:
IIUC, the `tar` rule basically dereferences symlinks, which causes a lot of duplication in the resulting archive.

manuelnaranjo:
It sounds like https://github.com/aspect-build/bazel-lib/blob/main/lib/private/tar.bzl#L227 is missing a bit for symlinks

manuelnaranjo:
I quickly hacked this https://github.com/bookingcom/bazel-lib/commit/c8ce6b85923bd8e800643f51c9b5c012e4a5c1ec, I'm going to test it in our use case, what you think? Does it look like something we could merge?

alexeagle:
Ooh thanks! @thesayyn could you TAL?

manuelnaranjo:
FYI I tested it for our case and made a few improvements:
* I made the symlinks relative so that if I inspect the .tar the links make sense
* Symlinks can't have path being absolute
* For type=link we need link=target
BTW I love this approach, it makes way more sense to express our container images with rules_oci and mtree manipulation than what the inspect and filter layer with rules_docker. At least now I can easily explain my peers what it does

thesayyn:
[...]

PR #869, opened 2024-06-19
feat: mtree_spec now supports symlinks

manuelnaranjo:
Changing mtree_spec so it can honor symlinks, if the target is a directory then the symlink will be to the directory, we don't need to expand those directories to make sure we follow the structure as expected.

From [issue 603](https://github.com/aspect-build/bazel-lib/issues/603) I could tell the support wasn't available, in the case of our internal Perl rules we rely heavily on symlinks to speed up times and files lookup, so not having the feature was making it harder for us to adopt rules_oci.

After researching the repo and some trialing I came with an implementation following the conventions of the repo that honors symlinks.

### Test plan

- Manual testing; please provide instructions so we can reproduce: have a binary that provide symlinks, the call to mtree_spec should have type=link entries with link= relative path to the target file, we did our tests again Bazel 6.5

### DIFF
--- a/lib/private/tar.bzl
+++ b/lib/private/tar.bzl
@@ -1,6 +1,6 @@
 "Implementation of tar rule"
 
-load("//lib:paths.bzl", "to_repository_relative_path")
+load("//lib:paths.bzl", "relative_file", "to_repository_relative_path")
 
 TAR_TOOLCHAIN_TYPE = "@aspect_bazel_lib//lib:tar_toolchain_type"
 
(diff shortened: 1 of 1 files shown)


thesayyn:
Sorry, this doesn't work at all with `ctx.actions.symlink` and `ctx.actions.declare_symlink`. We actually need a runtime tool similar to js_image_layer to detect symlinks and apply corrections to mtree.

manuelnaranjo:
@thesayyn shall we close it then?

thesayyn:
> @thesayyn shall we close it then?

Yes

thesayyn:
Closing this.

PR #873, opened 2024-06-25
fix(tar): expose package_dir argument in mtree_mutate

gzm0:
This was likely forgotten in #829 when making the parameters explicit during review.

---

- Searched for relevant documentation and updated as needed: yes (stardoc)
- Breaking change (forces users to change their own code or config): no (genrule does not have a package_dir argument)
- Suggested release notes appear below: no (suggest to use commit message / PR title)

### Test plan

- New test cases added

### DIFF
--- a/docs/tar.md
+++ b/docs/tar.md
@@ -95,7 +95,7 @@ Rule that executes BSD `tar`. Most users should use the [`tar`](#tar) macro, rat
 ## mtree_mutate
 
 <pre>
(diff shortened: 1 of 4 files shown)


gzm0:
Heads-up: Potentially interacts with #852. /cc @sin-ack

gzm0:
TY for checking. I doubt it's an issue, but I just wanted you to have a look at it, because it seems you have much more mtree spec experience then I do (I don't have any, I just wanted to prepend a string :man_shrugging: ).

After re-reading the description on #852, it indeed looks like the two changes should be independent.

Issue #747, opened 2024-02-06
[FR]: Add support for runtime_toolchain_type for jq/yq/...


BoleynSu:
### Describe the feature

The current toolchain only provides the binary for use during the build but no binary for the build output. People already misuse the cfg=exec binary in a cfg=target output, e.g. https://github.com/bazel-contrib/rules_oci/blob/main/oci/private/push.sh.tpl. Providing another runtime_toolchain_type (similar to what rules_java does) can help improve the current situation. Alternatively, we can provide both exec and target binary in the current toolchain and ask users to use the correct binary.

alexeagle:
We didn't intend to provide jq or yq for the target platform, and from this issue I don't understand why that would be useful. the rules_oci `push.sh` you point to is meant to be run on the host machine with `bazel run` so it shouldn't need `cfg=target` tools.

anguslees:
> from this issue I don't understand why that would be useful

Eg: using jq/yq in executable rules (ie: bazel tests or bazel run scripts)

PR #875, opened 2024-07-01
feat: Implement binary_toolchain

BoleynSu:
See #747


lib/binary_toolchain.bzl
alexeagle, line 1: This seems very similar to https://github.com/bazel-contrib/rules_oci/pull/590 by @edshouten

I wonder if bazel-lib could provide the abstraction that's needed for this case globally. We should bike-shed on this name because "BinaryInfo" can also apply to a binary run as an action on the exec platform.

Issue #877, opened 2024-07-09
[FR]: tar is non-hermetic when compress is set to zstd


ewhauser:
### What is the current behavior?

When `tar` is configured using `compress = "zstd"`, it looks for `zstd` on the path. If the `zstd_toolchain` is registered, then we should append this to path.

### Describe the feature

We want to make `tar` hermetic when using `zstd` compression.

alexeagle:
The bug is actually in the BCR entry for libarchive, see https://github.com/aspect-build/bazel-lib/pull/878#issuecomment-2225981766

mattem:
[Related to a Slack conversation](https://aspect.unthread.io/dashboard/inbox/all/conversations/2ac6b7f8-dbad-4346-a55a-d90afe6c0a15)

PR #878, opened 2024-07-09
feat: optionally register zstd toolchain for tar rule if it is available #877

ewhauser:
---

### Changes are visible to end-users: yes/no

- Searched for relevant documentation and updated as needed: yes/no
- Breaking change (forces users to change their own code or config): yes/no
- Suggested release notes appear below: yes/no

### Test plan

- New test cases added


alexeagle (commented):
We ought to have some usage of this in an example or test as well.

You could add a commit for that, or just point me to what the intended usage looks like and I'll add it.

ewhauser:
If the hermetic zstd toolchain is registered, then it will be on the `path` for `tar` which I'd argue is just expected behavior.

If we were going to document it, then maybe here? https://docs.aspect.build/rulesets/aspect_bazel_lib/docs/tar/#compress

alexeagle:
I see, `lib/tests/zstd/BUILD.bazel` creates a compressed tar file with a genrule, switching that to `tar` does exercise this codepath. It's just missing an assertion that it works without `zstd` on the PATH.

alexeagle:
But I am still not understanding - I tested with this diff

```
--- a/lib/tests/zstd/BUILD.bazel
+++ b/lib/tests/zstd/BUILD.bazel
@@ -1,21 +1,17 @@
+load("@aspect_bazel_lib//lib:tar.bzl", "tar")
 load("@aspect_bazel_lib//lib:testing.bzl", "assert_archive_contains")
 
 # Case 1: Can decompress gzip archive
-genrule(
+tar(
     name = "tar",
-    srcs = [
-        "srcfile",
-    ],
-    outs = ["1.tar.gz"],
-    cmd = "$(BSDTAR_BIN) --create --gzip --dereference --file $@ -s '#$(BINDIR)##' $(execpath srcfile)",
-    toolchains = ["@bsd_tar_toolchains//:resolved_toolchain"],
+    srcs = ["srcfile"],
+    out = "1.tar.zst",
+    compress = "zstd",
 )
```

and then 

```
alexeagle@aspect-build bazel-lib % brew uninstall --ignore-dependencies zstd
Uninstalling /opt/homebrew/Cellar/zstd/1.5.6... (31 files, 2.2MB)
alexeagle@aspect-build bazel-lib % zstd
zsh: command not found: zstd
alexeagle@aspect-build bazel-lib % bazel test lib/tests/zstd/...    
```

to force that `tar` has to use a hermetic `zstd` and it still works. So I'm failing to reproduce a bug that this PR fixes.

ewhauser:
I pushed a canonical example for `bazel-examples`: https://github.com/aspect-build/bazel-examples/compare/main...ewhauser:bazel-examples:zstd-example?expand=1

Without the patch, I get:

```
ERROR: /workspaces/bazel-examples/oci_python_image/hello_world/BUILD.bazel:37:13: Tar hello_world/image.packages_layer.tar.zst failed: (Exit 1): tar failed: error executing command (from target //hello_world:image.packages_layer) external/aspect_bazel_lib~2.7.8~toolchains~bsd_tar_linux_amd64/tar --create --zstd --file bazel-out/k8-fastbuild-ST-50e46de505e1/bin/hello_world/image.packages_layer.tar.zst ... (remaining 1 argument skipped)

Use --sandbox_debug to see verbose messages from the sandbox and retain the sandbox build root for debugging
tar: Can't launch external program: zstd --no-check -3
ERROR: /workspaces/bazel-examples/oci_python_image/hello_world/BUILD.bazel:37:13: Tar hello_world/image.interpreter_layer.tar.zst failed: (Exit 1): tar failed: error executing command (from target //hello_world:image.interpreter_layer) external/aspect_bazel_lib~2.7.8~toolchains~bsd_tar_linux_amd64/tar --create --zstd --file bazel-out/k8-fastbuild-ST-50e46de505e1/bin/hello_world/image.interpreter_layer.tar.zst ... (remaining 1 argument skipped)

Use --sandbox_debug to see verbose messages from the sandbox and retain the sandbox build root for debugging
tar: Can't launch external program: zstd --no-check -3
INFO: Elapsed time: 5.130s, Critical Path: 2.19s
INFO: 58 processes: 48 internal, 2 local, 8 processwrapper-sandbox.
FAILED: Build did NOT complete successfully
```

because the toolchain is not on the path and tar can't find it.

alexeagle:
Thanks for the repro!

Too bad it didn't repro on the local CI for this repo https://github.com/aspect-build/bazel-lib/pull/879
I'll look more into why that is tomorrow...

alexeagle:
Ah so @dzbarsky explains to me that the bug is really in another layer.   :princess-in-other-castle:

The fact that `tar` is looking on the `$PATH` is a sign that we failed to statically link zstd into the binary which was our intent. It is simpler to statically link of course, plus in the (admittedly obscure) use case where someone builds their own rule or uses genrule, they wouldn't pick up this fix to our `tar` rule.

here is the relevant path that falls back to external zstd https://github.com/libarchive/libarchive/blob/586a9645102c87d83919015a9e2e49aec7d47a63/libarchive/archive_write_add_filter_zstd.c#L139-L157 - this is the real bug.

libarchive gets compiled with `HAVE_ZSTD` here: [https://github.com/bazelbuild/bazel-central-registry/blob/746a2218044abcbeed47cd9d[…]7/modules/libarchive/3.7.4.bcr.2/patches/build_with_bazel.patch](https://github.com/bazelbuild/bazel-central-registry/blob/746a2218044abcbeed47cd9d725f82ab06ea7187/modules/libarchive/3.7.4.bcr.2/patches/build_with_bazel.patch#L1511)

There's also logic to check for `HAVE_ZSTD_compressStream` https://github.com/search?q=repo%3Alibarchive%2Flibarchive%20HAVE_ZSTD_compressStream&type=code, suspect we need to add that to the above patchfile

(Note, the BUILD file also sets HAVE_LIBZSTD_COMPRESSOR but that appears unused in libarchive...)

The fault is in bazel-central-registry so perhaps we should transfer the issue there. Do you think you have some spare cycles to track it down?

ewhauser:
That makes sense; will close this PR. Will try to take a look at the upstream issue.

Issue #748, opened 2024-02-06
`copy_to_directory` includes all files as runfiles


dieortin:
### What happened?

When I use `copy_to_directory` in any step of my build, all copied files get included into archives generated with `pkg_tar` and `include_runfiles=True`. This makes it so, in addition to our binaries, we also get source files in the archive.
This is because all files are added to the `runfiles` member of `DefaultInfo`: https://github.com/aspect-build/bazel-lib/blob/71a80525891942267bf1335d830652a0115dcfce/lib/private/copy_to_directory.bzl#L294C28-L294C36

Would it be possible to add an attribute to control whether files are added to runfiles or not?

### Version

Development (host) and target OS/architectures: SUSE SLE RT 15 SP4

Output of `bazel --version`: bazel 6.5.0

Version of the Aspect rules, or other relevant rules from your
`WORKSPACE` or `MODULE.bazel` file: 1.37.0

Language(s) and/or frameworks involved:

alexeagle:
I could imagine introducing a `data` attribute next to `srcs`, but it would be a breaking change to stop including the `srcs` in runfiles so I'm not sure that's feasible.

dieortin:
That’s a good idea. 
Our current workaround is to have a custom rule that removes all runfiles from a target (we’re copying only C++ source files in our case)

dieortin:
What about adding an attribute that does not add to runfiles? That would make the change non-breaking, although I’m not sure about naming

PR #886, opened 2024-07-25
feat: add an option to not include copy_to_directory output in runfiles

kormide:
Fixes https://github.com/aspect-build/bazel-lib/issues/748.

---

- Breaking change (forces users to change their own code or config: no

Added an attr `add_directory_to_runfiles` to `copy_to_directory` to not include the output directory in the target's runfiles.

### Test plan


kormide:
I'm having trouble testing this one due to the issue with `sh_test` I pointed out for the existing test that checks that files are added to runfiles. Even after excluding them they still appear because sh_test includes default outputs of targets passed to `data` in its own runfiles. Can you think of a way to test this? Should I write a custom sh_test just for these tests?

kormide:
@dieortin Does this solve your use case?

dieortin:
> @dieortin Does this solve your use case?

I haven't tested it, but it should solve it, yes!

alexeagle (approved):

PR #889, opened 2024-07-29
feat: export `extension_utils`

mgred:
### Summary

This PR exposes `extension_utils` from `lib/private/extension_utils.bzl`. 
I tend to copy this kind of functionality from extension to extension, so it would be really helpful to have this great method available in the public API. 

I hope this is fine and other users might benefit from it, too :smile: 

### Test plan

Do we need tests?


thesayyn (approved):
LGTM, though it was @gregmagolan or @kormide who wrote these utils.

kormide:
My concern with making this public was that I don't know if it's high enough fidelity and stable enough that we want to support the current iteration. I'll take another look through it..

fyi @alexeagle

kormide:
Yeah, I don't think I want to make extension_utils public in it's current form. The `toolchain_repos_bfs` macro finds the toolchain version closest to the root but the right solution is probably to find the highest version in the dependency graph, allowing the root to override it.

mgred:
thanks for the feedback. @kormide I can understand this decision. Currently I include the macro like [this](https://github.com/bzlparty/tools/tree/main/vendor/aspect_bazel_lib). I hope this is fine for you.

PR #901, opened 2024-08-14
chore: turn off bzlmod misguided warning

alexeagle:
These are misinformed, the module resolver should be permitted to find an MVS solution


kormide (approved):

jbedard (approved):

alexeagle:
FYI @gregmagolan - you didn't originally imagine this being a recommendation for our end users, but discussed: you don't feel strongly enough to revert this

PR #905, opened 2024-08-14
chore(deps): update coreutils to v0.0.27

plobsing:
This release has an `aarch64-apple-darwin` binary, eliminating the need for a `version_override` hack to support that platform.

---

### Test plan

```
bazel test ...
cd e2e/smoke
bazel test ...
```

All previously passing tests still pass; `//lib/tests/transitions:transitioned_go_test_arm64` fails in the same way it had before (on a Darwin / ARM machine).

### DIFF
--- a/lib/private/coreutils_toolchain.bzl
+++ b/lib/private/coreutils_toolchain.bzl
@@ -34,35 +34,33 @@ COREUTILS_PLATFORMS = {
     ),
 }
 
-DEFAULT_COREUTILS_VERSION = "0.0.26"
+DEFAULT_COREUTILS_VERSION = "0.0.27"
 
 # https://github.com/uutils/coreutils/releases
 #
 # The integrity hashes can be automatically fetched for the coreutils releases by running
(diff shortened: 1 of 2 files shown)


lib/private/coreutils_toolchain.bzl
```diff
@@ -34,35 +34,33 @@ COREUTILS_PLATFORMS = {
...
 
-DEFAULT_COREUTILS_VERSION = "0.0.26"
+DEFAULT_COREUTILS_VERSION = "0.0.27"
 
 # https://github.com/uutils/coreutils/releases
 #
 # The integrity hashes can be automatically fetched for the coreutils releases by running
 # `tools/coreutils_mirror_release.sh`.
 COREUTILS_VERSIONS = {
-    "0.0.26": {
```
alexeagle, line 44: nit: removing 0.0.26 could be breaking, in case someone has specified that version rather than the default.

I restored that one, just prepending the newest.

alexeagle (approved):

PR #907, opened 2024-08-15
fix: moving the preserve mtiem test logic to Go for portability

jpinkul:
The original bash version of the preserve mtime test doesn't work for Mac / Windows. For better portability I changed this to a `go_test`.

### Test plan

I've run this locally on a Linux machine to verify the basic functionality and ensured the test still fails if the `preserve_mtime` attributes are commented out. CI will be needed to test this on Mac / Windows.


jpinkul:
I think this feature is behaving unexpectedly with Bazel's caching but I'm not sure why. This sequence is not working as I expected:

1. Run `bazel clean --expunge`
2. Comment out `preserve_mtime = True` from one of the inputs to the test.
3. Run the test with `bazel test :test_preserve_mtime` and it fails as expected.
4. Add `preserve_mtime = True` back and run the test again. It now passes as expected.
5. Comment out `preserve_mtime = True` a 2nd time and run the test again. This time the test still passes with the cached test result.

For that last test run I would expect the test to not be cached since a command line argument changed for the copy directory action this test depends on.

alexeagle:
@thesayyn observes that RBE doesn't support modification times at all. I don't think that's the cause of the red CI here though. Sahin do you have any ideas?

Also, our CI here will trigger macos if `macos` is in the branch name and windows if `windows` is in the branch name. I expect that if you re-open the PR from a branch using the naming convention you'll be able to check those tests.

jpinkul:
> @thesayyn observes that RBE doesn't support modification times at all. I don't think that's the cause of the red CI here though. Sahin do you have any ideas?

Yeah I'm realizing this as well and I don't think remote caching is also incompatible with this feature. If the artifacts are copied from a remote cache then that process is probably going to reset the modify times. I tried adding the `no-remote` tag to the test targets and now the Linux version of these tests appears to be working on CI. 

For that unexpected sequence I posted earlier I'm guessing that both actions with `preserve_mtime = True` and `preserve_mtime = False` were hashing to the same cache value since the contents on disk was identical in both cases. I haven't dug that deep into how Bazel handles the various cache keys though so this is a guess. I added the `external` tag to the test only though and it now passes / fails as I expected.

These two caveats should definitely be added to the documentation.

jpinkul:
I documented these failure modes and pushed a new PR to trigger the mac/windows CI #908

Issue #891, opened 2024-07-31
[Bug]: `run_binary()` does not support `args` that contain spaces


walkerburgin:
### What happened?

`run_binary()` does not correctly pass arguments with spaces in them to `tool`.

### Version

Development (host) and target OS/architectures: MacOS

Output of `bazel --version`: `bazel 7.2.1` 

Version of the Aspect rules, or other relevant rules from your
`WORKSPACE` or `MODULE.bazel` file:
- bazel_dep(name = "aspect_bazel_lib", version = "2.7.7")

Language(s) and/or frameworks involved: None

### How to reproduce

Clone https://github.com/walkerburgin/run-binary-repro

Run `bazel build -s :repro_one`:

```console
➜  run-binary-repro git:(master) ✗ bazel build -s :repro_one
INFO: Analyzed target //:repro_one (0 packages loaded, 0 targets configured).
SUBCOMMAND: # //:repro_one [action 'RunBinary repro_one.txt', configuration: 096dcc84165363e69a851ebe8131b032f5448c94ddc4951775429dc78e79f898, execution platform: @@platforms//host:host, mnemonic: RunBinary]
(cd /private/var/tmp/_bazel_wburgin/cb3bef35f39127f502d8d180ce497830/execroot/_main && \
  exec env - \
  bazel-out/darwin_arm64-opt-exec-ST-d57f47055a04/bin/debug_args --output-file bazel-out/darwin_arm64-fastbuild/bin/repro_one.txt value with spaces)
# Configuration: 096dcc84165363e69a851ebe8131b032f5448c94ddc4951775429dc78e79f898
# Execution platform: @@platforms//host:host
INFO: Found 1 target...
Target //:repro_one up-to-date:
  bazel-bin/repro_one.txt
INFO: Elapsed time: 0.141s, Critical Path: 0.03s
INFO: 2 processes: 1 internal, 1 darwin-sandbox.
INFO: Build completed successfully, 2 total actions
```

Observe that `value with spaces` is not being quoted, and that the script received multiple arguments:

```console
➜  run-binary-repro git:(master) ✗ cat bazel-bin/repro_one.txt
• value
• with
• spaces
```

Run `bazel build -s :repro_two`:

```console
➜  run-binary-repro git:(master) ✗ bazel build -s :repro_two
INFO: Analyzed target //:repro_two (96 packages loaded, 494 targets configured).
[...]

PR #909, opened 2024-08-19
fix: correctly split quoted args

gregmagolan:
Fixes https://github.com/aspect-build/bazel-lib/issues/891 (requested by Bazel Open Source Support customer).

---

New function `split_args` is added to the public API of `strings.bzl`.

This fix touches a number of different rules:
- `jq` rule expansions now include make vars and rule fixed to single quote args that contain spaces
- `expand_template` optimized to not to bother with pointless `.split(" ")` followed by `" ".join()`
- `bats_test` optimized to not to bother with pointless `.split(" ")` followed by `" ".join()`
- `params_file` rule fixed to correctly split args on quotes
- `run_binary` rule fixed to correctly split args on quotes 

### Test plan

- New test cases added

Unit tests added for the new `split_args` function that handles parsing args and splitting them taking into account single quotes, double quotes and escapes.

### DIFF
--- a/lib/private/run_binary.bzl
+++ b/lib/private/run_binary.bzl
@@ -17,6 +17,7 @@
 load("@bazel_skylib//lib:dicts.bzl", "dicts")
 load("//lib:stamping.bzl", "STAMP_ATTRS", "maybe_stamp")
 load(":expand_variables.bzl", "expand_variables")
+load(":strings.bzl", "split_args")
 
 def _run_binary_impl(ctx):
     args = ctx.actions.args()
@@ -46,15 +47,11 @@ Possible fixes:
             rule_kind = str(ctx.attr.tool.label),
         ))
 
-    # `expand_locations(...).split(" ")` is a work-around https://github.com/bazelbuild/bazel/issues/10309
-    # _expand_locations returns an array of args to support $(execpaths) expansions.
-    # TODO: If the string has intentional spaces or if one or more of the expanded file
-    # locations has a space in the name, we will incorrectly split it into multiple arguments
     for a in ctx.attr.args:
-        args.add_all([expand_variables(ctx, e, outs = outputs) for e in ctx.expand_location(a, targets = ctx.attr.srcs).split(" ")])
+        args.add_all(split_args(expand_variables(ctx, ctx.expand_location(a, targets = ctx.attr.srcs), outs = outputs)))
     envs = {}
     for k, v in ctx.attr.env.items():
-        envs[k] = " ".join([expand_variables(ctx, e, outs = outputs, attribute_name = "env") for e in ctx.expand_location(v, targets = ctx.attr.srcs).split(" ")])
+        envs[k] = expand_variables(ctx, ctx.expand_location(v, targets = ctx.attr.srcs), outs = outputs, attribute_name = "env")
 
     stamp = maybe_stamp(ctx)
     if stamp:
--- a/lib/private/strings.bzl
+++ b/lib/private/strings.bzl
@@ -583,3 +583,73 @@ def hex(number):
         hex_string = "0"
 
     return "{}0x{}".format("-" if is_signed else "", hex_string)
+
+def split_args(s):
+    """Split a string into a list space separated arguments
+
+    Unlike the naive `.split(" ")`, this function takes quoted strings
+    and escapes into account.
+
+    Args:
+        s: input string
+
+    Returns:
+        list of strings with each an argument found in the input string
+    """
+    args = []
+    arg = ""
+    single_quote = False
+    double_quote = False
+    escape = False
+    for c in s.elems():
+        if c == "\\":
+            escape = True
+            continue
+        if escape:
+            # this is an escaped character
+            if c == " ":
+                # a dangling escape is not an escape, put the backslack back
+                arg = arg + "\\"
+            else:
+                escape = False
+        else:
+            # not an escaped character, look for quotes & spaces
+            if c == "'":
+                # single quote char
+                if double_quote:
+                    # we're in a double quote so single quotes are just chars
+                    pass
+                elif single_quote:
+                    # end of single quote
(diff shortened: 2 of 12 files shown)


lib/private/expand_locations.bzl
```diff
@@ -32,6 +32,9 @@ def expand_locations(ctx, input, targets = []):
 
     Returns:
       The expanded path or the original path
+
+    Deprecated:
```
jbedard, line 36: Why was this deprecated? Maybe put that in it's own PR instead of hiding it under "correctly split quoted args"?

lib/tests/base64_tests.bzl
```diff
@@ -45,5 +46,5 @@ base64_test = unittest.make(_base64_test_impl)
 def base64_test_suite():
     unittest.suite(
         "base64_tests",
-        base64_test,
+        partial.make(base64_test, timeout = "short"),
```
jbedard, line 49: What were all these for?

gregmagolan: Resolves warnings. We have the flag set in this repo that warns if tests are not sized correctly.

lib/private/expand_locations.bzl
gregmagolan: Mentioned it in the PR summary. Its a pointless since it just calls straight through to `ctx.expand_location` and returns the result. In the past their used to be an implementation and it was dropped a while ago.

I can break it out into a separate PR.

lib/private/run_binary.bzl
```diff
@@ -47,15 +47,11 @@ Possible fixes:
             rule_kind = str(ctx.attr.tool.label),
         ))
 
-    # `expand_locations(...).split(" ")` is a work-around https://github.com/bazelbuild/bazel/issues/10309
-    # _expand_locations returns an array of args to support $(execpaths) expansions.
-    # TODO: If the string has intentional spaces or if one or more of the expanded file
-    # locations has a space in the name, we will incorrectly split it into multiple arguments
     for a in ctx.attr.args:
-        args.add_all([expand_variables(ctx, e, outs = outputs) for e in expand_locations(ctx, a, ctx.attr.srcs).split(" ")])
+        args.add_all(split_args(expand_variables(ctx, ctx.expand_location(a, targets = ctx.attr.srcs), outs = outputs)))
```
jbedard, line 51: Can we split only if `expand_location` expanded to more then one file? Do we want to be splitting every arg 100% of the time? What if an arg is intentionally `foo bar` and we just want `args.add("foo bar")` to quote it for us?

jbedard: That might be a bigger question for a major semver release though? :/

lib/private/strings.bzl
```diff
@@ -583,3 +583,73 @@ def hex(number):
         hex_string = "0"
 
     return "{}0x{}".format("-" if is_signed else "", hex_string)
+
+def split_args(s):
```
jbedard, line 587: Does performance matter here? I'm trying to think of how frequently this would be invoked for the typical use case. Can you think of common cases where this would end up being invoked a very frequently?

lib/private/run_binary.bzl
gregmagolan: I thought about this and I think the answer is that if it expands to multiple files then it is split _unless_ you quote it in the args: 

```
args = ["$(location //multiple_files_target)"]
```
will expand to multiple args the same as
```
args = ["--one --two --three"]
```
would expand to multiple args

but you can prevent the expansion with quotes (single or double)

```
args = ["'$(location //multiple_files_target)'"]
```
or
```
args = ["\"$(location //multiple_files_target)\""]
```
or
```
args = ["'my args with spaces'"]
```
or
```
args = ["\"my args with spaces\""]
```
to keep it as a single arg

lib/private/strings.bzl
gregmagolan: It would be invoked once per arg per target at analysis time so it could be an issue if there are lots of targets with args in a repo. In the `jq` rule we added an opt-out from this processing:
```
expand_args: Run bazel's location and make variable expansion on the args.
```
=>
```
    if ctx.attr.expand_args:
        args = []
        for a in ctx.attr.args:
            args += split_args(expand_variables(ctx, ctx.expand_location(a, targets = ctx.attr.data), outs = [out]))
    else:
        args = ctx.attr.args
```
So if needed we can add the same opt-out to other rules in the future

lib/private/run_binary.bzl
jbedard: I'm still wondering about this, and maybe just overthinking it 🤔 

Should these all be the exact same?
```
args = ["--one --two --three"]
args = ["--one --two", "--three"]
args = ["--one \"--two\" --three"]
```
🤷

lib/strings.bzl
```diff
@@ -1,7 +1,8 @@
 "Utilities for strings"
 
-load("//lib/private:strings.bzl", _chr = "chr", _hex = "hex", _ord = "ord")
+load("//lib/private:strings.bzl", _chr = "chr", _hex = "hex", _ord = "ord", _split_args = "split_args")
 
 chr = _chr
 ord = _ord
 hex = _hex
+split_args = _split_args
```
jbedard, line 8: Can we avoid making this public for now? 🤷

lib/private/run_binary.bzl
gregmagolan: Yes, the they should all expand to the equivalent of:
```
args = ["--one", "--two", "--three"]
```

lib/strings.bzl
gregmagolan: It may be be useful to downstream rulesets that are have the same issues. I passed the naming by Alex and he was was with it.

jbedard: Lets just be careful when introducing this to other rules 😅

jbedard (approved):

PR #915, opened 2024-08-20
refactor: make root workspace bzlmod only

gregmagolan:
Reduce the maintenance burden caused by duplication of Bazel deps in the root workspace.

CI coverage for WORKSPACE is still in-place in all the e2e workspaces.

---

### Test plan


deps.bzl
```diff
@@ -1,7 +1,7 @@
 """This module contains the project repository dependencies.
 """
 
-load("@bazel_gazelle//:deps.bzl", "go_repository")
```
jbedard, line 4: This means previously we were fetching it twice?

WORKSPACE.bazel
```diff
@@ -0,0 +1,21 @@
+# An external repository for test to use
```
jbedard, line 1: Why is this file still needed at all?

deps.bzl
gregmagolan: Yes, we likely were under bzlmod

WORKSPACE.bazel
kormide: Should this be `WORKSPACE.bzlmod`?

gregmagolan: I chatted with Alex and that and the answer is that you should _only_ use `WORKSPACE.bzlmod` in workspaces that you want to support both bzlmod and non-bzlmod. The go-to is WORKSPACE.bazel if you only support one or the other.

jbedard (approved):

WORKSPACE.bazel
jbedard: I think we should still be able to drop the go deps from here and use pure bzlmod for go. We'd still have to generate `deps.bzl` for non-bzlmod repos (such as e2e tests) to use, but bazel-lib istelf would never use it.

That can be followup later though 🤷

PR #920, opened 2024-08-27
chore: now testing on Aspect Workflows RBE

gregmagolan:


kormide:
We can remove the existing buildbuddy bazelrc configuration as well, and rename our `aspect_rbe` config to just `rbe`.

PR #922, opened 2024-08-31
feat: performance preset: nobuild_runfiles_links

alexeagle:
Noticed this in https://github.com/EngFlow/bazel_invocation_analyzer/issues/107

---

### Test plan
Not sure. Maybe we should try enabling this flag at several client sites and see if it sticks.

### DIFF
--- a/.aspect/bazelrc/performance.bazelrc
+++ b/.aspect/bazelrc/performance.bazelrc
@@ -21,4 +21,4 @@ build --nolegacy_external_runfiles
 
 # Avoid creating the runfiles tree for binaries until it is needed.
 # See https://github.com/bazelbuild/bazel/issues/6627
-build --nobuild_runfiles_links
+build --nobuild_runfile_links
--- a/.aspect/bazelrc/performance.bazelrc
+++ b/.aspect/bazelrc/performance.bazelrc
@@ -19,6 +19,11 @@ build --experimental_reuse_sandbox_directories
 # Docs: https://bazel.build/reference/command-line-reference#flag--legacy_external_runfiles
 build --nolegacy_external_runfiles
 
-# Avoid creating the runfiles tree for binaries until it is needed.
+# Avoid creating a runfiles tree for binaries or tests until it is needed.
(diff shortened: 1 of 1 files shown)


.aspect/bazelrc/performance.bazelrc
```diff
@@ -18,3 +18,7 @@ build --experimental_reuse_sandbox_directories
 # author.
 # Docs: https://bazel.build/reference/command-line-reference#flag--legacy_external_runfiles
 build --nolegacy_external_runfiles
+
+# Avoid creating the runfiles tree for binaries until it is needed.
+# See https://github.com/bazelbuild/bazel/issues/6627
```
gregmagolan, line 23: Add link to docs:
```
Docs: https://bazel.build/reference/command-line-reference#flag--build_runfile_links
```

.aspect/bazelrc/performance.bazelrc
```diff
@@ -18,3 +18,7 @@ build --experimental_reuse_sandbox_directories
...
+build --nobuild_runfile_links
```
gregmagolan, line 24: Seems reasonable as a recommended default, however, I could see it causing issues for some local workflows where a `bazel build` of a binary target is followed by running that target's shell script instead of calling `bazel run` on the binary target. In those cases, the script will need to call `bazel build --build_runfile_links //my/binary:target`.

gregmagolan (approved):

fmeum (approved):

aaomidi:
This flag broke `bazel coverage` invocations as `gcov` ends up being not-runnable.

fmeum:
> This flag broke `bazel coverage` invocations as `gcov` ends up being not-runnable.

Could you file a Bazel bug for this?

alexeagle:
@fmeum sorry I should have been more diligent in #943  - there's already a bug for it: https://github.com/bazelbuild/bazel/issues/20577
looks like a fix is cherry-picked into the next release as well.

FYI @UebelAndre - you may not have known these presets exist :)

PR #931, opened 2024-09-07
perf: Improve copy_file.bzl's progress_message and do some cleanup

dzbarsky:
The change to `progress_message` saves around 100ms of analysis time for my big repo. The rest of the changes are cleanups to reflect that we have long since diverged from skylib's implementation.

### DIFF
--- a/docs/copy_directory.md
+++ b/docs/copy_directory.md
@@ -2,8 +2,7 @@
 
 A rule that copies a directory to another place.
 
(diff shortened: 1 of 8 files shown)


jbedard (approved):

jbedard:
This code removal on its own seems worth it to me, but even more so for the perf win (even if small).

@alexeagle is an extra `../` (sometimes?) showing up in a progress message a big deal?

PR #937, opened 2024-09-14
Stop requiring MANIFEST to deduce runfiles directory

cerisier:
Addresses https://github.com/bazel-contrib/bazel-lib/issues/936

### DIFF
--- a/lib/private/tar.bzl
+++ b/lib/private/tar.bzl
@@ -111,20 +111,6 @@ def _add_compression_args(compress, args):
     if compress == "zstd":
         args.add("--zstd")
 
-def _calculate_runfiles_dir(default_info):
-    manifest = default_info.files_to_run.runfiles_manifest
-
-    # Newer versions of Bazel put the manifest besides the runfiles with the suffix .runfiles_manifest.
-    # For example, the runfiles directory is named my_binary.runfiles then the manifest is beside the
-    # runfiles directory and named my_binary.runfiles_manifest
-    # Older versions of Bazel put the manifest file named MANIFEST in the runfiles directory
-    # See similar logic:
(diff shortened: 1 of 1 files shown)


cerisier:
cc @fmeum

lib/private/tar.bzl
```diff
@@ -224,10 +210,11 @@ def _mtree_impl(ctx):
 
     for s in ctx.attr.srcs:
         default_info = s[DefaultInfo]
-        if not default_info.files_to_run.runfiles_manifest:
+        if not default_info.files_to_run.executable:
             continue
 
-        runfiles_dir = _calculate_runfiles_dir(default_info)
+        # Expect a .runfiles directory for each executable.
+        runfiles_dir = default_info.files_to_run.executable.short_path + ".runfiles"
```
fmeum, line 217: This may need to be prefixed with `ctx.workspace_name`.

fmeum (commented):
Some tests are failing

cerisier:
> Some tests are failing

Tests are failing because it seems that `FilesToRunProvider.executable` is set for each of the `DefaultInfo` of each of the `srcs` even if they are not executable?

Could this be a Bazel bug ?

cerisier:
> Some tests are failing

Tests are failing because it seems that `FilesToRunProvider.executable` is set for existing files in `ctx.attr.srcs`.
Could this be a Bazel bug ?

fmeum:
Hmm, I will take a look. Since source files can't have runfiles, could you use that check to fix the failure?

fmeum:
On second thought, I don't think that this could be fixed. Bazel needs to set `executable` since the source file may be executable but it can't stat it at that point to know whether it is.

PR #939, opened 2024-09-16
fix: add empty files to tar

thesayyn:


alexeagle (approved):
maybe a test/example?

PR #943, opened 2024-09-20
fix(presets): restore build runfile links for the 'coverage' command

alexeagle:
A user reported that this flag broke `coverage` because it makes assumptions about locating tools: https://github.com/bazelbuild/bazel/issues/20577


.aspect/bazelrc/performance.bazelrc
```diff
@@ -27,3 +27,7 @@ build --nolegacy_external_runfiles
 # outside of `bazel run`. In those cases, the script will need to call
 # `bazel build --build_runfile_links //my/binary:target` and then execute the resulting program.
 build --nobuild_runfile_links
+
+# Needed prior to Bazel 7.4, see
```
fmeum, line 31: The fix won't be cherry-picked as it ultimately turned out to be too complicated. It will only land in Bazel 8.

jbedard (approved):

PR #945, opened 2024-09-23
fix: declare host_platform as non-dev dep

jbedard:
Fix https://github.com/bazel-contrib/bazel-lib/issues/944


alexeagle (approved):
thanks Jason

mortenmj:
@alexeagle could we have another release with this change?

PR #947, opened 2024-09-24
feat(presets): java bazelrc options

alexeagle:


.aspect/bazelrc/java.bzl
```diff
@@ -0,0 +1,8 @@
+# Pin java versions
+common --java_language_version=17
+common --java_runtime_version=remotejdk_17
```
gregmagolan, line 3: Also a short intro at the top of the file like we do for other: https://github.com/bazel-contrib/bazel-lib/blob/0db9fbe519b6cc2c7d0f3e66e600914b537202a0/.aspect/bazelrc/javascript.bazelrc#L1

alexeagle: maybe suggested edits? I'm so underwater, won't get to this

.aspect/bazelrc/java.bzl
```diff
@@ -0,0 +1,8 @@
...
+common --tool_java_language_version=17
+common --tool_java_runtime_version=remotejdk_17
```
gregmagolan, line 5: For each flag we also give the user a hand and drop the docs snippet inline and a link to the docs for that flag.
https://github.com/bazel-contrib/bazel-lib/blob/0db9fbe519b6cc2c7d0f3e66e600914b537202a0/.aspect/bazelrc/javascript.bazelrc#L9

.aspect/bazelrc/java.bzl
gregmagolan: low priority. I can take it over when I have a minute.

alexeagle: I got it now :)

gregmagolan (approved):
🌮

PR #948, opened 2024-09-24
fix: unknown repo error with mtree_mutate and Bzlmod

mering:
With Bzlmod, every repo has its own namespace. Using Label() should make sure it uses the namespace of the .bzl file instead of the caller's one.


alexeagle (approved):
good catch, thanks

PR #951, opened 2024-09-27
perf: report unused inputs for the tar rule

plobsing:
The `mtree` spec passed to the `tar` rule very often selects a subset of the inputs made available through the `srcs` attribute. In many cases, these subsets do not break down cleanly along dependency-tree lines and there is no simple way just pass less content to the `tar` rule.

One prominent example where this occurs is when constructing the tars for OCI image layers. For instance when [building a Python-based container image](https://github.com/bazel-contrib/rules_oci/blob/main/docs/python.md), we might want to split the Python interpreter, third-party dependencies, and application code into their own layers. This is done by [filtering the `mtree_spec`](https://github.com/aspect-build/bazel-examples/blob/85cb2aaf8c6e51d5e9e086cc94b94ab896903fb0/oci_python_image/py_layer.bzl#L39).

However, in the operation to construct a `tar` from a subsetted mtree, it is usually still an unsubsetted tree of `srcs` that gets passed. As a result, the subset tarball is considered dependent upon a larger set of sources than is strictly necessary.

This over-scoping runs counter to a very common objective associated with breaking up an image into layers - isolating churn to a smaller slice of the application. Because of the spurious relationships established in Bazel's dependency graph, all tars get rebuilt anytime any content in the application gets changed. Tar rebuilds can even be triggered by changes to files that are completely filtered-out from all layers of the container.

Redundant creation of archive content is usually not too computationally intensive, but the archives can be quite large in some cases, and avoiding a rebuild might free up gigabytes of disk and/or network bandwidth for
better use. In addition, eliminating the spurious dependency edges removes erroneous constraints applied to the build action schedule; these tend to push all Tar-building operations towards the end of a build, even when some archive construction could be scheduled much earlier.

## Risk assessment and mitigation

The `unused_inputs_list` mechanism used to report spurious dependency relationships is a bit difficult to use. Reporting an actually-used input as unused can create difficult to diagnose problems down the line.

However, the behaviour of the `mtree`-based `tar` rule is sufficiently simple and self-contained that I am fairly confident that this rule's used/unused set can be determined accurately in a maintainable fashion.

Out of an abundance of caution I have gated this feature behind a default-off flag. The `tar` rule will continue to operate as it had before - typically over-reporting dependencies - unless the `--@aspect_bazel_lib//lib:tar_compute_unused_inputs` flag is passed.

### Filter accuracy

The `vis` encoding used by the `mtree` format to resiliently handle path names has a small amount of "play" to it - it is reversible but the encoded representation of a string is not
unique. Two unequal encoded strings might decode to the same value; this can happen when at least one of the encoded strings contains unnecessary escapes that are nevertheless honoured by the decoder.

The unused-inputs set is determined using a filter that compares `vis`-encoded strings. In the presence of non-canonically-encoded paths, false-mismatches can lead to falsely reporting that an input is unused.

The only `vis`-encoded path content that is under the control of callers is the `mtree` content itself; all other `vis`-encoded strings are constructed internally to this package, not exposed publicly, and are all derived using the `lib/private/tar.bzl%_vis_encode` function; all of these paths are expected to compare exactly. Additionally, it is expected that many/most users will use this package's helpers (e.g. `mtree_spec`) when crafting their mtree content; such content is also safe. It is only when the user crafts their own mtree, or modifies an mtree spec's `content=` fields' encoding in some way, that a risk of inaccurate reporting arises. The chances for this are expected to be minor since this seems like an inconvenient and not-particularly-useful thing for a user to go out of their way to do.


lib/private/tar.bzl
```diff
@@ -125,6 +127,89 @@ def _calculate_runfiles_dir(default_info):
...
+    path = file.path
+    return _vis_encode(path) + " " + path
+
+def _fmt_keep_inputs_line(file):
+    # The tar.keep_inputs.txt file has a single column of vis-encoded paths of the files to keep.
+    return _vis_encode(file.path)
+
+def _configured_unused_inputs_file(ctx, srcs, keep):
+    """
+    Compute the unused_inputs_list, if configured.
```
alexeagle, line 143: nit: link to Bazel docs on unused_input feature

alexeagle (commented):
Seems generally great to me. @thesayyn want to TAL?

lib/private/tar.bzl
```diff
@@ -84,6 +85,7 @@ _tar_attrs = {
         doc = "Compress the archive file with a supported algorithm.",
         values = _ACCEPTED_COMPRESSION_TYPES,
     ),
+    "_compute_unused_inputs": attr.label(default = Label("//lib:tar_compute_unused_inputs")),
```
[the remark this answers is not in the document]
plobsing: Great idea. Done.

lib/private/tar.bzl
plobsing: Done in the doc for the tri-state attribute recommended by @thesayyn .

lib/BUILD.bazel
```diff
@@ -31,6 +31,11 @@ bool_flag(
     build_setting_default = True if is_bzlmod_enabled() else False,
 )
 
+bool_flag(
+    name = "tar_compute_unused_inputs",
+    build_setting_default = False,
```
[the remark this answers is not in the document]
plobsing: Done.

plobsing:
FYI, after having a bit more time to think about this, I think minor tweaks could make this a lot more exact and a lot less risky. I've iterated on areas that were inexact that were internal to the implementation.

What remains is exactness WRT vis encoding, which is a bigger issue (affects `_mtree_line` too). We can't do full Unicode correctly (because of https://github.com/bazel-contrib/bazel-lib/issues/794 ) but we could probably cover the full 7-bit ASCII range. Combined with a check for non-ASCII that fails the build when computing the unused inputs set (you'd have to disable input pruning if you work with exotically-named files), I think that could make this feature safe enough to become a default (in v3.0). WDYT? I'll leave that to a later change though, this change is large enough as it stands.

lib/private/tar.bzl
```diff
@@ -84,6 +85,40 @@ _tar_attrs = {
...
+comparison performed between `srcs` and `mtree` is currently inexact and may fail to
+handle handwritten or externally-derived mtree specifications. However, it is safe to use
+this feature when the lines found in `mtree` are derived from one or more `mtree_spec`
+rules, filtered and/or merged on whole-line basis only.
+
+Possible values:
+
+    - `compute_unused_inputs = 1`: Always perform unused input discovery and pruning.
+    - `compute_unused_inputs = 0`: Never discover or prune unused inputs.
+    - `stamp = -1`: Discovery and pruning of unused inputs is controlled by the
```
thesayyn, line 115:
```suggestion
    - `compute_unused_inputs = -1`: Discovery and pruning of unused inputs is controlled by the
```

thesayyn (approved):
Excellent!

PR #953, opened 2024-10-01
feat: support bzlmod runfiles lookups

plobsing:
Under bzlmod, repos have aliases in addition to their canonical names;
in order for lookups using these alias names to function properly,
a file name `_repo_mapping` is located in the root of the runfiles tree
and consulted to perform repo-name translation.

See: https://github.com/bazelbuild/proposals/blob/main/designs/2022-07-21-locating-runfiles-with-bzlmod.md

### DIFF
--- a/lib/private/tar.bzl
+++ b/lib/private/tar.bzl
@@ -259,7 +259,7 @@ def _mtree_impl(ctx):
         )
         if repo_mapping != None:
             content.add(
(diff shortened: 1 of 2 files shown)


0xflick:
thank you

alexeagle (approved):
Thanks!

PR #963, opened 2024-10-13
Make build pass with --incompatible_auto_exec_groups

dzbarsky:
I'm not sure about adding those toolchain params to the action helpers, we could make the test rules declare the toolchain dep instead?


alexeagle (commented):
shouldn't we also flip `incompatible_auto_exec_groups` in our bazelrc here to prove it's green (and keep it that way?)

dzbarsky:
> shouldn't we also flip `incompatible_auto_exec_groups` in our bazelrc here to prove it's green (and keep it that way?)

Yep, makes sense, done. I didn't see other flag flips there and wasn't sure what the policy was.

alexeagle (approved):
thanks!

PR #969, opened 2024-10-19
fix: go back to bsdtar-prebuilt 3.7.4 but with a release with a working windows binary

gregmagolan:
Fixes tar extract regression in [2.8.0](https://github.com/bazel-contrib/bazel-lib/releases/tag/v2.9.0) due to bump to bsdtar-prebuilt 3.7.5 in https://github.com/bazel-contrib/bazel-lib/pull/940.

bsdtar-prebuilt 3.7.5 (which depends on libarchive 3.7.5) has a regression that showed up in testing of our internal silo repo when attempting to bump bazel-lib. previously working tests with tar extracts starting to fail with `invalid tar header`. I also tested with a bsdtar-prebuilt 3.7.7 release, built with libarchive 3.7.7 to see if there was a regression in libarchive 3.7.5 but also hit the same failure. Another possibility is BCR patches for libarchive 3.7.5 were copied over from 3.7.4 and they might not be correct.

In any case, to get back to a working version of tar in bazel-lib, this change goes back to safety by switching to a bsdtar-prebuilt 3.7.4 release with a working Windows binary: https://github.com/aspect-build/bsdtar-prebuilt/releases/tag/v3.7.4-4.

### DIFF
--- a/lib/private/tar_toolchain.bzl
+++ b/lib/private/tar_toolchain.bzl
@@ -36,24 +36,24 @@ BSDTAR_PLATFORMS = {
 
 BSDTAR_PREBUILT = {
     "darwin_amd64": (
-        "https://github.com/aspect-build/bsdtar-prebuilt/releases/download/v3.7.5/tar_darwin_amd64",
-        "6c7752869833c5cd4cf9728f90afd37f6665d7f9e06a4fbfdeb0fa41267575cf",
+        "https://github.com/aspect-build/bsdtar-prebuilt/releases/download/v3.7.4-4/tar_darwin_amd64",
+        "a412e5a7b02979cee36dd6b05df62b8596e907c2d9b473786a5d9d5204a0b837",
     ),
     "darwin_arm64": (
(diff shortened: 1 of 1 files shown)


jbedard (approved):

jbedard:
Is there any way to test this here in bazel-lib?

gregmagolan:
> Is there any way to test this here in bazel-lib?

I'm not sure what it is about those particular tests in silo that causes tar extract to fail. Would be good to track it down and make a minimal repro test in the future. For now I tested this change with a patch in silo and verified it fixes the issue.

gregmagolan:
FYSA @alexeagle @dzbarsky

alexeagle:
Isn't this just going to break again the next time we upgrade libarchive?

gregmagolan:
> Isn't this just going to break again the next time we upgrade libarchive?

Likely yes. We have no idea why it was broken on silo. We need a repro outside of silo for this so we can figure out why it was built wrong so we can fix it without having to test in silo.

PR #970, opened 2024-10-21
feat(bzlmod): mark toolchains extension as reproducible

cerisier:
Marking the toolchains extension as reproducible since all underlying tools are already checked for integrity. Saving an entry in the lockfile for that extension.

Marking this has to be conditional to `bazel_features.external_deps.extension_metadata_has_reproducible`, hence bumping `bazel_features` to 1.9.0 which is the version introducing the check (https://github.com/bazel-contrib/bazel_features/releases/tag/v1.9.0)

### DIFF
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -8,6 +8,7 @@ module(
 )
 
 # Lower-bounds (minimum) versions for direct runtime dependencies
+bazel_dep(name = "bazel_features", version = "1.9.0")
 bazel_dep(name = "bazel_skylib", version = "1.5.0")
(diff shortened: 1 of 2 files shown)


alexeagle (approved):
Awesome thanks!

alexeagle:
@cerisier it seems you didn't allow maintainers to push to your PR branch, so I can't rebase this for you to pick up the updated GHA workflow that satisfies our required check. Could you rebase to current HEAD pls?

cerisier:
@alexeagle done !

PR #972, opened 2024-10-25
chore: start testing on Bazel 8

alexeagle:


fmeum:
The CI log output is extremely long. I don't see this with vanilla Bazel, maybe this can be fixed in Aspect CLI?

alexeagle:
I haven't found time to investigate that logging issue but since Bazel 8 is released, I think we need to land this :)

thesayyn (approved):

PR #973, opened 2024-10-25
fix: allow_tags_propagation defaults true in 7

alexeagle:
It was also renamed from experimental_ to incompatible_, though surprisingly the original flag name is still permitted in Bazel 7 without errors.

Rename and default flip was in https://github.com/bazelbuild/bazel/commit/d0625f5b37e7cfe4ecfcff02b15f634c53d7440c


fmeum (approved):

gregmagolan (approved):
🦖

PR #979, opened 2024-11-01
docs: warn about gzip timestamps

thesayyn:
For now we will just warn people to do the right thing. 
See: https://github.com/bazel-contrib/bazel-lib/issues/783


alexeagle (approved):
Better than nothing!

PR #982, opened 2024-11-09
feat: bats_test: add junit report to test results

pjjw:
small addition to the bats test runner to drop junit-formatted output where bazel expects it


thesayyn (approved):

alexeagle:
Oh, this broke our Windows CI. I'll take a stab at fixing it...

PR #983, opened 2024-11-18
docs: Update jq docs to show correct escaping

ChrisChinchilla:
As seen here: https://github.com/bazel-contrib/bazel-lib/blob/31e6c76b7d2e91657913232ec33da6a8b6ace90b/lib/jq.bzl#L87

The docs show the incorrect escaping sequence.

### DIFF
--- a/docs/jq.md
+++ b/docs/jq.md
@@ -91,7 +91,7 @@ jq(
         "--raw-input",
         "--slurp",
     ],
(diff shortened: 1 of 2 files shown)


gregmagolan (approved):

alexeagle:
Per the failing test, this needs to be updated in `jq.bzl` as well. Line 1 of the file you've edited is a hint that it's generated (would be nice if it said where it was generated from)

PR #985, opened 2024-11-22
feat: vis-encode all path characters except visible ASCII

plobsing:
This allows archive input files to include Unicode characters.
It should also support the more esoteric corners of the ASCII set
(e.g. backslashes, newlines, control codes), if the user can find a way
to convince Bazel to use such a file. This latter feature probably isn't
a good idea to actually use; but it is nice to have confidence that this
ruleset won't choke if it comes up.

Generating an archive with non-ASCII characters reports a diagnostic,
however this does not appear to correspond to a real issue — the file
is still placed into the archive with the expected path and contents.

```
INFO: From Tar lib/tests/tar/7.tar:
tar: lib/tests/tar/srcdir/Unicode® support?🤞: Can't translate pathname 'lib/tests/tar/srcdir/Unicode® support?🤞' to UTF-8
```

Encoding of 7-bit ASCII could be performed in Bazel, however Starlark does not provide
access to a string's bytes, only its codepoints, at least until the spec and implementation work
to get a [`bytes` type](https://github.com/bazelbuild/starlark/issues/112) lands.
So a second, out-of-process build-time pass is needed. Since a separate pass
is needed anyways, we move most of the encoding responsibility to this external script
(except cases that it cannot handle without interfering with the mtree file format: newline, space, and backslash).

We use `gawk` for this second pass. A [prior attempt](https://github.com/bazel-contrib/bazel-lib/pull/795)
at this feature had done a similar external pass with the `vis` utility, but has
not been accepted because the `vis` tool is not expected to be available everywhere.
`gawk` is (TODO: will be) provided by a toolchain from this repo.

We also take this opportunity to canonicalize the paths derived
from user-provided mtree specs when computing the unused
inputs set. This should ensure that the comparison is exact,
which substantially reduces the risks associated with this feature
and should enable it to become a default.

Fixes https://github.com/bazel-contrib/bazel-lib/issues/794


plobsing:
FYI, this patch is fairly substantive, but I believe there are separable elements. I've filed it as one PR because this is my desired end-state. The `str.maketrans`/`str.translate` feature could be its own PR if we wanted to expose it as a public API. And just the 7-bit encoding could be a separate, intermediate improvement. Please let me know if you want to land this in phases; either the ones I've outlined or some other partitioning scheme.

lib/private/gen_vis_scripts/BUILD.bazel
```diff
@@ -0,0 +1,39 @@
...
+        "vis_escape_ascii.bzl=$(location vis_escape_ascii.bzl)",
+        "vis_escape_nonascii.sed=$(location vis_escape_nonascii.sed)",
+    ],
+    tool = ":gen_vis_scripts",
+)
+
+write_source_files(
+    name = "write_vis_scripts",
+
+    # Required to support cross-package references.
```
alexeagle, line 30: nit: feels like this target just wants to move to that package instead (likely the `run_binary` too)

lib/private/gen_vis_scripts/gen_vis_scripts.go
```diff
@@ -0,0 +1,171 @@
+// Code generator for vis-encoding support scripts.
```
alexeagle, line 1: I'm a bit scared of having this volume of new code for the existing maintainers to understand and bugfix.

https://github.com/vbatts/go-mtree/tree/main/pkg/govis also has a Go implementation of the `vis(1)` BSD built-in. And @vbatts has indicated willingness to help out with our Bazel tar implementation. Do you think there's any code-sharing possible?

alexeagle (commented):
Awesome, thank you for taking this on. At a high level, the approach seems sound to me.

lib/private/gen_vis_scripts/BUILD.bazel
plobsing: Good point. This setup started entirely within `//lib/private`, and that's where much of it wants to live, but I realized I needed a split to accomodate `rules_go` being a dev dep. I do think though that I maybe put the divide in the wrong place. I've moved most of this back and moved the `gen_vis_scripts` tool to under `//tools`.

alexeagle:
thanks @plobsing  - sorry for the churn

PR #986, opened 2024-11-29
fix: honor umask in write_source_file

sitaktif:
In write_source_file, the output file should honor the current umask when marking files as writable. Not doing so ends up in situations where the file is e.g. created as 664 when initially generated with `write_source_file` but then a git clone/checkout/subtree would create it as 644. This can cause confusion to the user, and might unnecessarily be treated as a file change by some tooling.

From `man 2 chmod`, describing `chmod [who]+[perm]`:

> If no value is supplied for [who], each permission bit specified in
> [perm], for which the corresponding bit in the file mode creation mask
> (see umask(2)) is clear, is set.


gregmagolan (approved):
Thanks for the fix!

PR #989, opened 2024-12-02
fix: Disable the non-deterministic time header when using gzip compression

### LINKED ISSUE #783: [FR]: Tar should add !gzip:timestamp option when compress = gzip

_No response_

### Describe the feature

gzip:timestamp is a source of non-reprocubility; see https://man.freebsd.org/cgi/man.cgi?query=bsdtar&sektion=1&format=html

hofbi:
Maybe a naive approach since I don't know the lib very well. I found https://github.com/bazel-contrib/bazel-lib/blob/main/docs/tar.md#important-note and could not come up with any situation where a user does want to have a non-deterministic tarball.

Replace https://github.com/bazel-contrib/bazel-lib/pull/987

Closes #783

---

I was also thinking about adding an integration test for this such as building a tarball twice and make sure the second one is cached. Open for other suggestions though.

### DIFF
--- a/docs/tar.md
+++ b/docs/tar.md
@@ -223,3 +223,5 @@ tar_lib.mtree_implementation(<a href="#tar_lib.mtree_implementation-ctx">ctx</a>
 | Name  | Description | Default Value |
 | :------------- | :------------- | :------------- |
 | <a id="tar_lib.mtree_implementation-ctx"></a>ctx |  <p align="center"> - </p>   |  none |
+
+
--- a/docs/tar.md
+++ b/docs/tar.md
@@ -17,13 +17,6 @@ We also provide full control for tar'ring binaries including their runfiles.
 The `tar` binary is hermetic and fully statically-linked.
 It is fetched as a toolchain from https://github.com/aspect-build/bsdtar-prebuilt.
 
-## Important Note
-
-When using `compress = "gzip"` its important to disable the non-deterministic time header by providing the `--options=gzip:!timestamp` option.
-
-See: https://datatracker.ietf.org/doc/html/rfc1952#page-5
-See: https://github.com/bazel-contrib/bazel-lib/issues/783
-
 ## Examples
 
 See the [`tar` tests](/lib/tests/tar/BUILD.bazel) for examples of usage.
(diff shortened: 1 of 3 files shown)


lib/private/tar.bzl
```diff
@@ -133,6 +133,7 @@ def _add_compression_args(compress, args):
         args.add("--compress")
     if compress == "gzip":
         args.add("--gzip")
+        args.add("--options=gzip:!timestamp")
```
fmeum, line 136: Could you also remove the note from the docs about having to set this manually?

fmeum (commented):
@alexeagle Any objections?

alexeagle:
LGTM thanks! FYI @thesayyn who edited it recently too

docs/tar.md
```diff
@@ -17,13 +17,6 @@ We also provide full control for tar'ring binaries including their runfiles.
 The `tar` binary is hermetic and fully statically-linked.
 It is fetched as a toolchain from https://github.com/aspect-build/bsdtar-prebuilt.
 
-## Important Note
```
alexeagle, line 20: See line 1 - this file is generated so you'll need the corresponding change in the .bzl file it's generated from.

hofbi: Already done with https://github.com/bazel-contrib/bazel-lib/pull/989/commits/237bfeac2623dc7449ae1f0c3abad9ad8c77d45c

thesayyn:
So reason why i did not do this myself in the first place was to not take away the freedom of the user to choose. I'd say that we gate this behind an attribute and add the sensible defaults.  I can imagine something like 

(Naming is to be determined)

```
default_options = False | True
```

hofbi:
Is there any situation where a user might want to have a non-deterministic tarball?

thesayyn:
I can't think of any trivial cases where it might be useful under Bazel, but bsdtar already supports toggling it and we should do the same.

hofbi:
Not sure if I understand what you propose. Do you want a boolen flag to enable/disable a set of default options which currently is just this single option?

fmeum:
> I can't think of any trivial cases where it might be useful under Bazel, but bsdtar already supports toggling it and we should do the same.

If we can't think of a use case and our aim is to provide a flexible but hermetic tar, how about we don't make this configurable yet and wait for someone to ask for it first? I'm worried about increasing the complexity budget for what may very well never be needed.

alexeagle:
I agree, while our principle is to have a thin layering on top of the tools, any non-determinism in those tools is a bug from Bazel's perspective, and so I can't imagine any use case of a user actually wanting to control this particular option.

PR #994, opened 2024-12-06
chore(deps): Enable renovate for pre-commit

hofbi:
Renovate does not upgrade pre-commit hooks by default. I think, we should also keep them up to date.

Just enabling it will create a separate PR per hook. In case this becomes too noisy, we can also configure renovate to create a single PR updating all in one.


gregmagolan (approved):
🌮

PR #1002, opened 2024-12-08
ci(pre-commit): Add pre-commit CI job

hofbi:
After enabling renovate updates for pre-commit in #994, we should also make sure pre-commit works in CI.

I also proposed this for https://github.com/bazel-contrib/rules-template/pull/130.


tools/lint/vale/config/vocabularies/engineering/accept.txt
```diff
@@ -62,7 +62,7 @@ rebase
 mutex
 hotfix
 oncall
-poisioning
```
fmeum, line 65: Is this a spell checking lint spell checking a spell checking lint's test files? I feel like we have gone too far here...

@alexeagle Do we need all this spell checking setup?

hofbi: This word was an autofix done by the `typos` hook. I was not sure why this word was added here to the vocabulary. Should we rather exclude the vocabularies from the typos hook?

hofbi: Since we expect the list to in `accept.txt` to shrink, is this something that should happen before this PR or can we move on with this?

alexeagle: Sure, I'll reduce the unneeded bits in a follow-up (maybe)

alexeagle (approved):

PR #1019, opened 2024-12-19
fix(tar): allow --action_env to override LC_ALL

alexeagle:
`bsdtar` will fail to extract archives with unicode characters in a filename, on systems where the default locale is `C` or some non-UTF value.

Users encounter this as strange extract failures, but only on a subset of the systems they work on; this is non-hermetic behavior.

Workaround https://github.com/bazelbuild/bazel-central-registry/issues/2256

### DIFF
--- a/lib/private/tar.bzl
+++ b/lib/private/tar.bzl
@@ -333,7 +333,7 @@ def _tar_impl(ctx):
         mnemonic = "Tar",
         unused_inputs_list = unused_inputs_file,
         toolchain = "@aspect_bazel_lib//lib:tar_toolchain_type",
-	# Allow users to set --action_env=LC_ALL=C.UTF-8 for example,
+        # Allow users to set --action_env=LC_ALL=C.UTF-8 for example,
         # on systems where default locale is wrong.
         # See https://github.com/bazelbuild/bazel-central-registry/issues/2256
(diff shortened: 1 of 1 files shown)


fmeum:
I think that the default locale with no env vars set is always C on Linux and that's what happens in the sandbox, so I expect this problem to be rather common
 For Bazel's Java rules, I've been setting `LC_CTYPE` explicitly: https://github.com/bazelbuild/rules_java/commit/c22454fadb4773cbd202bfa3e28f1d6a88c4c94a

fmeum:
Since `--action_env` requires explicit configuration and makes it more easy to screw up cache performance, I would prefer a more explicit fix if we can find one.

alexeagle:
I'm not certain we should merge this; created it mostly as the alternative to discuss. Feels to me that we should have global guidance across the whole Bazel ecosystem about whether system locale is a hermeticity failure, and we should be forcing a locale globally.

See https://github.com/google/copybara/blob/master/.bazelrc#L3-L10 as one example I've noticed

alexeagle:
Yeah your Java example is like what I did with https://github.com/aspect-build/rules_js/pull/2044 - the hard thing for me is how hopeless the project of finding every action that needs an `env` set

PR #1030, opened 2025-01-15
feat: Add diff_test_failure_message attribute to write_source_file*

jameskuszmaul-brt:
This makes it so that developers can add a custom error message to the diff test failure messages when using `write_source_file` and `write_source_files`.

### DIFF
--- a/lib/private/write_source_file.bzl
+++ b/lib/private/write_source_file.bzl
@@ -20,6 +20,8 @@ def write_source_file(
         additional_update_targets = [],
         suggested_update_target = None,
         diff_test = True,
+        diff_test_failure_message = "{{DEFAULT_MESSAGE}}",
+        file_missing_failure_message = "{{DEFAULT_MESSAGE}}",
         check_that_out_file_exists = True,
         **kwargs):
     """Write a file or directory to the source tree.
@@ -47,6 +49,22 @@ def write_source_file(
         suggested_update_target: Label of the `write_source_files` or `write_source_file` target to suggest running when files are out of date.
 
         diff_test: Test that the source tree file or directory exist and is up to date.
+
+        diff_test_failure_message: Text to print when the diff test fails, with templating options for
+            relevant targets.
+
+            Substitutions are performed on the failure message, with the following substitutions being available:
+
+            `{{DEFAULT_MESSAGE}}`: Prints the default error message, listing the target(s) that
+              may be run to update the file(s).
+
+            `{{TARGET}}`: The target to update the individual file that does not match in the
+              diff test.
+
+            `{{SUGGESTED_UPDATE_TARGET}}`: The suggested_update_target if specified.
+
+        file_missing_failure_message: Text to print when the output file is missing. Subject to the same
+             substitutions as diff_test_failure_message.
 
         check_that_out_file_exists: Test that the output file exists and print a helpful error message if it doesn't.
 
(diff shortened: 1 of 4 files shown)


jameskuszmaul-brt:
Note: I would be happy to get either more automated tests added or change the name of the attribute; however, I didn't see any preexisting tests for these failure messages and so I didn't want to go around refactoring/messing with things without knowing where to look, and if you all don't consider it necessary to add automated tests for this then I didn't want to spend the time figuring it out.

kormide:
IMO appending to the end seems like a very specific use case. A more generally useful solution would be to replace the whole message while supporting templating for the target and suggested target to be inserted.

jameskuszmaul-brt:
@kormide Do you think such templating should just go off of the target name or off of the whole pre-provided string?

Also, is there any standard pattern for what syntax you'd use for this sort of templating? I'd probably default to just implenting templating for the two target names and make it so that if you wanted to reproduce the existing error message you would do something like:

```
If you want to update all the files, do bazel run %all-targets%; if you want to update just this target, run %target%.
```

Suggestions on names & syntax welcome; I'm not inclined to implement any larger a feature set than described here.

kormide:
> @kormide Do you think such templating should just go off of the target name or off of the whole pre-provided string?
> 
> Also, is there any standard pattern for what syntax you'd use for this sort of templating? I'd probably default to just implenting templating for the two target names and make it so that if you wanted to reproduce the existing error message you would do something like:
> 
> ```
> If you want to update all the files, do bazel run %all-targets%; if you want to update just this target, run %target%.
> ```
> 
> Suggestions on names & syntax welcome; I'm not inclined to implement any larger a feature set than described here.

I think the `{{VAR_NAME}}` syntax would be suitable since we use that elsewhere in `expand_template`. We could have three variables: `TARGET`, `SUGGESTED_UPDATE_TARGET`, and `ORIGINAL_MESSAGE`, the latter containing the full message. Then it would be trivial for you to append something after that.

jameskuszmaul-brt:
> I think the `{{VAR_NAME}}` syntax would be suitable since we use that elsewhere in `expand_template`. We could have three variables: `TARGET`, `SUGGESTED_UPDATE_TARGET`, and `ORIGINAL_MESSAGE`, the latter containing the full message. Then it would be trivial for you to append something after that.

Sounds good; I'm not sure if I'll implement `ORIGINAL_MESSAGE` or not (depends on how much complexity it seems like it adds).

jameskuszmaul-brt:
Need to fix some of the formatting & text in the docs, but does this look like what we want?

lib/private/write_source_file.bzl
```diff
@@ -394,3 +415,16 @@ def _is_file_missing(label):
...
+
+     Replaces the {{DEFAULT_MESSAGE}}, {{TARGET}}, and {{SUGGESTED_UPDATE_TARGET}} strings in
+     message with the corresponding arguments.
+
+     Args:
+         message: The user-provided message to do template replacement on.
+         default_message: The message to fill in for the {{DEFAULT_MESSAGE}} parameter.
+         target: The string to fill in for the {{TARGET}} parameter.
+         suggested_update_target: The string to fill in for the {{SUGGESTED_UPDATE_TARGET}} parameter."""
+    return message.replace("{{DEFAULT_MESSAGE}}", default_message).replace("{{TARGET}}", target).replace("{{SUGGESTED_UPDATE_TARGET}}", suggested_update_target if suggested_update_target else "")
```
jameskuszmaul-brt, line 430: I could be convinced to make `SUGGESTED_UPDATE_TARGET` default to `TARGET` when not available, but not sure what the intuitive behavior is.

lib/private/write_source_file.bzl
```diff
@@ -20,6 +20,7 @@ def write_source_file(
         additional_update_targets = [],
         suggested_update_target = None,
         diff_test = True,
+        diff_test_failure_message = "{{DEFAULT_MESSAGE}}",
```
kormide, line 23: If someone sets `diff_test_failure_message` then that will override both the not-exists and not-up-to-date messages, right? I think we may need a second property so that those can remain different.

kormide (changes requested):
Looks good. I think we just need a second property.

kormide (approved):

jameskuszmaul-brt:
@kormide I had to manually tweak the generated markdown to make the CI happy; FYI I seem to be running into the same thing as https://github.com/bazel-contrib/bazel-lib/pull/1032#issue-2794711472 alludes to where the markdown generation is not reproducible across different environments.

PR #1032, opened 2025-01-17
feat: Add option to pass args to `diff` command in `diff_test` on Linux/MacOS

r0bobo:
This makes it possible to pass extra arguments to the diff command that is run in a `diff_test()`.

~When running `bazel run @@//docs:update` I get a diff in all documents to add example `load()` statements. This was added in [stardoc 0.7.0](https://github.com/bazelbuild/stardoc/releases/tag/0.7.0). I'm not sure how it is leaking in to my local environment. :thinking:~

### DIFF
--- a/docs/diff_test.md
+++ b/docs/diff_test.md
@@ -15,7 +15,7 @@ See also: [rules_diff](https://gitlab.arm.com/bazel/rules_diff)
 ## diff_test
 
 <pre>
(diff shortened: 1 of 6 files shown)


alexeagle:
I hit that problem today - the workaround was `BAZELISK_BASE_URL= USE_BAZEL_VERSION=7.x bazel run docs:update`

r0bobo:
> I hit that problem today - the workaround was `BAZELISK_BASE_URL= USE_BAZEL_VERSION=7.x bazel run docs:update`

Thanks! That was it.

alexeagle (approved):

PR #1033, opened 2025-01-17
feat: Add option to control verbosity of `write_source_files`.

r0bobo:
This adds a verbosity option to `write_source_files` that makes it possible to control what the output message looks like when running the targets that copies the files to the source tree. It has the options `full` which is the same as the current one and the default value, `short` which is a shorter version and `quiet` which doesn't output anything at all.

When updating the docs I have a bigger diff than I expect to just as I explained in https://github.com/bazel-contrib/bazel-lib/pull/1032#issue-2794711472.


lib/private/write_source_file.bzl
```diff
@@ -71,6 +72,8 @@ def write_source_file(
             If `True`, the output file or directory must be in the same containing Bazel package as the target since the underlying mechanism
             for this check is limited to files in the same Bazel package.
 
+        verbosity: Verbosity of message being when the copy target is run. One of `full`, `brief`, `quiet`.
```
kormide, line 75: `brief` should be `short` here.

kormide (changes requested):

lib/private/write_source_file.bzl
r0bobo: Thanks. Fixed.

kormide (approved):

PR #1036, opened 2025-01-20
feat: add Support for Symlinks in tar Rule's Runfiles Handling

ewianda:
This Pull Request (PR) enhances the `tar` rule to better handle symlinks within runfiles, improving compatibility with various language ecosystems

Currently, symlinks within runfiles are not correctly reflected in the `mtree_spec` generated by Bazel, causing issues in packaging, especially for symlink-sensitive applications like Node.js with `pnpm`.

---

### Symlink Detection and Resolution
- **Logic Added**: Detect if a file in the runfiles is a symlink using `readlink`.
- **Runtime Resolution**: Resolves symlinks post-build to ensure the `mtree_spec` accurately represents symlink structures.

### MTree Mutation
- **Modification Script**: Utilized `awk` scripting to adjust the `mtree_spec` after it is generated.
  - Reads each line of the `mtree_spec`.
  - Verifies if entries reported as files are actually symlinks using `readlink -f`.
  - Updates entries to show `type=link` and the correct `link=` path when applicable.

closes #609


tools/mtree/main.go
```diff
@@ -0,0 +1,194 @@
...
+	"flag"
+	"fmt"
+	"os"
+	"path/filepath"
+	"regexp"
+	"strings"
+)
+
+func main() {
+	// Define command-line flags
```
thesayyn, line 14: Hmm, having yet another go program is something i'd like to avoid if possible.

thesayyn: what was the reason that awk did not work here?  I have used it extensively to mutate mtrees for couple of clients, and since `system` call is there, i don't see why it can't be used.

ewianda: > what was the reason that awk did not work here? I have used it extensively to mutate mtrees for couple of clients, and since `system` call is there, i don't see why it can't be used.

technically no. Will put what I was able to come up with

ewianda: @thesayyn change the go-script to awk-script. TAL

thesayyn: Looking great. Thank you :)

lib/private/modify_mtree.awk
```diff
@@ -1,4 +1,25 @@
 # Edits mtree files. See the modify_mtree macro in /lib/tar.bzl.
+function create_map(filename) {
```
thesayyn, line 2: couple high level things here could be nicer; 

Right now we are reading the mtree twice and manually parsing it, which can be delegated to awk if we change this a little. 

1- Introduce a feature flag to mutate_mtree, `preserve_symlinks = True | False`, 
2- Remove the read and parse logic from `BEGIN`
3- In the main block, which runs per line, continue operating like we did before, so keep it same as before
4- In the main block, If `preserve_symlinks = True`,  do not `print;` so we can post process all the lines
5- In the main block, If `preserve_symlinks = True`,  fill an array with information you need. maybe `map[$1] = [$0, $<index for contents>]`?
6- Create an  `END {}` block that only runs if `preserve_symlinks = True`, and iterates over the map it previously filled and checks for symlinks and adjusts the line if needed.

thesayyn:
Looking great already.

lib/private/tar.bzl
```diff
@@ -446,6 +446,80 @@ def _mtree_impl(ctx):
 
     return DefaultInfo(files = depset([out]), runfiles = ctx.runfiles([out]))
 
+def _mtree_mutate_impl(ctx):
+    srcs_runfiles = [
+        src[DefaultInfo].default_runfiles.files
+        for src in ctx.attr.srcs
+    ]
+
+    out_mtree = ctx.outputs.out
```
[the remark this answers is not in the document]
ewianda: Actually it doesn't work with binary rule. Adding `srcs` to `genrule(..., srcs)` only adds the executable in `bazel-bin` and not the entire runfiles tree. So the symlinks are not resolved.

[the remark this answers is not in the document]
thesayyn: Hmm, that's weird, perhaps adding srcs to `genrule#tools` array will cause the runfiles tree to be built?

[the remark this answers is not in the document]
ewianda: Troed that, but the runfiles end up in  `bazel-out/k8-fastbuild-ST-aa0d9d72ef61/` for example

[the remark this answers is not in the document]
thesayyn: I see, thanks for checking :)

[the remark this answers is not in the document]
ewianda: Should we revert back to implementing a rule for this then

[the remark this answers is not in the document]
thesayyn: If it works with genrule, then i prefer genrule, but a rule is a must then yeah it's okay to go back to rule.

ewianda:
> You have some buildifier errors.

Should be all good now

ewianda:
Hi @thesayyn I am noticing two potential issues
* For relatively large application for example our next build with about a 150K lines of `type=file`, it takes about 5 minutes to do `readlink`
* Some symlinks are relative, and because of the way the srcs files are laid out, `readlink -f` fails and `readlink` works. This implies that if we have to handle both cases in theory that will in theory increase the build time for resolving symlinks

I have thought about a few things
1. Us the file.is_symlink feature only available in bazel 8 to filter out symlink files making the preserve_symlink feature a bazel 8+ feature
 2. Have a separate action for evaluating symlink using bash to parallelize the process, something like `cat mtree.spec | xargs -P`

thesayyn:
I like the numbers, yes, i was also curious of performance implications for something like this.  5mins for 150k is unpleasant.

> Some symlinks are relative, and because of the way the srcs files are laid out, readlink -f fails and readlink works. This implies that if we have to handle both cases in theory that will in theory increase the build time for resolving symlinks

Oh, yes, there must be more to the logic that what we currently have here; See: https://github.com/aspect-build/rules_js/blob/main/js/private/image/index.ts#L84   it's been in use for a while and covers some edge cases. 

> Us the file.is_symlink feature only available in bazel 8 to filter out symlink files making the preserve_symlink feature a bazel 8+ feature

That is possible, yes, but i don't it will report correctly for things declared as file but is symlink, eg treeartifact containing a symlink. there is no way to know about symlinks inside treeartifacts.

> Have a separate action for evaluating symlink using bash to parallelize the process

I have little to no faith that it would be faster than what we right now, since `which readlink` is `/usr/bin/readlink` so it will likely call out to external binary and suffer from the same problem as awk.


My current thinking is that, while writing a go tool for this is convenient, its probably overkill, and will fall short if people need to customize it, i'd like to use that effort to make a proper awk setup. 

https://github.com/thesayyn/prebuilt_toolchains has an AWK recipe for Bazel, i am planning to make a toolchain for it. 

AWK has this concept of extensions https://www.gnu.org/software/gawk/manual/html_node/gawkextlib.html which we can use to introduce a new readlink function to get near native performance within awk, so instead of `system` function, we'd use `readlink` function which will just make a syscall within the same process.

References: 

https://sourceforge.net/projects/gawkextlib/
https://www.gnu.org/software/gawk/manual/html_node/gawkextlib.html

thesayyn:
All that said, i don't mind landing this as is, since its behind a flag, (which we should mark as experimental BTW), and get some feedback and interest before investing more into this.

ewianda:
Fair enough, 
I will like to add the fallback of `readlink -f` to `readlink` to address the issue that I had with rules_python

thesayyn:
No need for the experimental prefix, we can just put a `EXPERIMENTAL! We may remove or change it at any point without further notice.` in the `doc` of the attribute.

ewianda:
hi @thesayyn `//lib/tests/tar:test_17_after_processing` is failing for bazel `7.4.1` and I can't seem to be able to reproduce it locally

lib/tests/tar/BUILD.bazel
```diff
@@ -503,3 +505,109 @@ assert_tar_listing(
...
+        "drwxr-xr-x  0 0      0           0 Jan  1  2023 lib/tests/tar/native_binary_bin.runfiles/_main/lib/tests/tar/node_modules/.pnpm/a@0.0.0/node_modules/a",
+        "-rwxr-xr-x  0 0      0           0 Jan  1  2023 lib/tests/tar/native_binary_bin.runfiles/_main/lib/tests/tar/node_modules/.pnpm/a@0.0.0/node_modules/a/package.json",
+        "drwxr-xr-x  0 0      0           0 Jan  1  2023 lib/tests/tar/native_binary_bin.runfiles/_main/lib/tests/tar/node_modules/.pnpm/b@0.0.0",
+        "drwxr-xr-x  0 0      0           0 Jan  1  2023 lib/tests/tar/native_binary_bin.runfiles/_main/lib/tests/tar/node_modules/.pnpm/b@0.0.0/node_modules",
+        "drwxr-xr-x  0 0      0           0 Jan  1  2023 lib/tests/tar/native_binary_bin.runfiles/_main/lib/tests/tar/node_modules/.pnpm/b@0.0.0/node_modules/a",
+        "l---------  0 0      0           0 Dec 31  1969 lib/tests/tar/native_binary_bin.runfiles/_main/lib/tests/tar/node_modules/.pnpm/b@0.0.0/node_modules/a/package.json -> lib/tests/tar/native_binary_bin.runfiles/_main/lib/tests/tar/node_modules/.pnpm/a@0.0.0/node_modules/a/package.json",
```
thesayyn, line 603: using absolute symlinks is something i'd like to avoid since it makes relocation of the symlinks harder, we might want to make symlinks relative to allow easy manipulation of the mtree by another tool. 

Here how js_image_layer does it; https://github.com/aspect-build/rules_js/blob/f99b0703d7dbb81e6ff6db4ea6c68797853d8b6a/js/private/test/image/custom_owner_test_node_modules.bzlmod.listing#L14

ewianda:
@thesayyn added `readpath` to get the relative path. From

>AWK has this concept of extensions https://www.gnu.org/software/gawk/manual/html_node/gawkextlib.html which we can use to introduce a new readlink function to get near native performance within awk, so instead of system function, we'd use readlink function which will just make a syscall within the same process.

Will be possible to replace `readpath` as well

lib/tests/tar/BUILD.bazel
```diff
@@ -503,3 +505,109 @@ assert_tar_listing(
...
+        "drwxr-xr-x  0 0      0           0 Jan  1  2023 lib/tests/tar/native_binary_bin.runfiles/_main/lib/tests/tar/node_modules/.pnpm/a@0.0.0/node_modules/a",
+        "-rwxr-xr-x  0 0      0           0 Jan  1  2023 lib/tests/tar/native_binary_bin.runfiles/_main/lib/tests/tar/node_modules/.pnpm/a@0.0.0/node_modules/a/package.json",
+        "drwxr-xr-x  0 0      0           0 Jan  1  2023 lib/tests/tar/native_binary_bin.runfiles/_main/lib/tests/tar/node_modules/.pnpm/b@0.0.0",
+        "drwxr-xr-x  0 0      0           0 Jan  1  2023 lib/tests/tar/native_binary_bin.runfiles/_main/lib/tests/tar/node_modules/.pnpm/b@0.0.0/node_modules",
+        "drwxr-xr-x  0 0      0           0 Jan  1  2023 lib/tests/tar/native_binary_bin.runfiles/_main/lib/tests/tar/node_modules/.pnpm/b@0.0.0/node_modules/a",
+        "l---------  0 0      0           0 Dec 31  1969 lib/tests/tar/native_binary_bin.runfiles/_main/lib/tests/tar/node_modules/.pnpm/b@0.0.0/node_modules/a/package.json -> ../../../../../../../../../../../../lib/tests/tar/native_binary_bin.runfiles/_main/lib/tests/tar/node_modules/.pnpm/a@0.0.0/node_modules/a/package.json",
```
thesayyn, line 603: This doesn't look right, there should be only four `../` at most looking at these paths, what happens here is that realpath resolves symlinks all the way out of the sandbox.

ewianda:
@thesayyn made the suggested change

thesayyn (approved):

lib/private/modify_mtree.awk
```diff
@@ -1,4 +1,46 @@
...
+    for (i = 1; i <= min_length; i++) {
+        if (segments1[i] != segments2[i]) {
+            break
+        }
+        common_path = (common_path == "" ? segments1[i] : common_path "/" segments1[i])
+    }
+
+    return common_path
+}
+function make_relative_link(path1, path2, i, common, target, relative_path, back_steps) {
```
thesayyn, line 25: nit:  would be great if you could drop this link here for future reference. https://github.com/bazelbuild/bazel-skylib/blob/7209de9148e98dc20425cf83747613f23d40827b/lib/paths.bzl#L217

thesayyn:
Hmm, the test is failing due to mtime,  https://github.com/bazel-contrib/bazel-lib/actions/runs/13054891062/job/36493038666?pr=1036#step:6:252


Can you set it to a static value in your tests?

ewianda:
> Hmm, the test is failing due to mtime, https://github.com/bazel-contrib/bazel-lib/actions/runs/13054891062/job/36493038666?pr=1036#step:6:252
> 
> Can you set it to a static value in your tests?

Just fixed it

thesayyn (approved):

PR #1052, opened 2025-02-27
fix: Handle when readlink -f goes out of BIN-DIR

ewianda:
Noticed that on MacOS readlink -f goes out the sandbox in some cases

### DIFF
--- a/lib/private/modify_mtree.awk
+++ b/lib/private/modify_mtree.awk
@@ -117,8 +117,9 @@ function make_relative_link(path1, path2, i, common, target, relative_path, back
 	    cmd = "readlink -f \"" path "\""
 	    cmd | getline resolved_path
 	    close(cmd)
-	    # If readlink -f fails use readlink for relative links
-	    if (resolved_path == "") {
+	    # 1. If readlink -f fails use readlink for relative links
+	    # 2. Make sure readlink -f doesn't escape BIN_DIR
+	    if (resolved_path == "" || !(resolved_path ~ bin_dir)) {
 		cmd = "readlink \"" path "\""
 		cmd | getline resolved_path
 		close(cmd)


ewianda:
@thesayyn Notice that the preserve symlink fails on MacOs for  rules-python. Not sure how to repro this in CI.

lib/private/modify_mtree.awk
```diff
@@ -117,8 +117,9 @@ function make_relative_link(path1, path2, i, common, target, relative_path, back
 	    cmd = "readlink -f \"" path "\""
 	    cmd | getline resolved_path
 	    close(cmd)
-	    # If readlink -f fails use readlink for relative links
-	    if (resolved_path == "") {
+	    # 1. If readlink -f fails use readlink for relative links
+	    # 2. Make sure readlink -f doesn't escape BIN_DIR
+	    if (resolved_path == "" || !(resolved_path ~ bin_dir)) {
```
thesayyn, line 122: I believe this is wrong, we need to handle running out of bindir cases for two reasons; 

- Execution strategies on macos use symlinks, there readlink might go out of sandbox, so we need to find things relative to execroot
- When running the action with no sandbox. eg `tags = ["local"]`

ewianda: Sorry,  is it the checking logic that is wrong, or the reasons stated for the check wrong

thesayyn: Its insufficent is the problem. This reference implementation that correctly detects if a file is a symlink under most circumstances 
https://github.com/aspect-build/rules_js/blob/1c1823d605578d7266fad8b5410d0aae6b9bfc6d/js/private/js_image_layer.mjs#L55-L90

thesayyn: Following on this, so the right logic for this would be, we follow the symlink only one level, check if its out of the bindir, and if it is then its not a symlink that stays within the output tree therefore an implementation detail symlink.

PR #1053, opened 2025-03-04
fix: expose LC_ALL variable in tar toolchain

alexeagle:
Replaces #1019 
Fixes https://github.com/bazelbuild/bazel-central-registry/issues/2256 (provided that callers such as rule implementations use this new property)

### DIFF
--- a/lib/private/tar_toolchain.bzl
+++ b/lib/private/tar_toolchain.bzl
@@ -1,5 +1,7 @@
 "Provide access to a BSD tar"
 
+load(":utf8_environment.bzl", "Utf8EnvironmentInfo")
+
 BSDTAR_PLATFORMS = {
     "darwin_amd64": struct(
         compatible_with = [
@@ -87,7 +89,10 @@ bsdtar_binary_repo = repository_rule(
 TarInfo = provider(
     doc = "Provide info for executing BSD tar",
     fields = {
-        "binary": "bsdtar executable",
+        # environment appears on the toolchain since it's platform-specific, and we want to pair it with the tool.
+        # See https://github.com/aspect-build/rules_js/issues/2039
+        "default_env": "environment variables which should be set when spawning tar, to ensure reproducible results",
+        "binary": "tar executable",
     },
 )
 
@@ -104,6 +109,7 @@ def _tar_toolchain_impl(ctx):
         files = depset(ctx.files.binary + ctx.files.files),
     )
     tarinfo = TarInfo(
+        default_env = ctx.attr._utf8_environment[Utf8EnvironmentInfo].environment,
         binary = binary,
     )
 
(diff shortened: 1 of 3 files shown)


lib/private/tar_toolchain.bzl
```diff
@@ -87,7 +89,11 @@ bsdtar_binary_repo = repository_rule(
 TarInfo = provider(
     doc = "Provide info for executing BSD tar",
     fields = {
-        "binary": "bsdtar executable",
+        # environment appears on the toolchain since it's platform-specific, and we want to pair it with the tool.
+        # Currently known to be required only for the extract mode; see
+        # https://github.com/bazelbuild/bazel-central-registry/issues/2256
+        "default_env": "environment variables which should be set when spawning tar, to ensure reproducible results",
+        "binary": "tar executable",
```
jbedard, line 96: Why change from "bsdtar" to "tar"?

lib/private/utf8_environment.bzl
```diff
@@ -0,0 +1,60 @@
+# Vendored from
```
jbedard, line 1: Is there an issue we can reference stating why it was vendored?

lib/private/tar_toolchain.bzl
alexeagle: just a matter of taste, as I was writing the comment above it - the user runs `tar, really there's "BSD tar" and "libarchive"
the term "bsdtar" is our own invention

lib/private/utf8_environment.bzl
alexeagle: Oh I thought it was obvious: we don't want to introduce a dependency from bazel-lib to rules_java

jbedard (approved):

lib/private/utf8_environment.bzl
```diff
@@ -0,0 +1,60 @@
...
+# > and the behavior of character classes.
+#
+# However libarchive only looks for LC_ALL, otherwise it spams warnings like "tar: Failed to set default locale"
+# https://github.com/libarchive/libarchive/blob/65196fdd1a385f22114f245a9002ee8dc899f2c4/tar/bsdtar.c#L192
+# Docs for LC_ALL:
+# > Overrides the value of the LANG environment variable and the values of any other LC_* environment variables.
+_LOCALE_VAR = "LC_ALL"
+
+# The default UTF-8 locale on all recent Linux distributions. It is also available in Cygwin and
+# MSYS2, but doesn't matter for determining the JVM's platform encoding on Windows, which always
```
thesayyn, line 43:
```suggestion
# MSYS2, but doesn't matter for determining the bsdtar encoding on Windows, which always
```

?

thesayyn (approved):

PR #1056, opened 2025-03-06
diff_test: adding a hermetic toolchain

manuelnaranjo:
Trying to add a diff hermetic toolchain based on
https://github.com/uutils/diffutils


alexeagle (approved):
Nice! Thank you!

alexeagle:
I got stuck trying to decide if https://gitlab.arm.com/bazel/rules_diff fully covers this use case or not.

manuelnaranjo:
> I got stuck trying to decide if https://gitlab.arm.com/bazel/rules_diff fully covers this use case or not.

When I did the contribution I think I tried to research the differences and can't remember why I didn't go that way, most likely it's because ir wasn't in the BCR yet. Also I'm not 100% sure you want bazel-lib to depend on it, maybe it's best to vendor the binaries once again

PR #1059, opened 2025-03-20
feat(mtree_mutate): support file_mode argument

### LINKED ISSUE #921: [FR]: mtree_mutate should support changing mode

Currently, there is no way to set mode for files via mtree_mutate, or preserve the original mode.

### Describe the feature

```
# preserves the mode from source file
mtree_mutate(mode = "preserve") 
# sets the mode for all files to 0777
mtree_mutate(mode = "0777")
```

skeggse:
fixes bazel-contrib/tar.bzl#18

### DIFF
--- a/lib/private/tar.bzl
+++ b/lib/private/tar.bzl
@@ -163,6 +163,9 @@ _mutate_mtree_attrs = {
     "mtime": attr.string(
         doc = "Specifies the modification time (mtime) to be applied to all files in the tar file. Used for deterministic builds.",
     ),
+    "file_mode": attr.string(
+        doc = "Specifies the file mode to be applied to all the files in the tar file.",
+    ),
     "owner": attr.string(
         doc = "Specifies the numeric user ID (UID) for the owner of the files in the tar archive.",
     ),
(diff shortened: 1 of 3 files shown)


skeggse:
Tested locally using
```starlark
    http_archive(
        name = "aspect_bazel_lib",
        integrity = "sha256-xi9efFcuNYR8c50Ji0+yj1cxy5kW6c+sQBDrDFMKuKM=",
        strip_prefix = "bazel-lib-3e312b35178fae5de0cad2a1fadcd03d9819b3b1",
        url = "https://github.com/skeggse/bazel-lib/archive/3e312b35178fae5de0cad2a1fadcd03d9819b3b1.tar.gz",
    )
```

seems to work

lib/private/tar.bzl
```diff
@@ -153,6 +153,9 @@ _mutate_mtree_attrs = {
     "mtime": attr.string(
         doc = "Specifies the modification time (mtime) to be applied to all files in the tar file. Used for deterministic builds.",
     ),
+    "file_mode": attr.string(
```
thesayyn, line 156: name "file_mode" is confusing since this applies to everything, of any `type`.  Should this be just mode?

thesayyn: No, mtree contains everything, intermediary directories, files, symlinks, fifo etc. in this case, this variable will apply to all directories and files.

skeggse: i'm still not sure i understand. the line of code i linked to specifically checks for `type=file`. how is the `mode` getting changed for non-`file`s?

thesayyn: Oh my bad, that’s true. Do we something more generic than this though? I feel like we can do better in terms of Public API.

thesayyn: Let me thinker over the weekend and see if I could come up with a better api

alexeagle:
Hey, I'm sorry this code has moved over to https://github.com/bazel-contrib/tar.bzl - do you mind re-opening there? (fyi @thesayyn )

PR #1063, opened 2025-04-06
chore: docs without compiling protoc

alexeagle:


alexeagle:
@JesseTatasciore here's where I should also make write_source_files produce the patch file in a named output group :)

thesayyn (approved):
Nice one, forgot to actually stamp

PR #1066, opened 2025-04-12
feat: add riscv64 support to yq toolchain

kxxt:
This patch adds riscv64 support to yq toolchain. 

The last version 4.33.3 in `lib/private/yq_toolchain.bzl` does not have riscv64 binary releases in yq repository so I added the latest release to it.


kxxt:
It looks that yq update breaks some tests. I have pushed some fixes.
PTAL again. Thanks!

kxxt:
The style check should pass now. I forgot to reformat the bazel files. Sorry about that.

Docs update_27_test should also pass now. Somehow all the docs tests are failing in my environment  so my local testing didn't catch that.

image

PR #1068, opened 2025-04-14
feat(mtree-spec): add include_runfiles flag

skeggse:


lib/private/tar.bzl
```diff
@@ -446,43 +455,49 @@ def _mtree_impl(ctx):
...
+                allow_closure = True,
             )
+            if repo_mapping != None:
+                content.add(
+                    _mtree_line(_vis_encode(runfiles_dir + "/_repo_mapping"), "file", content = _vis_encode(repo_mapping.path)),
+                )
 
     ctx.actions.write(out, content = content)
 
-    return DefaultInfo(files = depset([out]), runfiles = ctx.runfiles([out]))
```
thesayyn, line 485: I don't think this needs to change at all.

thesayyn: These are probably equivalent but lets keep it same to make sure we don't break things.

skeggse: huh yeah look at that, same behavior!

thesayyn (approved):

PR #1069, opened 2025-04-21
feat: publish release attestations

alexeagle:
Allows users to trust that release artifacts are not tampered

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -35,7 +35,7 @@ jobs:
           subject-path: artifacts/*
   release:
     needs: build
-    uses: bazel-contrib/.github/.github/workflows/release_ruleset.yaml@443541da08f32816d70070ea30b5b0ee53d5e53d
+    uses: bazel-contrib/.github/.github/workflows/release_ruleset.yaml@v7.2.1
     with:
       release_files: |
         artifacts/*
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yml
```diff
@@ -26,22 +30,22 @@ jobs:
...
-      - uses: softprops/action-gh-release@v2
-        with:
-          # Use GH feature to populate the changelog automatically
-          generate_release_notes: true
-          files: |
-            artifacts/*
-            bazel-lib-*.tar.gz
-          body_path: release_notes.txt
-          fail_on_unmatched_files: true
+    uses: bazel-contrib/.github/.github/workflows/release_ruleset.yaml@443541da08f32816d70070ea30b5b0ee53d5e53d
```
kormide, line 38: This won't pass the SLSA verifier unless it's semver.

kormide: If you want to just test changes to the release_ruleset workflow you could use one of my test fixture repos like https://github.com/publish-to-bcr-dev/fixture-versioned, or I could test it for you.

alexeagle: Yup, this was a placeholder, now that the other PR landed I change this to v7.2.1

kormide (approved):

PR #1080, opened 2025-04-28
chore: update bsdtar to latest

alexeagle:


lib/private/tar_toolchain.bzl
```diff
@@ -34,28 +34,39 @@ BSDTAR_PLATFORMS = {
             "@platforms//cpu:x86_64",
         ],
     ),
+    "windows_arm64": struct(
+        release_platform = "win64",
```
thesayyn, line 38: hmm, i don't see where `release_platform` is being used. maybe remove?

thesayyn (approved):

PR #1083, opened 2025-05-06
refactor: delete tar macro/rule code

alexeagle:
It was split out into a separate module for discoverability. Leave behind a re-export to make this a non-breaking change for users.

FYI @zaucy


alexeagle:
Note, I should do another tar.bzl release before landing to avoid the breaking change of the flag flip of the unused_inputs feature.

dzbarsky (approved):

thesayyn (approved):

PR #1090, opened 2025-05-26
Add an executable_test rule and use it in write_source_file

calliecameron:
Adds executable_test, which checks that a file or directory's executable permissions are as intended. Implementation is mostly copied from diff_test. Tested on Linux, since that's what I have. As far as I remember, the executable permission doesn't exist on Windows, so I've made the Windows version of the test always pass.

Added this to write_source_files, so that when diff_test is true, the executable permission will be checked as well as diffing the contents.

Fixes https://github.com/bazel-contrib/bazel-lib/issues/1089.


calliecameron:
@alexeagle any update on reviewing this?

PR #1092, opened 2025-05-28
feat(diff_test): allow inline content for expected

alexeagle:
Shortens a lot of our BUILD files by avoiding an extra write_file call for literal content


kormide:
I don't love how build formatting puts `content2` before `file1`. How would you feel about changing the attr to `file_content_2`?

alexeagle:
Yeah polymorphic attribute accepting a list is a good idea, we do that in other places in this repo

PR #1095, opened 2025-05-30
refactor: Bazelrc presets: Almost always use common

hofbi:
Best practice seems to be to almost always use `common`(reminding me to C++'s [AAA rule](https://www.fluentcpp.com/2020/12/04/auto-for-types-but-not-for-concepts/)). Using `common` ensures that `build`, `fetch`, and `query` use the same flags and is easier for users since there is one recommended way to do it.

With https://github.com/bazelbuild/bazel/pull/18609, Bazel does not fail in case a flag does not belong to a common, so there is no harm in using common for Bazel 6.3+. Because of this, bazel6.bazelrc does not use common.

I kept all `test` flags as is for now, but we can also make them common if we want, either here or as follow up PR.


hofbi:
> I am not aware of any downside and there is even a minor upside: Before Bazel 9, `--test_env` invalidates `build` commands and thus shouldn't be set on `test`.

https://github.com/bazel-contrib/bazel-lib/pull/1099 does it

PR #1098, opened 2025-05-31
fix: Remove references to `@local_config_platform`

fmeum:
This repo is no longer available in Bazel@HEAD.


.github/workflows/release_prep.sh
```diff
@@ -80,6 +80,15 @@ aspect_bazel_lib_dependencies()
 
 aspect_bazel_lib_register_toolchains()
 
+# Create the host platform repository transitively required by bazel-lib
```
fmeum, line 83: We can't hide this step in `aspect_bazel_lib_register_toolchains` since it is located in the same file as `aspect_bazel_lib_dependencies`, which introduces the `platforms` repo, and we thus can't `load` from it.

lib/private/bats_toolchain.bzl
```diff
@@ -17,7 +17,7 @@ BATS_FILE_VERSIONS = {
 }
 
 BATS_CORE_TEMPLATE = """\
-load("@local_config_platform//:constraints.bzl", "HOST_CONSTRAINTS")
+load("@platforms//host:constraints.bzl", "HOST_CONSTRAINTS")
```
alexeagle, line 20: thanks! looks like https://github.com/bazelbuild/platforms/pull/86 was the change that made it possible for us to cleanup our workaround

alexeagle (approved):

.github/workflows/release_prep.sh
alexeagle: Frustrating but WORKSPACE will be done soon... 👍🏻

PR #1099, opened 2025-06-02
Make all bazelrc presets common

hofbi:
Follow up of https://github.com/bazel-contrib/bazel-lib/pull/1095 which is not only easier for users to follow but comes with a small improvement according to https://github.com/bazel-contrib/bazel-lib/pull/1095#issuecomment-2929695227.

### DIFF
--- a/.aspect/bazelrc/javascript.bazelrc
+++ b/.aspect/bazelrc/javascript.bazelrc
@@ -7,5 +7,5 @@
 # https://nodejs.org/de/docs/guides/debugging-getting-started/#command-line-options for more
 # details.
 # Docs: https://nodejs.org/en/docs/guides/debugging-getting-started/#command-line-options
-run:debug -- --node_options=--inspect-brk
-test:debug --test_env=NODE_OPTIONS=--inspect-brk
+common:debug -- --node_options=--inspect-brk
+common:debug --test_env=NODE_OPTIONS=--inspect-brk
--- a/lib/tests/bazelrc_presets/subset/javascript.bazelrc
+++ b/lib/tests/bazelrc_presets/subset/javascript.bazelrc
@@ -7,5 +7,5 @@
 # https://nodejs.org/de/docs/guides/debugging-getting-started/#command-line-options for more
 # details.
 # Docs: https://nodejs.org/en/docs/guides/debugging-getting-started/#command-line-options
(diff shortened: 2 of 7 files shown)


lib/tests/bazelrc_presets/subset/javascript.bazelrc
```diff
@@ -7,5 +7,5 @@
 # https://nodejs.org/de/docs/guides/debugging-getting-started/#command-line-options for more
 # details.
 # Docs: https://nodejs.org/en/docs/guides/debugging-getting-started/#command-line-options
-run:debug -- --node_options=--inspect-brk
-test:debug --test_env=NODE_OPTIONS=--inspect-brk
+common:debug -- --node_options=--inspect-brk
```
fmeum, line 10:
```suggestion
run:debug -- --node_options=--inspect-brk
```

.aspect/bazelrc/javascript.bazelrc
```diff
@@ -7,5 +7,5 @@
 # https://nodejs.org/de/docs/guides/debugging-getting-started/#command-line-options for more
 # details.
 # Docs: https://nodejs.org/en/docs/guides/debugging-getting-started/#command-line-options
-run:debug -- --node_options=--inspect-brk
-test:debug --test_env=NODE_OPTIONS=--inspect-brk
+common:debug -- --node_options=--inspect-brk
```
fmeum, line 10:
```suggestion
run:debug -- --node_options=--inspect-brk
```

hofbi:
@alexeagle tagging you on this since you also were involved in the first PR

jscissr:
performance.bazelrc now both enables and disables the build_runfile_links option, which doesn't make sense:

```
common --nobuild_runfile_links

# Needed prior to Bazel 8; see
# https://github.com/bazelbuild/bazel/issues/20577
common --build_runfile_links
```

Before these changes, it looked like this:
```
build --nobuild_runfile_links

# Needed prior to Bazel 8; see
# https://github.com/bazelbuild/bazel/issues/20577
coverage --build_runfile_links
```

I think the change from coverage to common should be reverted.

hofbi:
Good catch, fix is https://github.com/bazel-contrib/bazel-lib/pull/1111

PR #1101, opened 2025-06-05
feat: apply rules_shell fixes to WORKSPACE

novas0x2a:
772d605 fixed the bzlmod files but not the workspace files. I should note that there doesn't seem to be a way to test this locally, just in CI, which makes it pretty hard to figure out what's broken...


lib/repositories.bzl
```diff
@@ -360,6 +367,9 @@ def register_expand_template_toolchains(name = DEFAULT_EXPAND_TEMPLATE_REPOSITOR
         user_repository_name = name,
     )
 
+def register_shell_toolchains():
+    rules_shell_toolchains()
```
fmeum, line 371: This also has to be moved to the relnotes, the load from rules_shell otherwise still causes a cycle

novas0x2a: fixed, all green now.

fmeum (approved):

PR #1108, opened 2025-06-11
Use default value of 5 for repository downloader

hofbi:
Similar to https://github.com/bazel-contrib/bazel-lib/pull/1106 using the retry default from bazel 8 should help with stability issues


arrdem (approved):

alexeagle (approved):
nice, thanks

PR #1112, opened 2025-06-17
fix: fix ci for bazelrc preset docs

hofbi:
https://github.com/bazel-contrib/bazel-lib/commit/1ab2b68b4288037e4eaffbf946b6a986b298ab51 did not change the source document but only the generated file


alexeagle (approved):

alexeagle:
Oops I was naughty by going too fast 😅

PR #1113, opened 2025-06-23
feat(bazelrc): Add bazelrc preset for protoc

hofbi:
https://github.com/aspect-build/toolchains_protoc requires some bazel flags to be set, so why not make it a preset to be more user friendly.


alexeagle:
Could you re-file as discussion issue on bazelrc-preset.bzl please?

PR #1119, opened 2025-07-02
Rewrite write_source_files into a fast language

dzbarsky:
On a `write_source_files` tree with ~20 chained targets and a few files each (so total 90 files + directories) the runtimes goes from 5-7s to 20ms.

We will want to precompile the helper binary like the rest of the helpers.

Putting this up to get feedback first.

### DIFF
--- a/tools/write_source_files/main.go
+++ b/tools/write_source_files/main.go
@@ -1,0 +1,171 @@
+package main
+
+import (
+	"errors"
+	"fmt"
+	"io"
+	"os"
+	"path/filepath"
+	"strconv"
+	"strings"
+	"sync"
+	"syscall"
+)
+
+const (
+	maxConcurrency = 16
+	bufSize        = 128 * 1024 // 128 KB
+)
+
+var bufPool = sync.Pool{
+	New: func() any {
+		return make([]byte, bufSize)
+	},
+}
+
+type copyAction struct {
+	srcPath, dstPath string
+	executable       bool
+}
+
+// openDst tries to create dst (0755 if exec, default otherwise).
+// On ENOENT it will mkdir parent and retry once.
+func openDst(dst string, exec bool) (*os.File, error) {
+	try := func() (*os.File, error) {
+		if exec {
+			return os.OpenFile(dst, os.O_CREATE|os.O_WRONLY|os.O_TRUNC, 0755)
+		}
+		return os.Create(dst)
+	}
+
+	out, err := try()
+	if err == nil {
+		return out, nil
+	}
+	if os.IsNotExist(err) {
+		// parent missing → create and retry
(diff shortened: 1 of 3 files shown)


dzbarsky:
@alexeagle I updated the PR with a sketch of how some code can be shared. There's a few things bothering me:
1) I don't see how the existing implementation of copy_directory handles the executable bit (I'm not sure how Clonefile works but the fallback implementation just copies the bytes and doesn't seem to create the file properly?)
2) I think the implementation I came up with does fewer syscalls (for example it only issues MkdirAll when writes fail, it first attempts to copy a file and then fallsback to directory handling instead of issuing upfront Stat, etc)
3) The existing io.Copy does not reuse buffers, we should likely allocate a single buffer in the worker and reuse that with io.CopyBuffer
4) It feels like we are promoting the `copy_directory` walker as the "common" one, which creates a bit of an oddness for `copy_to_directory`
5) `CopyFile` requires an upfront Stat on `srcPath` to get the info, and we also need that to do that to decide whether to issue a CopyFile or CopyDirectory. In the current draft this happens serially and then execution of all copy actions is done by the worker pool, but ideally there wouldn't be a serial portion at all. This would mean doing the "try to copy as file and fallback to directory" inside the worker itself, are we ok with that? (It would likely slightly speedup existing copy_directory as well...)

Overall I can't tell if going down this path will create a mess or result in improvements to the existing implementations, but if we do want to unify we will probably want some prep PRs

r0bobo:
It would be great if `write_source_files` also didn't touch the destination file if there is no diff. Otherwise it unnecessarily triggers lsp:s to reload files and such.

tools/write_source_files/main.go
```diff
@@ -0,0 +1,171 @@
...
+
+	out, err := openDst(dst, exec)
+	if err != nil {
+		return err
+	}
+	defer out.Close()
+
+	buf := bufPool.Get().([]byte)
+	defer bufPool.Put(buf)
+	if _, err := io.CopyBuffer(out, in, buf); err != nil {
```
fmeum, line 71: Did you check that this is actually faster than plain `io.Copy`? I would expect that to use `sendfile` and its analogues under the hood which don't even need a (userspace) buffer.

peakschris:
Any chance of this moving forwards?
I'd very much like diff_test to be part of this, as that is very challenging cross platform (ignoring line endings, for example, is difficult with built-in tools on windows)

dzbarsky:
We've been using this patch for over a year :) it didn't seem like there was a ton of appetite from the project owners here to get it upstreamed. Alex was mentioning a desire to get rid of these rules in favor of proprietary Aspect CLI workflows, though I'm not sure how current Aspect folks are thinking about it

fmeum:
Happy to review and merge this when someone sets up the release process for the precompiled binaries

PR #1123, opened 2025-07-08
refactor(3.0): rename module to bazel_lib

alexeagle:
_targets new 3.0 branch_

Remove Aspect branding, as this was donated to LF and is community-maintained.

Note, I am not sure about aliasing toolchain_type symbols back to aspect_bazel_lib for compatibility.

BREAKING CHANGES
- jq rule was removed, use https://github.com/bazel-contrib/jq.bzl instead
- yq rule was removed, use https://github.com/bazel-contrib/yq.bzl instead
- tar rule was removed, use https://github.com/bazel-contrib/tar.bzl instead
- bazelrc presets are removed, use https://github.com/bazel-contrib/bazelrc-preset.bzl instead


arrdem (approved):

fmeum (approved):
Yay, thanks for the PR (and the donation)!

If we think that most users (and in particular BCR modules) won't register their own toolchains for the types defined in this module, I don't see a need to alias them.

What about providers? Are there any rules for which it would become impossible to add the equivalent rule from aspect_bazel_lib to `deps`?

alexeagle:
I'll scan the repo for public provider symbols, that's a good point. DirectoryPathInfo is the only one that comes to mind.

PR #1125, opened 2025-07-13
chore: add windows_arm64 support to yq

dennisameling:
Partially addresses https://github.com/bazel-contrib/bazel-lib/issues/1076

Upstream has introduced support for windows_arm64 recently, so it can be referenced now.


alexeagle (approved):
Thanks!

PR #1135, opened 2025-07-26
refactor: remove yq toolchain

alexeagle:
BREAKING CHANGE:
The yq toolchain type is no longer in bazel-lib. Use the symbol @yq.bzl//yq/toolchain:type in case you were registering a custom YQ toolchain.


fmeum:
Does the new toolchain type exist yet? I only see references to bazel-lib in yq.bzl. Otherwise this LGTM.

fmeum (approved):

jbedard:
@alexeagle shouldn't we move the implementation to `yq.bzl` first? I guess when this is merging to a `3.0` branch it doesn't really matter what order we do it, but we'll have to do it...

jbedard (approved):

PR #1148, opened 2025-08-15
chore: simplify CI checks on GitHub Actions

alexeagle:


alexeagle:
Not much point landing this change on the `main` branch which is for the 2.x releases. Will just work on the 3.x branch now.

PR #1159, opened 2025-08-22
feat: support diff_test/write_source_files on Windows and enable most tests

peakschris:
Enable write_source_files with diff_test for windows; allowing documentation checks and updates to run on Windows for repos that use these macros. Enable most disabled windows tests.

**Features:**
* write_source_files fixed for windows
* diff_test enabled for windows; non-binary comparison so allows different line endings
* improved runfiles lookup for both rules:
    * windows: runfiles are looked up via runfiles manifest so --enable_runfiles not required
    * other platforms: left runfiles handling untouched to avoid regressing working code
* assert_contains now uses runfiles manifest lookup
* fixed a bug in copy_directory and copy_to_directory that would result in 'failed to get realpath of dangling symlink' error
* most tests that were disabled for windows / enable_runfiles only have been enabled for windows without runfiles
* enable docs folder tests on windows CI (for extra test coverage of the diff_test macro)

**Fixes:**
* https://github.com/bazel-contrib/bazel-lib/issues/1142


alexeagle:
I'll definitely do a lot of the lifting to get the 3.0 rolled out to everything I maintain!
Thank you!

alexeagle (approved):

PR #1187, opened 2025-10-01
Split excludes into separate package rules

hofbi:
As per suggestion from https://github.com/bazel-contrib/bazel-lib/commit/220e23f19f49cb1e458ae0534a30624366a411a3#commitcomment-166896515


alexeagle (approved):
thanks!

PR #1188, opened 2025-10-02
refactor: delete stardoc_with_diff_test

### LINKED ISSUE #1185: 3.0: drop stardoc
https://github.com/bazel-contrib/bazel-lib/blob/main/lib/docs.bzl relies on the stardoc module, which has a big fat dependency list, doesn't work with Bazel 6, and more. See "Legacy WORKSPACE Setup" https://github.com/bazelbuild/stardoc/releases/tag/0.8.0 for the ugliness.

Rules_go has their own `stardoc_with_diff_test` https://github.com/bazel-contrib/rules_go/blob/master/docs/doc_helpers.bzl
so we know users could just do this themselves.

We no longer use this rule ourselves since we can publish docs for BCR to render, see https://github.com/bazel-contrib/rules-template/pull/160 and my upcoming blog post https://hashnode.com/draft/68dc18b3df4b794c7b0074a6

Perhaps we move it to the stardoc repo, or just delete it with no replacement.

@fmeum @malt3 FYI

alexeagle:
The stardoc module has heavy dependencies. See remaining justification in the attached issue.

Fixes #1185

BREAKING CHANGE: bazel-lib no longer provides a convenience for checking in generated docs. Alternatives:
- (recommended): publish `starlark_doc_extract` rules on releases instead of checking in generated files; see https://blog.aspect.build/stardocs-on-bcr
- Copy the macro into your own repo like rules_go does, combining a stardoc target with a `diff_test`
- Use `aspect_bazel_lib` 2.x along with `bazel_lib`; the former provides this rule.


jbedard (approved):

jbedard:
Great 👍

PR #1189, opened 2025-10-03
chore: test on Bazel 9, drop Bazel 6

alexeagle:
Before 3.0 final we can drop this maintenance burden


jbedard (approved):

fmeum (approved):

alexeagle:
Yeah that's a nice solution, I think a dep an with_cfg is okay

fmeum:
Pushed a simpler solution. `cgo_off` is something we want to get rid of anyway.

PR #1206, opened 2025-11-03
fix(run_binary): include RunEnvironmentInfo

### LINKED ISSUE #1200: run_binary does not include executable environment from RunEnvironmentInfo
If the `tool` provided to `run_binary` has a `RunEnvironmentInfo` provider, that environment is not included when running the tool. For example, this is how `py_binary` from rules_py [sets provided environment variables](https://github.com/aspect-build/rules_py/blob/340ac788719a3013f08f63c442c986b9f55188b9/py/private/py_binary.bzl#L131-L134).

This basic patch includes the environment and allows envs specified directly to `run_binary` to override the ones provided by the tool:
```diff
diff --git lib/private/run_binary.bzl lib/private/run_binary.bzl
index b480a27..9f12b21 100644
--- lib/private/run_binary.bzl
+++ lib/private/run_binary.bzl
@@ -51,6 +51,8 @@ Possible fixes:
     for a in ctx.attr.args:
         args.add_all(split_args(expand_variables(ctx, ctx.expand_location(a, targets = ctx.attr.srcs), inputs = ctx.files.srcs, outs = outputs)))
     envs = {}
+    if RunEnvironmentInfo in ctx.attr.tool:
+        envs = ctx.attr.tool[RunEnvironmentInfo].environment
     for k, v in ctx.attr.env.items():
         envs[k] = expand_variables(ctx, ctx.expand_location(v, targets = ctx.attr.srcs), inputs = ctx.files.srcs, outs = outputs, attribute_name = "env")
```

The patch is working for our use cases right now. Happy to submit it as a PR if the implementation seems generally correct.

alexeagle:
Thanks for the patch @brett-patterson-ent

Fixes #1200

### DIFF
--- a/lib/private/run_binary.bzl
+++ b/lib/private/run_binary.bzl
@@ -51,6 +51,8 @@ Possible fixes:
     for a in ctx.attr.args:
         args.add_all(split_args(expand_variables(ctx, ctx.expand_location(a, targets = ctx.attr.srcs), inputs = ctx.files.srcs, outs = outputs)))
     envs = {}
+    if RunEnvironmentInfo in ctx.attr.tool:
+        envs = ctx.attr.tool[RunEnvironmentInfo].environment
     for k, v in ctx.attr.env.items():
         envs[k] = expand_variables(ctx, ctx.expand_location(v, targets = ctx.attr.srcs), inputs = ctx.files.srcs, outs = outputs, attribute_name = "env")
 


lib/private/run_binary.bzl
```diff
@@ -51,6 +51,8 @@ Possible fixes:
     for a in ctx.attr.args:
         args.add_all(split_args(expand_variables(ctx, ctx.expand_location(a, targets = ctx.attr.srcs), inputs = ctx.files.srcs, outs = outputs)))
     envs = {}
+    if RunEnvironmentInfo in ctx.attr.tool:
```
fmeum: I can see why this is useful, but it also deviates from pretty much all other rules by applying `env` transitively instead of only with `bazel run/test`. This could result in confusion and uncertainty, which may be worse than the status quo.

hofbi: Is there some good documentation that shows when `env` is forwarded and when not?

fmeum: Yes, that's the documented status quo. I'm not against changing this in general, but doing so for a single rule feels wrong.

PR #1207, opened 2025-11-05
chore(deps): Bazel 9 is RC now

alexeagle:


fmeum:
The failure should go away with an update to latest rules_go.

PR #1212, opened 2025-11-24
chore: Update OS version in presubmit configuration

hofbi:
Don't test on an EOL OS


.bcr/presubmit.yml
```diff
@@ -1,7 +1,7 @@
 bcr_test_module:
   module_path: "e2e/smoke"
   matrix:
-    platform: ["debian10", "macos", "ubuntu2004", "windows"]
+    platform: ["debian10", "macos", "ubuntu2204", "windows"]
```
hofbi:
```suggestion
    platform: ["debian11, "macos", "ubuntu2204", "windows"]
```

alexeagle (approved):

PR #1219, opened 2025-12-16
fix(2.x): alias DirectoryPath to bazel_lib

jbedard:
The goal is to make libraries such as rules_js not have to depend on both v2 and v3 for backward compatibility.


jbedard:
@fmeum I tagged a [v2.22.1](https://github.com/bazel-contrib/bazel-lib/releases/tag/v2.22.1) to release this change but [publish is failing](https://github.com/bazel-contrib/bazel-lib/actions/runs/20259633819/job/58169292542).

Do you understand what's going on there?

PR #1229, opened 2026-01-08
fix(2.x): alias copy_directory_bin_action to bazel_lib

jbedard:
So 2.x does not need to refer to the private `COPY_EXECUTION_REQUIREMENTS`.


jbedard:
See https://github.com/bazel-contrib/bazel-lib/pull/1233

PR #1233, opened 2026-01-12
fix(2.x): reference aspect_bazel_lib toolchains in public consts

jbedard:
The `COPY_FILE_TO_BIN_TOOLCHAINS` simply [references COPY_FILE_TOOLCHAINS](https://github.com/bazel-contrib/bazel-lib/blob/v3.1.0/lib/private/copy_to_bin.bzl#L20). In `2.x` we must reference the 2.x `COPY_FILE_TOOLCHAINS` in this constant so we only directly expose the `aspect_bazel_lib` labels, even thought they are just aliases to `bazel-lib`.

### DIFF
--- a/e2e/copy_action/copy.bzl
+++ b/e2e/copy_action/copy.bzl
@@ -1,6 +1,7 @@
 """Rule that uses copy actions"""
 
 load("@aspect_bazel_lib//lib:copy_file.bzl", "COPY_FILE_TOOLCHAINS", "copy_file_action")
+load("@aspect_bazel_lib//lib:copy_to_bin.bzl", "COPY_FILE_TO_BIN_TOOLCHAINS")
 
 def _simple_copy_file_impl(ctx):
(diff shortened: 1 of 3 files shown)


e2e/copy_action/copy.bzl
```diff
@@ -22,5 +23,5 @@ simple_copy_file = rule(
         "src": attr.label(mandatory = True, allow_files = True),
         "out": attr.output(mandatory = True),
     },
-    toolchains = COPY_FILE_TOOLCHAINS,
+    toolchains = COPY_FILE_TOOLCHAINS + COPY_FILE_TO_BIN_TOOLCHAINS,
```
alexeagle: This doesn't make sense to me, why is the shape of this changing? Shouldn't it just be a  reference to aspect_bazel_lib <-> bazel_lib and/or an `alias` ?

PR #1252, opened 2026-04-21
feat: add `tools` parameter to `run_binary()`

### LINKED ISSUE #676: [FR]: Allow additional tools to be passed to `run_binary`

Currently the arguments to `run_binary` include `tool` which is used to specify the tool which will be executed during the build. This argument is used in variable expansion in `env` and `args`.

However, if you want to use another tool in `env` or `args`, it's not possible.

### Describe the feature

Add `tools` to `run_binary`. This `attr` is used in variable expansion, but the tools are otherwise not executed

acozzette:
One missing feature in `run_binary()` is that it does not support a way to provide any additional dependencies for the execution platform. All inputs other than the `tool` binary itself are built for the target platform. This means that, for example, there is no reliable way to provide another binary for the tool to invoke during the build action. If the execution and target platforms are sufficiently different then this will fail.

This change remedies the problem by introducing a new `tools` parameter analogous to the one on
[genrule](https://bazel.build/reference/be/general#genrule.tools). This parameter works similarly to `srcs` except that it builds the dependencies in the `exec` configuration.

Fixes #676.

### DIFF
--- a/lib/private/run_binary.bzl
+++ b/lib/private/run_binary.bzl
@@ -50,13 +50,15 @@ Possible fixes:
 
     # Location and Make variable expansion can reference paths that aren't path mapped.
     can_path_map = True
+    targets = ctx.attr.tools + ctx.attr.srcs
+    inputs = ctx.files.tools + ctx.files.srcs
     for a in ctx.attr.args:
-        expanded = expand_variables(ctx, ctx.expand_location(a, targets = ctx.attr.srcs), inputs = ctx.files.srcs, outs = outputs)
+        expanded = expand_variables(ctx, ctx.expand_location(a, targets = targets), inputs = inputs, outs = outputs)
         can_path_map = can_path_map and expanded == a
         args.add_all(split_args(expanded))
     envs = {}
     for k, v in ctx.attr.env.items():
-        envs[k] = expand_variables(ctx, ctx.expand_location(v, targets = ctx.attr.srcs), inputs = ctx.files.srcs, outs = outputs, attribute_name = "env")
+        envs[k] = expand_variables(ctx, ctx.expand_location(v, targets = targets), inputs = inputs, outs = outputs, attribute_name = "env")
         can_path_map = can_path_map and envs[k] == v
 
     stamp = maybe_stamp(ctx)
@@ -72,6 +74,7 @@ Possible fixes:
         outputs = outputs,
         inputs = inputs,
         executable = ctx.executable.tool,
+        tools = ctx.files.tools,
         arguments = [args],
         resource_set = resource_set(ctx.attr),
         mnemonic = ctx.attr.mnemonic if ctx.attr.mnemonic else None,
@@ -93,6 +96,10 @@ _run_binary = rule(
             executable = True,
             allow_files = True,
             mandatory = True,
+            cfg = "exec",
+        ),
+        "tools": attr.label_list(
+            allow_files = True,
             cfg = "exec",
         ),
         "env": attr.string_dict(),
@@ -122,6 +129,7 @@ def run_binary(
         execution_requirements = None,
         use_default_shell_env = False,
         stamp = 0,
+        tools = [],
         **kwargs):
     """Runs a binary as a build action.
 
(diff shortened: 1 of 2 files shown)


lib/private/run_binary.bzl
```diff
@@ -113,6 +116,7 @@ def run_binary(
         name,
         tool,
         srcs = [],
+        tools = [],
```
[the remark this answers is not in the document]
acozzette: Fixed

jbedard (approved):

jbedard:
LGTM but I'd like at least one more reviewer to take a look before adding a new param like this. @alexeagle or @fmeum?

jbedard:
or @kormide?

lib/private/run_binary.bzl
```diff
@@ -51,12 +51,12 @@ Possible fixes:
     # Location and Make variable expansion can reference paths that aren't path mapped.
     can_path_map = True
     for a in ctx.attr.args:
-        expanded = expand_variables(ctx, ctx.expand_location(a, targets = ctx.attr.srcs), inputs = ctx.files.srcs, outs = outputs)
+        expanded = expand_variables(ctx, ctx.expand_location(a, targets = ctx.attr.tools + ctx.attr.srcs), inputs = ctx.files.tools + ctx.files.srcs, outs = outputs)
```
jbedard: Should `ctx.attr.tools + ctx.attr.srcs` be done once outside the loops?

acozzette: Sure, I made that change and also did the same for `ctx.files.tools + ctx.files.srcs`.

lib/tests/run_binary/BUILD.bazel
```diff
@@ -80,3 +82,216 @@ assert_contains(
     actual = ":dir_a_rootpath",
     expected = "lib/tests/run_binary/dir_a_out",
 )
+
+# Verify that items in `tools` are built in the exec configuration.
+# In Bazel, exec-configured outputs live under a bazel-out subdirectory
+# containing "-exec-", which distinguishes them from target-configured outputs.
```
kormide: Is this output path segment is stable? Can be modified by any Bazel flags?

I guess it doesn't matter since these are just tests and can be adjusted if that ever changes.

lib/tests/run_binary/BUILD.bazel
```diff
@@ -80,3 +82,216 @@ assert_contains(
...
+    srcs = [":gen_exec_os"],
+    target_platform = ":linux",
+)
+
+platform_transition_filegroup(
+    name = "exec_os_macos",
+    srcs = [":gen_exec_os"],
+    target_platform = ":macos",
+)
+
```
kormide, line 215: Do you also want a Windows transition? Or would that be too annoying to write a `cmd_ps`/`cmd_bat` in the genrule?

lib/tests/run_binary/BUILD.bazel
```diff
@@ -80,3 +82,216 @@ assert_contains(
...
+    outs = ["platform_test.sh"],
+    # The genrule runs in the execroot so execpaths resolve directly. It reads
+    # the two files and bakes their paths and contents into the test script as
+    # shell assignments; the assertions then run at test time via sh_test.
+    cmd = """
+        exec_os_linux_path=$(execpath :exec_os_linux)
+        exec_os_macos_path=$(execpath :exec_os_macos)
+        {
+            echo '#!/usr/bin/env sh'
+            echo 'set -eu'
```
kormide: nit: prefer long names of the flags for readability

kormide (approved):
LGTM. I appreciate the thorough and clever testing.

fmeum (approved):

lib/tests/run_binary/BUILD.bazel
acozzette: I suspect it's probably not stable actually. But I realized that we can just use a genrule to figure out what the correct path should be. So I just updated the test to assert that run_binary and genrule agree with each other on the path for a `tools` input.

lib/tests/run_binary/BUILD.bazel
acozzette: Sure, I went ahead and added a Windows transition. Luckily this did not require `cmd_ps` or `cmd_bat` or anything, because we don't need anything to be able to run on Windows (we're just producing a trivial text file for Windows).

lib/tests/run_binary/BUILD.bazel
acozzette: Done

PR #1253, opened 2026-04-22
feat: add `repo_utils.arch()` and use `rctx.os.arch` where available.

mattyclarkson:
Catch failing `uname` executions rather than returning empty strings.


thesayyn (approved):

thesayyn:
you have some failures

mattyclarkson:
> you have some failures

Fixed up spelling mistake.

PR #1254, opened 2026-04-29
fix: restore missing release binaries

acozzette:
The release process seems to have broken due to this bug in download-artifact: https://github.com/actions/download-artifact/issues/455 We use the release_ruleset.yaml reusable workflow, which downloads all artifacts [here](https://github.com/bazel-contrib/.github/blob/af28703814cf26522cf5e44d219efece91293123/.github/workflows/release_ruleset.yaml#L103). But since there is only one artifact (`go-binaries.zip`), we hit that bug which causes the binaries to appear directly in the working directory instead of in `./go-binaries`.

This change fixes the problem by nesting everything inside a `go-binaries` directory inside `go-binaries.zip`. This way when the file is unzipped, all the binaries appear to be in the same place they were before.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -32,9 +32,15 @@ jobs:
       - uses: actions/checkout@v4
       - name: Build Go Binaries
         env:
-          # NB: this variable is read by tools/release/copy_release_artifacts.sh
-          # and must match the path in release_prep.sh under RELEASED_BINARY_INTEGRITY
-          DEST: binaries
+          # NB: this variable is read by
+          # tools/release/copy_release_artifacts.sh.
+          # We intentionally nest everything inside a go-binaries/ directory
+          # inside go-binaries.zip. This works around a unintended change in
+          # actions/download-artifact:
+          # https://github.com/actions/download-artifact/issues/455
+          # The release_ruleset reusable workflow picked up that change in
+          # https://github.com/bazel-contrib/.github/pull/50.
+          DEST: binaries/go-binaries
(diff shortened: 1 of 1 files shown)


kormide (approved):

.github/workflows/release.yaml
```diff
@@ -32,9 +32,13 @@ jobs:
...
-          # NB: this variable is read by tools/release/copy_release_artifacts.sh
-          # and must match the path in release_prep.sh under RELEASED_BINARY_INTEGRITY
-          DEST: binaries
+          # NB: this variable is read by
+          # tools/release/copy_release_artifacts.sh.
+          # We intentionally nest everything inside a go-binaries/ directory
+          # inside go-binaries.zip. This works around a unintended change in
+          # actions/download-artifact:
+          # https://github.com/actions/download-artifact/issues/455
+          DEST: binaries/go-binaries
```
kormide, line 43: nit: for a clearer paper trail, note the reusable release workflow change that updated the action version

acozzette: Sure, I updated the comment to link to that pull request.

jbedard (approved):

jbedard:
FYI @alexeagle @fmeum this was introduced in https://github.com/bazel-contrib/bazel-lib/pull/1249 and caused the [3.3.0](https://github.com/bazel-contrib/bazel-lib/releases/tag/v3.3.0) release to not have the artifacts in it

fmeum (approved):

PR #1257, opened 2026-06-19
fix: use git apply on Windows when patch_directory is set

thii:
patch.exe is not available on stock Windows, so fall back to git apply when applying patches in a patch_directory.


alexeagle:
thanks @thii - how come it's a draft?

thii:
The change itself is fine. It's still arguable that git is also not available in stock Windows (though less likely if you're building with Bazel). I think a better fix is probably adding support for `patch_directoy` in Bazel's native patch:

- https://github.com/bazelbuild/bazel/issues/19772

thii:
Sorry, closing this since I think this is not the right approach.

PR #1260, opened 2026-06-29
feat: add stderr, exit_code_out, chdir, silent_on_success to run_binary

### LINKED ISSUE #575: run_binary is missing features from js_run_binary
### What happened?

run_binary doesn't have any wrapper program, so it can't do these things:

-   capture stdout/stderr/exit code as "outputs"
-   `chdir` to a specific working directory
-   throw away logspam on success

Now that we have a good pattern for shipping binaries written in Go, we could have a wrapper program for all run_binary calls to provide these features, similar to how we added stamping ability to expand_template.

### Version

HEAD

### How to reproduce

_No response_

### Any other information?

_No response_

alexeagle:
## Summary

Completes the `run_binary` feature set requested in #575. PR #1259 added the `spawn_binary` wrapper + toolchain and the first feature (stdout capture); this PR extends the wrapper with the remaining `js_run_binary`-style features:

| attribute | behavior |
|---|---|
| `stderr` | Capture the binary's standard error to a file (mirrors `stdout`). |
| `exit_code_out` | Write the binary's exit code to a file and let the action **succeed** even when the binary exits non-zero. |
| `chdir` | Run the binary with a different working directory (location/make-var expanded). |
| `silent_on_success` | Buffer non-captured stdout/stderr and only forward it to the console when the binary exits non-zero — silences logspam from successful steps. |

As before, the wrapper is used **only** when one of these features is requested; a plain `run_binary` spawns its tool directly and does not require the (optional) `spawn_binary` toolchain.

## Wrapper behavior / transparency

- stdin and the environment are passed through verbatim; a stream is only redirected when its feature is requested.
- Termination signals are relayed to the child; exit status is reported faithfully (`128+signum` on signal death).
- `chdir` resolves a relative tool path to absolute before changing the child's directory, so the tool still resolves; the wrapper itself stays in the exec root so capture-file paths open correctly.
- Capture files (`stdout`/`stderr`/`exit_code_out`) are real action outputs but are kept out of make-variable (`$@`, `$(@D)`) expansion, consistent with #1259.

## Tests

Added to `lib/tests/run_binary`:
- `captured_stderr_test` — stderr captured, stdout passes through.
- `captured_exit_code_test` — tool exits 3, action succeeds, exit code file contains `3`.
- `silent_success_test` — successful run still produces declared outputs through the buffered path.
- `chdir_test` — tool reads a file by a path relative to the `chdir` directory.

All `run_binary` / `run_binary_expansions` / `stamping` tests pass; `gofmt` + `buildifier` clean. The new wrapper behavior was also smoke-tested directly (silent-on-success reveals output only on failure; `exit_code_out` forces exit 0; chdir + relative read).

## Docs

Updated the `run_binary` overview and added attribute docs for `stderr`, `exit_code_out`, `chdir`, and `silent_on_success` — the same documentation spots introduced in #1259.

## Note

The wrapper changes ship in prebuilt `spawn_binary` binaries on the next release; dev/prerelease trees build it from source via the existing `source_toolchains_repo` path, so no integrity changes are needed here.

Closes #575

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/lib/private/run_binary.bzl
+++ b/lib/private/run_binary.bzl
@@ -35,27 +35,40 @@ def _run_binary_impl(ctx):
         outputs.append(out_dir)
 
     # When a feature that requires interposing on the spawned process is requested
-    # (currently only stdout capture), the tool is launched through the spawn_binary
-    # wrapper instead of directly. When no such feature is used the tool is spawned
-    # directly, avoiding the extra process and keeping the wrapper out of the action
-    # inputs entirely.
+    # (capturing stdout/stderr/exit code, changing directory, or silencing output on
+    # success), the tool is launched through the spawn_binary wrapper instead of
+    # directly. When no such feature is used the tool is spawned directly, avoiding
+    # the extra process and keeping the wrapper out of the action inputs entirely.
     #
-    # `outputs` (outs + out_dirs) are also what make variables like $@ and $(@D)
-    # expand to. The stdout capture file is a real action output but is intentionally
-    # kept out of that set, so that a tool with a single declared `outs` file keeps
-    # working with $@; it is only used for expansion when it is the sole output.
+    # The capture output files (stdout/stderr/exit_code_out) are real action outputs,
+    # but are intentionally kept out of the make-variable output set (`outputs`, used
+    # by $@ and $(@D)), so that a tool with a single declared `outs` file keeps
+    # working with $@. They are only used for expansion when there is no other output.
     stdout = ctx.outputs.stdout
+    stderr = ctx.outputs.stderr
+    exit_code_out = ctx.outputs.exit_code_out
+    silent_on_success = ctx.attr.silent_on_success
+    chdir = ctx.attr.chdir
+    use_wrapper = bool(stdout or stderr or exit_code_out or silent_on_success or chdir)
+
+    capture_outputs = [o for o in [stdout, stderr, exit_code_out] if o]
+    action_outputs = outputs + capture_outputs
+    expansion_outputs = outputs if outputs else action_outputs
+
     spawn_binary = None
-    action_outputs = list(outputs)
-    if stdout:
-        action_outputs.append(stdout)
+    if use_wrapper:
         spawn_binary_toolchain = ctx.toolchains["//lib:spawn_binary_toolchain_type"]
         if not spawn_binary_toolchain:
-            fail("run_binary target {} sets the `stdout` attribute, which requires the spawn_binary toolchain, but none is registered. Register it with register_spawn_binary_toolchains() (WORKSPACE) or the bazel_lib `spawn_binary` toolchain extension (bzlmod).".format(ctx.label))
+            fail("run_binary target {} requests a feature that requires the spawn_binary toolchain (one of stdout, stderr, exit_code_out, silent_on_success, chdir), but none is registered. Register it with register_spawn_binary_toolchains() (WORKSPACE) or the bazel_lib `spawn_binary` toolchain extension (bzlmod).".format(ctx.label))
         spawn_binary = spawn_binary_toolchain.spawn_binary_info.bin
-        args.add("--stdout", stdout)
-        args.add("--")
-        args.add(ctx.executable.tool)
+        if stdout:
+            args.add("--stdout", stdout)
+        if stderr:
+            args.add("--stderr", stderr)
+        if exit_code_out:
+            args.add("--exit-code-out", exit_code_out)
+        if silent_on_success:
+            args.add("--silent-on-success")
 
     if len(action_outputs) < 1:
         fail("""\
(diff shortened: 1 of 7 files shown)


tools/spawn_binary/main.go
```diff
@@ -1,71 +1,105 @@
...
 	}
 	if len(cmdArgs) == 0 {
 		return fatalf("no command specified; expected: spawn_binary [flags] -- <tool> [args...]")
 	}
 
+	// When changing the child's working directory, resolve a relative tool path to
+	// an absolute one first, so it still refers to the same file (the wrapper itself
+	// keeps running in the original working directory, so the capture output paths,
+	// which it opens, do not need adjusting).
+	if opts.chdir != "" && strings.ContainsRune(cmdArgs[0], '/') {
```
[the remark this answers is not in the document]
alexeagle: Good catch — fixed in 7aea646. The wrapped tool is always a declared input file (never a PATH command), so the wrapper now resolves it to an absolute path (relative to the execution root) unconditionally before spawning, rather than only when the path contains a `/`. This fixes the bare workspace-root execpath case (`exec.Command("tool.sh")` no longer does a PATH lookup) as well as the `--chdir` case. Added Go unit tests for bare tool paths with and without `--chdir`, plus arg parsing and exit-code capture/propagation. Verified it also resolves the latent issue on the already-released stdout path.

kormide:
What do you think of an attribute that allows you to fail the action on a specific code regardless of whether `exit_code_out` is set? For example, in the case of `diff`, it exits 0 on no diff, 1 on diff, and 2 if something went wrong. In diff.bzl I'm interested in failing the action on `2` and I'm currently doing that via bash.

Perhaps you could add a `fail_on = [...]` or `succeed_on = [..]`?

lib/private/run_binary.bzl
```diff
@@ -72,12 +85,21 @@ Possible fixes:
...
-
     # Location and Make variable expansion can reference paths that aren't path mapped.
     can_path_map = True
     targets = ctx.attr.tools + ctx.attr.srcs
     inputs = ctx.files.tools + ctx.files.srcs
+
+    if use_wrapper:
+        if chdir:
+            expanded_chdir = expand_variables(ctx, ctx.expand_location(chdir, targets = targets), inputs = inputs, outs = expansion_outputs, attribute_name = "chdir")
+            can_path_map = can_path_map and expanded_chdir == chdir
```
kormide: nit: the `can_path_map and` portion isn't necessary

kormide: Could you also add a clarifying comment describing what the path mapping does?

tools/spawn_binary/BUILD.bazel
```diff
@@ -16,3 +16,12 @@ go_binary(
...
 )
+
+go_test(
+    name = "spawn_binary_test",
+    srcs = [
+        "parse_test.go",
+        "run_unix_test.go",
+    ],
+    embed = [":spawn_binary_lib"],
+)
```
kormide, line 34: Should this test only run on linux platforms?

kormide (commented):
Looks good, but I'd like to discuss whether we want something to differentiate exit codes that fail or succeed the action because I think it'll have some interactions with `silent_on_success`.

alexeagle:
Thanks for the review — addressed the nits (path-mapping comment, `can_path_map` simplification, and `NOT_WINDOWS` on `spawn_binary_test` since the integration tests are unix-only).

On selective exit-code handling: I agree this is worth adding, and it should share the same definition of "success" as `silent_on_success` (i.e. buffered output is revealed when the action is treated as a failure, not when `code != 0`).

**Proposed API: `fail_on`**

```python
run_binary(
    name = "diff",
    tool = ":diff",
    exit_code_out = "exit_code.txt",
    fail_on = [2],  # diff: 0=no diff, 1=diff found, 2=error
    silent_on_success = True,
    ...
)
```

**Semantics**

| Configuration | Action succeeds when |
|---|---|
| Default (no `exit_code_out`, no `fail_on`) | Child exits 0 |
| `exit_code_out` only | Always (current behavior) |
| `fail_on = [2]` | Child exit code ∉ `fail_on` |
| `exit_code_out` + `fail_on = [2]` | Same as `fail_on` alone; exit code is still written to the file |

For the `diff` case: exit `0` and `1` succeed the action; exit `2` fails it. With `silent_on_success`, exits `0` and `1` stay quiet; exit `2` reveals buffered stdout/stderr.

**Why `fail_on` over `succeed_on`**

For tools like `diff`, the failure set is small and explicit (`[2]`). An allow-list (`succeed_on = [0, 1]`) works too but is easier to get wrong when a tool adds new exit codes later. I'd ship `fail_on` first; we can add `succeed_on` later if there's demand (they're dual views of the same policy).

**Implementation sketch**

- `run_binary` attribute: `fail_on = attr.int_list()`
- `spawn_binary` flag: `--fail-on=2` (repeatable or comma-separated)
- After the child exits: write `exit_code_out` if set, then return `0` unless the code is in `fail_on`
- `silent_on_success` uses that same success/failure decision instead of `code == 0`

Happy to add this in a follow-up commit on this PR if that works for you.

kormide:
That sounds fine to me. Supporting `fail_on` and optionally `succeed_on` later on would be nice, and both appear orthogonal to what's in this PR.

kormide (approved):

alexeagle:
okay, I'll send a follow-up to keep things cooking, thanks!

PR #1261, opened 2026-07-03
fix: resolve copy helper toolchains as Labels

keegancsmith:
The --incompatible_auto_exec_groups flag requires Starlark actions to declare whether their executable/tools come from an action toolchain. This is not a Bazel default today, but it is exposed through strict preset configurations and is intended to catch rules that rely on implicit action toolchain inference.

The copy helpers currently pass several action and rule toolchain labels as strings. Those strings are resolved in the consuming repository's mapping, which can fail for rules instantiated in extension-generated repositories where @bazel_lib is not visible as an apparent repository name. Use pre-resolved Label values so the toolchain behavior is explicit without changing the helpers' inputs or execution model.


fmeum (approved):

keegancsmith:
@fmeum CI is failing, but its the same failure on main so I believe is unrelated to this PR.

keegancsmith:
Whoops, accidentally pushed a potential fix for the CI failure on main onto this branch, will open another PR rather

PR #1262, opened 2026-07-04
fix(copy_directory): avoid BSD diff directory loop

keegancsmith:
The copy_directory_bin_action tests intentionally compare a source fixture that contains symlinks against copy_directory outputs that have dereferenced those symlinks. That means the test harness needs recursive diff to follow the source symlinks and compare contents rather than symlink metadata.

The previous fixture made the directory symlink `ds` point back at `d`, a directory that recursive diff had already visited directly. Newer `macos-latest` jobs that land on the `macos-26-arm64` runner image use Apple BSD diff behavior that follows that symlink and reports `Directory loop detected`. The same source revisions could still pass on `macos-15-arm64`, so this was runner-image-dependent fixture fragility rather than a copy_directory implementation regression.

Point `ds` at a separate fixture directory and exclude that direct target from the source-vs-copy diff. The tests still compare the logical `ds` path, so they continue to verify that copy_directory dereferences directory symlinks into real copied directories, including nested file symlinks. The exclusion only prevents BSD diff from visiting the same fixture contents once directly and once through the symlink.


fmeum:
Thanks, please run buildifier to fix the lint failures.

keegancsmith:
> Thanks, please run buildifier to fix the lint failures.

amended the commit and pushed

fmeum (approved):

PR #1265, opened 2026-07-07
feat: add fail_on to run_binary

alexeagle:
## Summary

Follow-up to #1260 implementing the `fail_on` API discussed in https://github.com/bazel-contrib/bazel-lib/pull/1260#issuecomment-4898266475.

- Adds `fail_on` to `run_binary` so only listed exit codes fail the action (e.g. `fail_on = [2]` for `diff`-like tools where exit `1` is a normal result).
- `spawn_binary` gains `--fail-on` (repeatable; comma-separated values in one flag are also accepted).
- When `fail_on` is set it takes precedence over the unconditional success implied by `exit_code_out`; the exit code is still written when `exit_code_out` is set.
- `silent_on_success` now uses the same success/failure decision as `fail_on` when deciding whether to forward buffered output.

## Test plan

- [x] `bazel test //tools/spawn_binary:spawn_binary_test` (parse, fail_on semantics, silent_on_success interaction)
- [x] `bazel test //lib/tests/run_binary/...` (integration tests for exit codes 0 and 1 with `fail_on = [2]`)

### DIFF
--- a/lib/private/run_binary.bzl
+++ b/lib/private/run_binary.bzl
@@ -300,10 +300,10 @@ def run_binary(
             listed exit codes cause the action to fail. All other exit codes are treated
             as success. This is useful for tools such as `diff` that use non-zero exit
             codes for non-error conditions (for example exit `1` when files differ and
-            exit `2` on error). Can be combined with `exit_code_out` to record the exit
-            code while still failing the action on specific codes. `silent_on_success`
-            uses the same success/failure decision when deciding whether to forward
-            buffered output.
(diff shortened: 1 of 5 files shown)


alexeagle:
Good catch — fixed in 1c054da. When `fail_on` matches a child exit of 0, the wrapper now returns 1 so Bazel still fails the action (supports grep-style `fail_on = [0]`). Added a unit test.

kormide (approved):

PR #1267, opened 2026-07-09
feat: return repo_metadata from repository rules

keith:
This allows the remote repo contents cache to cache the result of these


alexeagle (approved):
thanks!

PR #1269, opened 2026-07-16
feat: pass bare $(BINDIR) args in run_binary as File objects for path mapping

acozzette:
`$(BINDIR)` bakes in the unmapped `bazel-out/<cfg>/bin` prefix at analysis time, which breaks under Bazel's path mapping. When an args element is exactly `$(BINDIR)` (no prefix or suffix), add the root of any declared output directly via `Args.add_all`/`map_each` instead, keeping the action eligible for path mapping.

The motivation for this is to enable path mapping for `js_run_binary` in `aspect_rules_js`. Running a `js_binary` requires providing the Bazel output bin directory, so we need a path-mapping-friendly way to convey that information.

### DIFF
--- a/lib/private/run_binary.bzl
+++ b/lib/private/run_binary.bzl
@@ -122,6 +122,7 @@ Possible fixes:
             args.add_all([expansion_outputs[0]], map_each = _bindir_path, expand_directories = False)
         else:
             expanded = expand_variables(ctx, ctx.expand_location(a, targets = targets), inputs = inputs, outs = expansion_outputs)
+            # Disable path mapping if any location or variable expansion occurred
             can_path_map = can_path_map and expanded == a
             args.add_all(split_args(expanded))
     envs = {}
--- a/lib/private/run_binary.bzl
+++ b/lib/private/run_binary.bzl
@@ -122,6 +122,7 @@ Possible fixes:
             args.add_all([expansion_outputs[0]], map_each = _bindir_path, expand_directories = False)
         else:
             expanded = expand_variables(ctx, ctx.expand_location(a, targets = targets), inputs = inputs, outs = expansion_outputs)
+
             # Disable path mapping if any location or variable expansion occurred
             can_path_map = can_path_map and expanded == a
             args.add_all(split_args(expanded))
--- a/lib/private/run_binary.bzl
+++ b/lib/private/run_binary.bzl
@@ -20,6 +20,9 @@ load("//lib:stamping.bzl", "STAMP_ATTRS", "maybe_stamp")
 load(":expand_variables.bzl", "expand_variables")
 load(":strings.bzl", "split_args")
 
+def _bindir_path(file):
+    return file.root.path
+
 def _run_binary_impl(ctx):
     args = ctx.actions.args()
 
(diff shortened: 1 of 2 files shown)


lib/private/run_binary.bzl
```diff
@@ -108,9 +111,16 @@ Possible fixes:
...
     for a in ctx.attr.args:
-        expanded = expand_variables(ctx, ctx.expand_location(a, targets = targets), inputs = inputs, outs = expansion_outputs)
-        can_path_map = can_path_map and expanded == a
-        args.add_all(split_args(expanded))
+        if a == "$(BINDIR)":
+            # As a special case, we expand arguments exactly equal to
+            # "$(BINDIR)" directly, in a way that is compatible with path
+            # mapping. Every output has the same bin directory, so we just
+            # examine the first one.
+            args.add_all([expansion_outputs[0]], map_each = _bindir_path)
```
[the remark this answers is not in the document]
acozzette: Good catch. This is now fixed.

lib/private/run_binary.bzl
```diff
@@ -108,9 +111,19 @@ Possible fixes:
...
+            # "$(BINDIR)" directly, in a way that is compatible with path
+            # mapping. Every output has the same bin directory, so we just
+            # examine the first one. expand_directories is disabled since
+            # the first output may be a not-yet-created output directory
+            # (from out_dirs); we want its path, not its (currently
+            # nonexistent) contents.
+            args.add_all([expansion_outputs[0]], map_each = _bindir_path, expand_directories = False)
+        else:
+            expanded = expand_variables(ctx, ctx.expand_location(a, targets = targets), inputs = inputs, outs = expansion_outputs)
+            can_path_map = can_path_map and expanded == a
```
thesayyn, line 127: This deserves a comment a believe.

lib/private/run_binary.bzl
```diff
@@ -108,9 +111,19 @@ Possible fixes:
         args.add(ctx.executable.tool)
 
     for a in ctx.attr.args:
-        expanded = expand_variables(ctx, ctx.expand_location(a, targets = targets), inputs = inputs, outs = expansion_outputs)
-        can_path_map = can_path_map and expanded == a
-        args.add_all(split_args(expanded))
+        if a == "$(BINDIR)":
```
thesayyn, line 114: Should this only take affect when path mapping is requested?

lib/private/run_binary.bzl
acozzette: I added a short comment here explaining why we are updating `can_path_map`.

lib/private/run_binary.bzl
acozzette: I think I actually prefer doing it unconditionally, because this way we only have one code path to test. We also can't easily know whether we're doing path mapping until we've already gone through all the args and env variables.

fmeum:
Just fyi, I'm working on a generalized version of this that covers location expansion as well. This looks reasonable in the meantime!

acozzette:
@fmeum Yeah, coincidentally I happened to come across your PR this morning: https://github.com/bazelbuild/bazel/pull/30291 That will be a very nice feature to have, thanks for working on it!

alexeagle (approved):

PR #872, opened 2024-06-22
feat: support --noenable_runfiles on linux and windows, new runfiles bat macro, fix tests for windows

peakschris:
This PR contains an update of the runfiles code in windows_utils.bzl to match the runfiles v3 standard (compatible with latest bzlmod). I've also taken the opportunity to fix many tests to work under windows, both with --enable_runfiles and --noenable_runfiles, which required using the new runfiles code to correctly resolve data.

I suspect that the existing tests on linux don't/can't currently validate --noenable_runfiles behaviour of the underlying rules.

Whilst under linux `bazel test` always enables runfiles (ignoring the --noenable_runfiles flag https://github.com/bazelbuild/bazel/issues/22856), under windows this flag does have effect, and is the default, so getting tests clean on windows require test cases to handle runfiles manifests. Also, this would seem to be the only way that we can validate that the rules we are testing work correctly without runfiles.

I have added a test runner script that runs each test case with `bazel run`, simulating windows behaviour with --noenable_runfiles. This script has some caveats, namely that it writes test outputs to source tree. See lib/tests/README.md for more.

The only windows failures are now in directories with missing windows packages: lib/tests/tar,lib/tests/assert_archive_contains,lib/tests/zstd. Missing packages are unzip, bsdtar, zstd. Additionally, 3 bats tests timeout on windows, there seems to be some kind of concurrency issue. With tags=["exclusive"], they pass occasionally but most often still timeout.

@meteorcloudy @fmeum 

---

### Changes are visible to end-users: yes

- Searched for relevant documentation and updated as needed: no
- Breaking change (forces users to change their own code or config): no
- Suggested release notes appear below: yes

#### Suggested release notes

- Some aspect rules now support --noenable_runfiles on linux and windows (the default on windows)

---

### Test plan

- It would be good to get some CI runners enabled to ensure windows does not regress
- the test.sh script could be called on all runners to test all combinations of test/run and enable/noenable runfiles

**Test results: linux - before**
```
[test enable runfiles  ] Executed 179 out of 187 tests: 179 tests pass and 8 were skipped.
[test noenable runfiles] Executed 179 out of 187 tests: 179 tests pass and 8 were skipped.
[run enable runfiles   ] Executed 179 out of 187 tests: 179 tests pass and 8 were skipped.
Updated 6 paths from the index
[run noenable runfiles ] Executed 179 out of 187 tests: 150 tests pass, 29 fail and 8 were skipped.
Updated 0 paths from the index
```
**Test results: linux - after**
```
[...]

PR #1274, opened 2026-07-24
fix: support directory runfiles in Windows launcher

### LINKED ISSUE #1273: Windows native launcher fails with directory-based runfiles
`create_windows_native_launcher_script` in `lib/windows_utils.bzl` currently forces:

```bat
set RUNFILES_MANIFEST_ONLY=1
```

and its `:rlocation` implementation requires a runfiles manifest.

This fails for Windows remote execution environments where Bazel provides a runfiles directory but no manifest. Bazel’s native Windows launcher supports this case by falling back to directory-based lookup through `RUNFILES_DIR`.

### Current behavior

When the generated launcher runs without a manifest, it exits with an error similar to:

```text
ERROR: Manifest file ... does not exist.
```

The non-manifest branch in `:rlocation` also returns the logical runfiles path unchanged rather than resolving it relative to `%RUNFILES_DIR%`, so removing `RUNFILES_MANIFEST_ONLY=1` alone would not fix the issue.

### Expected behavior

The generated launcher should support both standard Windows runfiles layouts:

1. Use `RUNFILES_MANIFEST_FILE` when a valid manifest is available.
2. Otherwise resolve runfiles relative to `RUNFILES_DIR`.
3. When neither variable identifies a valid location, discover the usual adjacent manifest and `.runfiles` paths.

This would match the behavior of Bazel’s native [Windows launcher](https://github.com/bazelbuild/bazel/blob/df5a815925dfb5acfbedbd799fdfc18d153fa634/src/tools/launcher/launcher.cc#L81), which supports manifest-based local execution and directory-based remote execution.

### Relevant code

* `lib/windows_utils.bzl`
* `create_windows_native_launcher_script`
* The generated batch file’s `:rlocation` routine

### Prior work

PR #872 implemented this fallback as part of a broader Runfiles v3 change:

That PR was closed after a request to split it into smaller, independently reviewable changes. This issue is intended to track only the Windows manifest-to-directory fallback.

A targeted fix could extract the relevant behavior from that PR without including its unrelated test and runfiles changes.

alex-the-third:
## Summary

Support Windows remote execution environments where Bazel provides a runfiles directory without a manifest.

The generated launcher currently forces `RUNFILES_MANIFEST_ONLY=1`, and its `:rlocation` routine requires a manifest. Simply removing that override is insufficient because the non-manifest path is returned unchanged instead of being resolved beneath `RUNFILES_DIR`.

This change:

- uses `RUNFILES_MANIFEST_FILE` when it identifies a valid manifest
- otherwise resolves runfiles relative to a valid `RUNFILES_DIR`
- falls back to the usual adjacent `.runfiles/MANIFEST`, `.runfiles_manifest`, and `.runfiles` locations
- updates the exported runfiles variables so downstream Bash runfiles lookups use the selected layout
- removes the unconditional `RUNFILES_MANIFEST_ONLY=1` override
- adds targeted Windows tests for explicit and adjacent manifest/directory layouts and their precedence

This matches Bazel’s native Windows launcher behavior while keeping the change scoped to the manifest-to-directory fallback tracked by the linked issue.

Fixes #1273 

## Comments

This extracts the gist of #872.

This should be the followup of #1272 <s>as the first commit here is from the other PR.</s> EDIT: the PR was merged, so rebased.

### DIFF
--- a/lib/tests/windows_utils/BUILD.bazel
+++ b/lib/tests/windows_utils/BUILD.bazel
@@ -1,10 +1,48 @@
+load(":directory_runfiles_test.bzl", "windows_directory_runfiles_test")
 load(":exact_manifest_match_test.bzl", "windows_exact_manifest_match_test")
+
+WINDOWS_ONLY = select({
+    "@platforms//os:windows": [],
+    "//conditions:default": ["@platforms//:incompatible"],
+})
 
 windows_exact_manifest_match_test(
     name = "exact_manifest_match_test",
     size = "small",
-    target_compatible_with = select({
-        "@platforms//os:windows": [],
-        "//conditions:default": ["@platforms//:incompatible"],
-    }),
+    target_compatible_with = WINDOWS_ONLY,
+)
+
+windows_directory_runfiles_test(
+    name = "explicit_manifest_test",
+    size = "small",
+    scenario = "explicit_manifest",
+    target_compatible_with = WINDOWS_ONLY,
+)
+
+windows_directory_runfiles_test(
+    name = "explicit_directory_test",
+    size = "small",
+    scenario = "explicit_directory",
+    target_compatible_with = WINDOWS_ONLY,
+)
+
+windows_directory_runfiles_test(
+    name = "adjacent_nested_manifest_test",
+    size = "small",
+    scenario = "adjacent_nested_manifest",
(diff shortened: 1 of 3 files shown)


jbedard:
@codex 🙏

